PR #1, opened 2021-10-14
Fix trailspace issues

levouh:
Fixes two issues:

1. Highlight group can't be changed pending load order as it stands right now, use `default` when calling `hi` to handle these cases
2. Wrap the call to `vim.fn.matchdelete` with a `pcall` to avoid breaking things like _Telescope_ if the match doesn't exist

Also changes some calls that use `vim.fn` to use `vim.api` and check valid state, etc.

---

Curious, is there a reason to `defer_fn` on the `highlight` `:h autocmd` call? If we do that, the call to get the "current window" might not be accurate. Wanted to check before trying to change it here.

### DIFF
--- a/lua/mini/trailspace.lua
+++ b/lua/mini/trailspace.lua
@@ -63,7 +63,7 @@ function MiniTrailspace.setup(config)
   )
 
   -- Create highlighting
-  vim.api.nvim_exec([[hi link MiniTrailspace Error]], false)
+  vim.api.nvim_exec([[hi default link MiniTrailspace Error]], false)
 end
 
 -- Module config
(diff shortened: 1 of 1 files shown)


levouh:
> Nope :( Having `vim.fn.expand("<abuf>")` [in here](https://github.com/echasnovski/mini.nvim/blob/9bce332c0d76cd9a26e58fbe9795b82931a6f5cd/lua/mini/trailspace.lua#L84) (and removing `defer_fn`) doesn't seem to work.
> 
> What I believe might be happening is that `modified` is still `false` at the moment of autocommand execution. It will be changed later by plugins themselves.

Gotcha. So it seems then that one other solution would be to save off the `winid` while the `autocmd` is executing, but then perform the rest of the functionality we want later. Something like:

```lua
  vim.api.nvim_exec(
    [[augroup MiniTrailspace
        au!
        au WinEnter,BufWinEnter,InsertLeave * lua MiniTrailspace.defer_highlight()
        au WinLeave,BufWinLeave,InsertEnter * lua MiniTrailspace.unhighlight()

        au FileType TelescopePrompt let b:minitrailspace_disable=v:true
      augroup END]],
    false
  )

  -- Create highlighting
  vim.api.nvim_exec([[hi link MiniTrailspace Error]], false)
end

-- Module config
MiniTrailspace.config = {}

function MiniTrailspace.defer_highlight()
  local win_id = vim.api.nvim_get_current_win()

  vim.schedule(function()
    MiniTrailspace.highlight(nil, { win_id = win_id })
  end)
end

-- Functions to perform actions
--- Highlight trailing whitespace
---
---@param check_modifiable boolean: Whether to check |modifiable| (if it is off, don't highlight). Default: `true`.
---@param opts table: Additional options/state for highlighting, currently:
---                     - win_id: The window when the autocommand was invoked
function MiniTrailspace.highlight(check_modifiable, opts)
  check_modifiable = check_modifiable or true

  -- Highlight only in normal mode
  if H.is_disabled() or vim.fn.mode() ~= 'n' then
    MiniTrailspace.unhighlight()
    return
  end

  if check_modifiable and not vim.bo.modifiable then
    return
  end

  local win_id = opts.win_id or vim.api.nvim_get_current_win()
  if not vim.api.nvim_win_is_valid(win_id) then
    return
  end

...
```

would work then?

echasnovski:
Just let you know that I removed `defer_fn()` by slightly reframing the problem it used to solve. Now instead of `modifiable`, tracking of `buftype` option is done via `if-else` and `OptionSet` autocommand. Activated with `config.only_in_normal_buffers`.

PR #7, opened 2021-11-01
Implemented 'Fly Mode', inspired by jiangmiao/auto-pairs

xigoi:

[It's fine if you don't think this belongs in the source repo, I primarily implemented it for myself and it can stay in my fork if you don't like it]

I've implemented the "Fly Mode" functionality from [jiangmiao/auto-pairs](https://github.com/jiangmiao/auto-pairs). It allows you to easily jump out of a pair of brackets no matter where you are. I personally find it very convenient, especially when dealing with nested brackets of different kinds.


echasnovski:
Thank you for this contribution!

I personally never really found fly mode convenient enough, so it wasn't included. Is it really that useful?

For me there are several concerns here:
- It somewhat contradicts the overall idea that functionality depends only on cursor's neighborhood.
- It also makes function names `close()` and `closeopen()` not quite aligned.
- In 'jiangmiao/auto-pairs' it works only for `)`, `}`, `]`. So I wonder if `config.fly_mode` should be supplied as table with closing symbols for which fly mode will be active.
- In original functionality there is also a `AutoPairsShortcutBackInsert` for when flying was not intended. Is current functionality sufficiently useful for you without this behavior?

So overall I don't see a problem with (and not opposed to) functionality itself. How to nicely fit it inside current framework, that is an issue. I'll ponder on it for a bit (*quite* a bit actually, as there are other plans for this plugin) and see what I can come up with. If you have any suggestions on these concerns, I'll gladly hear them!

xigoi:
I find that it makes opening and closing brackets more symmetrical. When inserting an opening brace, it's always inserted as a pair, and if you don't want this, you use `<c-v>`. Analogically, a closing brace will always close a pair (that is, go past an already inserted closing brace) and if you don't want that, you can use `<c-v>`.

> It somewhat contradicts the overall idea that functionality depends only on cursor's neighborhood.

I did find it weird in this way when I learned about it in auto-pairs, but I decided to try it and started to see the usefulness.

> It also makes function names close() and closeopen() not quite aligned.

That's true, but I don't really expect self-paired delimiters to behave consistently with paired delimiters anyway. Maybe your expectations are different. I'd kind of unfortunate that both are commonly used in programming.

> In 'jiangmiao/auto-pairs' it works only for ), }, ]. So I wonder if config.fly_mode should be supplied as table with closing symbols for which fly mode will be active.

That's a good idea. Maybe someone will want to be able to quickly jump to the end of a block in C-like languages, but not find the functionality useful for parentheses.

> In original functionality there is also a AutoPairsShortcutBackInsert for when flying was not intended. Is current functionality sufficiently useful for you without this behavior?

To be honest, I've never used it and forgot that it exists. I just use `<c-v>` when I want to insert an unpaired closing bracket.

echasnovski:
> I find that it makes opening and closing brackets more symmetrical. When inserting an opening brace, it's always inserted as a pair, and if you don't want this, you use `<c-v>`. Analogically, a closing brace will always close a pair (that is, go past an already inserted closing brace) and if you don't want that, you can use `<c-v>`.

Well that is a very nice way of thinking about it!

I too find explicitly using `<C-v>` as the most straightforward way to not use auto pairs.
I'll think about this and eventually get back. Thank you again.

echasnovski:
Sorry, Adam. I've looked into 'mini.pairs' and updated it where it felt lacking functionality. Currently I don't think that 'Fly Mode' is something I want to implement and support in 'mini.pairs'.
But anyway, thank you for your contribution.

xigoi:
No problem, I'm going keep it in my fork.

PR #54, opened 2022-03-20
Feat(starter): Allow pass commands to sections.telescope

milanglacier:

`starter.sections.telescope` is very convenient, but allowing customized commands would be best.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 doc/tags
 deps
+.DS_Store
--- a/doc/mini.txt
+++ b/doc/mini.txt
@@ -3035,8 +3035,12 @@ Return~
 
 ------------------------------------------------------------------------------
                                               *MiniStarter.sections.telescope()*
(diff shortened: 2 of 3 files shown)


echasnovski:
Thanks for suggestion!

At first glance, I am not really convinced that this adds much non-trivial functionality compared to added code base complexity. Also it creates names identical to `Telescope` action. Probably, having `key` represent `name` might be better, but this also has hidden inconvenience of user not being able to define order of items (`pairs` iterates in pre-defined order of keys).

Do you think that having to do manual set of items is so bad that it needs this functionality?

milanglacier:
Yep, this is very trivial. It is just that I only need telescope and a wrapper is convenient than manually write all stuffs. 
So if you want to keep the code simplicity, I totally understand.

echasnovski:
Thanks for understanding! Considering growing code base, its complexity is something that I feel should be highly valued.

PR #87, opened 2022-06-13
Fallback if nerdfonts is not enabled

tani:

Thank you for the great project. According to the documentation, Nerd Fonts is an option for statusline. However, the current default behavior requires Nerd Fonts and moreover, it is a bit hard to disable Nerd Fonts. Hence, we add the new option 'nerdfonts' to disable easily. Cheers,


echasnovski:
Hi! Thanks for contribution, it looks relevant.

Is there a reason of why not use `Git` and `LSP` as custom icons in `args`? Like `MiniStatusline.section_git({ trunc_width = 75, icon = 'Git' })`. I understand that this will need custom active content function (possibly quite verbose), but anything else besides this?

I am more curious about `MiniStatusline.section_fileinfo()` with `nvim-web-devicons` installed. Does it work incorrectly without Nerd font (shows weird square in place of an actual icon)?

tani:
> Is there a reason of why not use Git and LSP as custom icons in args? Like MiniStatusline.section_git({ trunc_width = 75, icon = 'Git' }). I understand that this will need custom active content function (possibly quite verbose), but anything else besides this?

True. This is the reason why I created it. Currently, I am using `MiniStatusline.section_git` to suppress NerdFonts. Let me introduce my problem more.

MiniStatusline has  fantastic optional parameters to decorate statusline.

- `git` status: it will be disabled if `gitsign. nvim` is disabled.
- `fileinfo` status: it will fall back if `kyazdani42/nvim-web-devicons` is disabled.

The above option provides the fallback-mechanism if they are disabled.
However, the NerdFonts has no fallback-mechanism, that is, we need to manually make
statusline function to fallback. This is harder than others. Of course, I know that font detection is impossible from the terminal version of Neovim. Instead, I propose an easier way to disable it manually to fill the gap. 

Further, we do not change any lines about section_fileinfo. Thus, it should work well.

echasnovski:
Thanks for explanation. I understand your reasoning more clear now.

> Further, we do not change any lines about section_fileinfo. Thus, it should work well.

I understand that. What I tried to ask you (as at the moment I can't reproduce "non-nerd-font" environment), is how it works *right now* under following conditions:
- Non-nerd-font.
- 'nvim-web-devicons' installed.
- `section_fileinfo()` active.

I imagine, it should act similarly to icons in git and diagnostic sections, but not sure.

All in all, at the moment I am more inclined to add this option, but in more general `use_icons` form. If `false`:
- In `section_git()` fallback to `"Git"` (as you implemented).
- In `section_diagnostic()` fallback to some short string indicating diagnostic. `"LSP"` is fine, but something more targeted would be better.
- In `section_fileinfo()` fallback to as if 'nvim-web-devicons' is not installed.

I'll think about it a little bit more and get back to you, hopefully, tomorrow.

tani:
Noted. For the above situation (Non NerdFonts, 'nvim-web-devicons' is installed, `section_fileinfo` active), it should be disabled because the double-width icons breaks the window rendering for some neovim frontend. We need to disable it for such a frontend.
 Thus, I change the behavior to return a blank string in the latest commits. Further, I agree the option name (`nerdfonts`) is not so general name. Hence, I have changed it to `use_icons`. Please check it.

tani:
Thank you for adopting my suggestion. I'm very happy. I'm sorry about the lack of testing and documentation, but I don't know enough about them, so please work on them yourself.

echasnovski:
> Thank you for adopting my suggestion. I'm very happy. I'm sorry about the lack of testing and documentation, but I don't know enough about them, so please work on them yourself.

Already done in latest `main`.

It was not that complex, rather tedious with current testing setup. And yeah, implementing new option is especially hard on documentation and implementation. Probably should document it in full somewhere.

PR #88, opened 2022-06-14
Add autochdir

tani:

Automatically change directory to a project root.
Main inspiration is a "mattn/vim-findroot" and "airblade/vim-rooter" plugin.

Features:
- Automatically change directory to a project root defined in
 `MiniAutochdir.config`. The patterns are written in the glob pattern.

General overview of how to find a project root:
- Check whether the designated file or the directory exists in current directory.
- If not, move to the parent directory to find a project root.

### DIFF
--- a/doc/mini.txt
+++ b/doc/mini.txt
@@ -224,8 +224,8 @@ Setup with `require('mini.autochdir').setup({})` (replace `{}` with your
 you can use for scripting or manually (with `:lua MiniAutochdir.*`).
 
 # Example usage~
-- Modify default patterns to find a project root: >
-  require('mini.autochdir').setup({ patterns = { '.git' }})
+- Modify default root_pattern to find a project root: >
+  require('mini.autochdir').setup({ root_pattern = { '.git' }})
 - At the time changing a buffer, it also changes CWD.
 
 # Disabling~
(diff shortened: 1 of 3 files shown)


echasnovski:
I am really, **really** grateful for all your time and effort you put in this PR. But please, before committing yourself to doing some code related work, make sure that it aligns with project vision. To avoid situations like this, I explicitly put [those lines](https://github.com/echasnovski/mini.nvim/blob/066688ae4aa3c7293009cc62768495a149e9ef9c/CONTRIBUTING.md?plain=1#L11) in contributing guidelines (maybe should make them more visible, to be honest).

The 'airblade/vim-rooter' is an essential part of my workflow. For the long time I had certain ideas of what would make its good replacement in 'mini.nvim' (probably, named 'mini.root'). So I came up with a more "functional" design where user can define a root-finding function (instead of patterns) as a predicate for a directory path. It will work by traversing ancestors of buffer path and stopping when predicate gives `true`. To make it more convenient, there would be some predicate helpers (like `has_child()`, `has_sibling()`, etc).

So unfortunately, your implementation doesn't go well with this intended design. I am not opposed to someone else making this module, but there will be **a lot of small details** that I want to get right with it. If you are willing to spend considerably more time on it, I am happy to collaborate.

echasnovski:
Not a problem. Huge thanks again for your efforts!

PR #91, opened 2022-06-23
Fix mini.jump highlighting & command output

thenyeguy:
I started using mini.jump recently, and noticed a couple issues.

1. Highlights don't include capital letters, but you still jump to them by default. After a bit of testing, I confirmed this is because I use the `ignorecase` and `smartcase` vim options.
2. Highlights appear when used with an operator, despite the fact that you can't actually perform a repeat.
3. In operator pending mode, we print a spurious `lua MiniJump.jump()` to the statusline.

I made a few commits to polish these up.

Regarding case sensitivity, I personally _like_ that this plugin follows `ignorecase`, and so all I did was fix the disparity between jumping and highlighting. However, it does seem like respecting `ignorecase` in this plugin may not have been intentional. If you'd rather fix this, I would be happy to instead work on adding explicit `ignorecase`/`smartcase` options instead.

### DIFF
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -168,17 +168,21 @@ function MiniJump.jump(target, backward, till, n_times)
 
   -- Construct search and highlight patterns
   local flags = MiniJump.state.backward and 'Wb' or 'W'
-  local pattern, hl_pattern = [[\V%s]], [[\V%s]]
+  local hl_case = ''
+  if vim.o.ignorecase and (not vim.o.smartcase or escaped_target == escaped_target:lower()) then
+    hl_case = [[\c]]
+  end
+  local pattern, hl_pattern = [[\V%s]], [[\V%s%s]]
   if MiniJump.state.till then
     if MiniJump.state.backward then
-      pattern, hl_pattern = [[\V\(%s\)\@<=\.]], [[\V%s\.\@=]]
+      pattern, hl_pattern = [[\V\(%s\)\@<=\.]], [[\V%s%s\.\@=]]
       flags = ('%se'):format(flags)
     else
-      pattern, hl_pattern = [[\V\.\(%s\)\@=]], [[\V\.\@<=%s]]
+      pattern, hl_pattern = [[\V\.\(%s\)\@=]], [[\V%s\.\@<=%s]]
     end
   end
 
-  pattern, hl_pattern = pattern:format(escaped_target), hl_pattern:format(escaped_target)
+  pattern, hl_pattern = pattern:format(escaped_target), hl_pattern:format(hl_case, escaped_target)
 
   -- Delay highlighting after stopping previous one
   H.timers.highlight:stop()
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -13,6 +13,11 @@
 --- - Normal, Visual, and Operator-pending (with full dot-repeat) modes are
 ---   supported.
 ---
+--- This module follows vim's 'ignorecase' and 'smartcase' options. When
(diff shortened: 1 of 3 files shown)


lua/mini/jump.lua
```diff
@@ -271,7 +275,7 @@ function MiniJump.expr_jump(backward, till)
   end
   H.update_state(target, backward, till, vim.v.count1)
 
-  return vim.api.nvim_replace_termcodes('v:<C-u>lua MiniJump.jump()<CR>', true, true, true)
+  return vim.api.nvim_replace_termcodes('v<cmd>lua MiniJump.jump()<CR>', true, true, true)
```
echasnovski, line 278: Please, use `<Cmd>` instead of `<cmd>`. I try to consistently use it across all modules, as it is how it is written in Vim help docs.

lua/mini/jump.lua
```diff
@@ -388,6 +392,11 @@ function H.highlight(pattern)
     return
   end
 
+  -- Do not highlight in operator-pending mode, as it does not support repeating.
+  if string.find(MiniJump.state.mode, 'o') then
```
echasnovski, line 396: Let's not do that. Although `fFtT` is not repeatable in this mode, there is a dot-repeat. It proved to be useful when doing something like `dte` and then `.` to repeat deleting.

echasnovski (changes requested):
This looks great! Thanks!

It indeed was not a conscious decision to support 'ignorecase' and 'smartcase', but glad it came out naturally. So yeah, supporting coherent highlighting in these cases is a must.

Please, add separate tests for both 'ignorecase' and 'smartcase'. For this, you probably need to update [main helper](https://github.com/echasnovski/mini.nvim/blob/cc8fb168b95620ec1611916a6d2a74e6a5cdcbdb/tests/jump_spec.lua#L22) and add correct test cases at the [end of corresponding block](https://github.com/echasnovski/mini.nvim/blob/cc8fb168b95620ec1611916a6d2a74e6a5cdcbdb/tests/jump_spec.lua#L867-L868)

lua/mini/jump.lua
thenyeguy: Ahh, dot repeating is a good point! I discarded this commit.

thenyeguy:
Added tests and explicit documentation for both `ignorecase` and `smartcase`. I tried messing with the `highlights_target` but its behavior seemed odd. I ended up just constructing highlight test cases that only had one match and that seemed to work fine.

tests/jump_spec.lua
```diff
@@ -662,6 +662,54 @@ describe('Jumping with f/t/F/T', function()
...
+
+    set_cursor(1, 10)
+    type_keys('2F', 'E')
+    eq(get_cursor(), { 1, 6 })
+
+    set_cursor(1, 10)
+    type_keys('2T', 'E')
+    eq(get_cursor(), { 1, 7 })
+
+    child.cmd('set noignorecase')
```
echasnovski, line 686: For possible future: no need to unset anything, as child process is restarted before every case here.

tests/jump_spec.lua
```diff
@@ -662,6 +662,54 @@ describe('Jumping with f/t/F/T', function()
...
+  end)
+
+  it('respects vim.o.smartcase', function()
+    child.cmd('set ignorecase')
```
echasnovski, line 690: Nitpick: this could have been a single call `set ignorecase smartcase`.

echasnovski (approved):
Thanks!

It has slight differences to how I would make this, but I plan to rewrite all tests for different framework, so patch it up later.

PR #94, opened 2022-07-03
 add only_in_cursor_line option

ryoppippi:

Hi!
I use this plugin everyday and I love it. Thank you for maintaining it.

One of my favorite is `mini.jump`, and I would like to add one option for that.

This is called `only_in_cursor_line`. It is similar to `g:clever_f_across_no_line` in clever-f (https://github.com/rhysd/clever-f.vim/blob/master/doc/clever_f.txt).

If it looks good for you, please consider to merge it.
Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,11 @@
 <img src="logo.png" width="800em"/> <br>
 
 <!-- badges: start -->
+
 [![GitHub license](https://badgen.net/github/license/echasnovski/mini.nvim)](https://github.com/echasnovski/mini.nvim/blob/main/LICENSE)
 [![GitHub tag](https://badgen.net/github/tag/echasnovski/mini.nvim)](https://github.com/echasnovski/mini.nvim/tags/)
 [![Current version](https://badgen.net/badge/Current%20version/development/cyan)](https://github.com/echasnovski/mini.nvim/blob/main/CHANGELOG.md)
+
 <!-- badges: end -->
 
 Collection of minimal, independent, and fast Lua modules dedicated to improve [Neovim](https://github.com/neovim/neovim) (version 0.5 and higher) experience. Think about it as "Swiss Army knife" among Neovim plugins: it has many different independent tools (modules) suitable for most common tasks. Each module can be used as a separate sub-plugin without any startup and usage overhead.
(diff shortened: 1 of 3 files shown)


echasnovski:
Thanks for your kind words and usage!

For some reason, I am not sure about adding this option. It will complicate logic of already quite complex code for something I perceive as not much benefit. Besides, I like the current situation of minimum options.

As a side note, highlighting should also be updated for this option. It should be possible in theory, but needs further investigation.

I'll think about it for some time and get back to you, ok?

ryoppippi:
Thank you for reviewing!
Oh yeah I forgot to manage highlights. 
The reason to add this option is because it is better when I execute mini.jump as a macro per line. 

Anyway I'll wait until you are ready. Thanks;)

echasnovski:
Thanks again for suggestion!

I think I am going through with this suggestion, but want to make it more aligned with 'mini.jump2d' module by adding `allowed_lines` setting. Like this:
```lua
...
  allowed_lines = {
    cursor_before = true, -- Lines before cursor line
    cursor_at = true, -- Cursor line
    cursor_after = true, -- Lines after cursor line
  },
...
```

With this, your request is covered by `cursor_before = false, cursor_after = false`.

I believe this is doable by combining special `\%<23l`, `\%23l`, `\%>23l` patterns to match specific lines (not `\%<.l` and friends, because they are not present in Neovim<=0.6). This should handle both highlighting and search.

I am comfortable with either implementing this myself or guiding you through. It will require implementing custom helper for dealing with this setting, as well as properly documenting and extensively testing it.

ryoppippi:
Thank you for suggesting this to me.

I think your suggested new option looks good. I would like to implement this.

echasnovski:
> Thank you for suggesting this to me.
> 
> I think your suggested new option looks good. I would like to implement this.

Great!

Couple of useful places where to look:
- [Here](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md#implementation-notes) are some implementation notes with checklist for implementing new config setting. When you'll be editing 'CHANGELOG.md' don't forget to put the PR number and your Github username for credit.
- [TESTING.md](https://github.com/echasnovski/mini.nvim/blob/main/TESTING.md) has more general overview of testing infrastructure if reading existing tests is not enough.

Maybe it is a good idea to do this in a new PR? Everything from this one is already outdated.
Oh, and please don't change general formatting of 'README.md'.

PR #100, opened 2022-07-30
(mini.surround) Implement dedicated visual mode mapping

Oliver-Leete:

I ran into an issue using the suggested method for tpope/vim-surround style mappings. The add mapping (`ys`) was still being made in visual mode, so it was adding a delay to visual yanking. 

My suggestion is to add an optional extra mapping for visual_add. If it is not given then the mappings default to their previous behaviour.

I think I've updated all relevant documentation. And all tests are passing. Hopefully I haven't missed anything.

### DIFF
--- a/lua/mini/surround.lua
+++ b/lua/mini/surround.lua
@@ -166,6 +166,7 @@ end
 ---     },
 ---     mappings = {
 ---       add = 'ys',
+---       visual_add = 'S',
 ---       delete = 'ds',
 ---       find = '',
 ---       find_left = '',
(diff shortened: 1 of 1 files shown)


echasnovski:
Maybe it better to document deleting `ys` mapping in Visual mode?
Add `vim.api.nvim_del_keymap('x', 'ys')` after `require()` call in example.

Oliver-Leete:
That is defiantly a simpler solution. I always forget about the ability to delete mappings.

echasnovski:
> That is defiantly a simpler solution. I always forget about the ability to delete mappings.

Knowing much does that sometimes :)

PR #105, opened 2022-08-06
(mini.starter) Reset query after activation when evaluate_single is true

assistcontrol:

When `evaluate_single == true`, it makes sense to me to clear the query after an item is activated.

If I'm reading it correctly, the only benefit to keeping the query string is to add more onto it, but `evaluate_single` has already ensured that the query string is at a leaf---adding more to the query string will never match any more or less than it does already.

Resetting the query string after activating a single item brings mini.nvim more in line with startify and alpha. Start screen items are logically discrete, and after activating one the need to backspace out the current query feels like a "gotcha."

### DIFF
--- a/lua/mini/starter.lua
+++ b/lua/mini/starter.lua
@@ -1241,6 +1241,7 @@ H.make_query = function(buf_id, query)
   -- Possibly evaluate single active item
   if H.get_config().evaluate_single and n_active == 1 then
     MiniStarter.eval_current_item(buf_id)
+    MiniStarter.set_query(nil, buf_id)
     return
   end
 


echasnovski:
Thanks for suggestion!

If implemented as is, it feels a bit inconsistent with hitting `<CR>`, when item gets evaluated. In other words, if it is to be changed, it should be "reset query when item is evaluated". More so, it should be implemented to work when evaluation means "close this buffer and open another" (which is a most common use case). Quick test shows that it might cause hit-enter-prompt.

Initial motivation for keeping a query string was to help when user mistypes. And also it seems to me a bit more intuitive to not automatically change query without user requesting so.

At the moment I am hesitant to add this. Main reason is that the design of item evaluation in Starter buffer is that it leads to closing current buffer, so it should not really matter. To mitigate the inconvenience, there is `<Esc>` that clears all query at once: one extra keystroke seems to not be *that* inconvenient. Also, if you use custom items, you should be able to add query reset in actions themselves.

Would you mind sharing your use case in which not having this behavior causes you troubles?

assistcontrol:
Hrmmn, I hadn't considered that, and that's a good point.

I have multiple items in the Starter buffer that don't lead to closure. I have an options that run `PackerSync`, `PackerCompile`, and `GoUpdateBinaries`. I've been toying with items that open an `nvim-tree` window, reload my nvim config, etc. There are definitely creative things that could be good options for linkage from the Starter buffer. It's not unusual that I run `PackerSync` and then fire a `Telescope file_browser`, cancel it and browse a specfic dir, cancel it and open a blank buffer. Once an item is activated, my intent is always to run another; keeping the query string doesn't benefit me.

The behaviour that makes sense to me is for the query to be reset when an item is activated. If I activate an item and mini.starter is still open, I'm going to want to activate another item. Having the query retain the previous characters doesn't provide me any benefit.

You're absolutely right that `<Esc>` is a very minor inconvenience. My argument is that if it's not necessary, perhaps there's an opportunity to support that behaviour.

Perhaps there's an opportunity for a `reset_query_after_action` option that globally clears the query after activation of any item? In startify and alpha modes, setting that option makes sense as it matches their behaviour.

echasnovski:
> Having the query retain the previous characters doesn't provide me any benefit.

Yes, I understand that. My original thinking was that it might benefit if latest character was a typo (not having to type whole query is a plus here) while having `<Esc>` allows dealing when this is not a typo. While resetting after evaluation doesn't provide an easy way in case of a typo.

> Perhaps there's an opportunity for a reset_query_after_action option that globally clears the query after activation of any item?

Not a fan of another option at the moment.

I'll think about this a little more.

assistcontrol:
> Yes, I understand that. My original thinking was that it might benefit if latest character was a typo (not having to type whole query is a plus here) while having `<Esc>` allows dealing when this is not a typo. While resetting after evaluation doesn't provide an easy way in case of a typo.

I can’t dispute that; it would certainly negate the benefits of typo correction.

I will note that the startify/alpha paradigm commonly has single-character activators. In that modality, typo correction is not really a factor.

echasnovski:
In the end I opted to adopting your suggestion. Should be available on latest `main`.
Thanks again fro constructive feedback!

PR #107, opened 2022-08-09
(mini.ai) Update  to allow custom treesitter directives

litao91:

### DIFF
--- a/doc/mini.txt
+++ b/doc/mini.txt
@@ -931,9 +931,6 @@ Notes:
 - It uses buffer's |filetype| to determine query language.
 - On large files it is slower than pattern-based textobjects. Still very
   fast though (one search should be magnitude of milliseconds).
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks! I don't know why I didn't think about this in the first place, because custom directives in 'nvim-treesitter' are quite common. Probably because of small performance penalty and handling of dependency.

It needs some further non-trivial updates (a little bit of refactor and tricky tests), so I'll take it from here.

echasnovski:
I experimented a bit with using 'nvim-treesitter.queries'. On large files it takes about 3 to 4 times more time to produce output. Not sure how to deal with this at the moment. Will investigate further.

echasnovski:
@litao91, I added using 'nvim-treesitter' plugin's query mechanism if it is present. It is indeed 3 to 4 times slower, so I decided to also include option to ignore it.

Thanks for this idea and initial implementation.

PR #115, opened 2022-08-18
(mini.surround): Fix visual mode when selection=exclusive.

wodesuck:


echasnovski:
Oh, thanks, this is useful! I personally don't use `'exclusive'`, so didn't account for this. However, this should take into account multibyte characters.

I'll get back to this after (or during) undergoing update of 'mini.surround'.

wodesuck:
Multibyte is already taken into account, since `'>'` mark the position of the first byte next to the selected range when `'exclusive'`, you could just `-1` to get the last byte in the selected range.

echasnovski:
Thanks for your contribution! This should be implemented on latest `main`.

Issue #125, opened 2022-09-08
indent: When the indent marker is displayed it removes 'listchar' for tabs before it.


cryptomilk:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

indent

### Description

When the indent marker is displayed and you use:

```lua
vim.opt.listchars       = { tab     = '»·', }
```

Then the it will remove the listchar for indented code for more than one level.

![ident_with_tab_listchar](https://user-images.githubusercontent.com/545480/189061539-c9f2081f-38f0-4763-93da-d16013eebec1.png)

### Neovim version

0.7.2

### Steps to reproduce

1. In your config set:
```lua
vim.opt.listchars       = { tab     = '»·', }
```
2. configure mini.nvim indent
3. Open source code with tabs e.g. Samba source code

### Expected behavior

The listchar should not be removed by the indent marker.

### Actual behavior

The listchar is removed by the indent marker. See image above.

echasnovski:
Thanks for the issue!

Unfortunately, this is how extmarks (which are used for displaying scope) work right now, to the best of my knowledge. This is a combination of two limitations:
- Marks can only be placed beside text, but can be arbitrary width. So to enable showing scope on empty lines, extmarks are placed in the first column and have padded text for them.
- There is no way to have different priority for different parts of extmark text (highlight group - yes, priority - no). So until neovim/neovim#17366 is resolved, I can't see a way to solve this reliably. And even after that, I am not sure how priority of extmarks will interact with listchars.

cryptomilk:
Thank you very much for looking into that.

echasnovski:
This should be fixed in latest `main`.

cryptomilk:
Works great, thank you very much!

PR #161, opened 2022-11-15
(mini.indentscope) Add option for a prefix symbol

mivort:
This small change adds an ability to use a user-defined symbol for a prefix instead of only a hard-coded space character. The idea is to allow the usage of this plugin with a leading space 'listchar' option enabled.

## Context

Hi! Thanks a lot for a very cool and minimalistic plugin to provide a nice indent visualization without introducing a performance penalty!

I've been using a `listchars` vim option for a while to visualize leading tabs and whitespaces, so when I've stared using `mini.indentscope` plugin, I've noticed that it overwrites existing `NonText` marks with a blank space:

before

I think the option to provide a user-defined prefix character instead of hard-coded `' '` could be useful to `listchars` users without introducing a performance penalty. With this patch it's possible to keep the visible whitespaces while getting a highlighted scope under cursor (using the following config):
```lua
vim.o.listchars = 'lead:┆'
require('mini.indentscope').setup { prefix = '┆' }
```

after

It can also provide users with ability to customize scope visualization with existing `MiniIndentscopePrefix` highlight.

In any case, thanks again for a great and very useful Neovim plugin!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## mini.indentscope
 
-- FEATURE: Add 'prefix' option to specify a prefix character in place of virtual text hard-coded space. Empty string value provides the ability to use this module with 'listchar' options ('lead', 'tab' and 'leadmultispace') without overwriting user-selected characters.
+- FEATURE: Add 'prefix' option to specify a prefix character in place of virtual text hard-coded space. Empty string value provides the ability to use this module with 'listchars' option ('lead', 'tab' and 'leadmultispace') without overwriting user-selected characters.
 
 # Version 0.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## mini.indentscope
 
-- FEATURE: Add 'prefix' option to specify a prefix character in place of virtual text hard-coded space. It provides the ability to use this module with 'listchars=lead:<char>' option without overwriting user-selected character with spaces.
+- FEATURE: Add 'prefix' option to specify a prefix character in place of virtual text hard-coded space. Empty string value provides the ability to use this module with 'listchar' options ('lead', 'tab' and 'leadmultispace') without overwriting user-selected characters.
 
 # Version 0.6.0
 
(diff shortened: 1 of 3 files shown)


echasnovski:
Hi! Thanks for the PR!

This is an interesting idea. Seems like related to #125.

Would you mind sharing relevant config parts with which you achieved the second screenshot?

mivort:
@echasnovski Sure, here's options I've used:
```lua
vim.o.listchars = 'lead:┆'
require('mini.indentscope').setup { prefix = '┆' }
```

mivort:
I think I found a way to handle nicely all possible `listchar`s combinations. The idea is to support an 'empty' value, in which case prefix won't be prepended, but rather `virt_text_win_col` increased to the proper amount. That way prefix will be 'transparent' and will not overwrite any listchars.

Here how it looks with `prefix = ''` and `vim.o.listchars = 'leadmultispace:*-'`:

leading

I've updated PR and added a remark about empty string value case handling in the documentation.

echasnovski:
I am a little bit confused right now... Using `virt_text_win_col` option seems to solve the issue quite nicely even without the new `prefix` option. I thought maybe it was added in Neovim 0.8, but no, it is present in Neovim 0.6.1. And all tests seem to (virtually) pass...

I'll need to investigate this further. Maybe your PR will indirectly lead to removing some code instead of adding one, which is even better :)

mivort:
Without at least some option it will make `MiniIndentscopePrefix` redundant and it wouldn't be possible to have something like this (e.g. provide a significantly more visually pronounced indent, which is nice):
``` lua
require('mini.indentscope').setup { symbol = ' ', prefix = ' ' }
vim.cmd [[hi MiniIndentscopeSymbol ctermbg=203 | hi MiniIndentscopePrefix ctermbg=210]]
```
image

So I think keeping this option may provide a benefit (and also allows to keep the old behavior, since the default value is `' '`).

echasnovski:
> Without at least some option it will make `MiniIndentscopePrefix` redundant and it wouldn't be possible to have something like this (e.g. provide a significantly more visually pronounced indent, which is nice):

True. But the only reason for `MiniIndentscopePrefix` was to create "invisible" highlight group to mitigate the (what I thought was) a necessary approach with prefix. At the moment I am leaning towards deprecating this highlight group, but I'll think it through.

echasnovski:
In the end, I decided to make a breaking change of removing `MiniIndentscopePrefix` highlight group in favor of always showing single line.

Thanks for making PR and showing `virt_text_win_col` capabilities!

PR #179, opened 2022-12-24
Added border config for info and signature window

mainendra:

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -242,6 +242,12 @@ MiniCompletion.config = {
     signature = { height = 25, width = 80 },
   },
 
+  -- Floating window border
+  window_border = {
+    info = 'none',
+    signature = 'none',
+  },
+
   -- Way of how module does LSP completion
   lsp_completion = {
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for making a PR!

Initially I intentionally didn't add this customization because it added spaces for a little gain (in my opinion). But recently I changed opinions about aesthetic side of Neovim. So I'll take time thinking if this should be added and maybe repurpose `config.window_dimensions` into `config.window_config` to allow further customization (like `winblend` or any future ones).

mainendra:
Yeah, I thought of using window_config but don't  want to add any breaking changes. ☺️

echasnovski:
@mainendra, thanks for this PR!

I ended up making breaking change of renaming `config.window_dimensions` to simply `window` and adding both `config.window.info.border` and `config.window.signature.border` (with you as co-author). Should be no in `main`.

(Side note: separate thank you for bringing up window borders. 'mini.base16' default highlight group for borders makes them not show border symbols (as per original base16), so I didn't really look into customizing them. No test-driving some options, and I really like them!)

PR #183, opened 2022-12-28
[mini.starter] `stopinsert` instead of `normal! <Esc>` to force normal mode

treywood:
For whatever reason, `normal! <ESC>` triggers the `InsertEnter` event, which interferes with some autocommands that I would like not to tigger on the start page. An alternative way to force normal mode is the `stopinsert` command:

```help
:stopi[nsert]		Stop Insert mode or |Terminal-mode| as soon as
			possible.  Works like typing <Esc> in Insert mode.
			Can be used in an autocommand, example:  
				:au BufEnter scratch stopinsert

```

### DIFF
--- a/lua/mini/starter.lua
+++ b/lua/mini/starter.lua
@@ -1280,7 +1280,7 @@ end
 H.apply_buffer_options = function(buf_id)
   -- NOTE: assumed that it is executing with `buf_id` being current buffer
   -- Force Normal mode
-  vim.cmd('stopinsert')
+  vim.cmd('normal! <C-\\><C-n>')
 
   -- Set buffer name
   H.buffer_number = H.buffer_number + 1
--- a/lua/mini/starter.lua
+++ b/lua/mini/starter.lua
@@ -1280,7 +1280,7 @@ end
 H.apply_buffer_options = function(buf_id)
   -- NOTE: assumed that it is executing with `buf_id` being current buffer
   -- Force Normal mode
-  vim.cmd('normal! <ESC>')
+  vim.cmd('stopinsert')
 
   -- Set buffer name
   H.buffer_number = H.buffer_number + 1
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for PR!

`stopinsert` only exits from Insert and Terminal mode (as documented). There are other modes which, namely Visual mode. I think `'normal! <C-\><C-n>'` should be a better choice. I'll look into it.

treywood:
> Thanks for PR!
> 
> `stopinsert` only exits from Insert and Terminal mode (as documented). There are other modes, like Visual mode. I think `normal! <C-\><C-n>` should be a better choice. I'll look into it.

Ah good catch. I tested with `normal! <C-\><C-n>` like you suggest and it seems to do the trick - the `InsertEnter` event isn't fired on start.

treywood:
Actually I revisited this. I wrote some tests to verify that `MiniStarter.open()` forces normal mode when executed from various other modes, and `stopinsert` is the only thing that seems to work for all the modes I could think of and also doesn't trigger `InsertEnter`

echasnovski:
> ... and `stopinsert` is the only thing that seems to work for all the modes I could think of and also doesn't trigger `InsertEnter`

Almost :) It doesn't seem to work in Command mode. Based on documentation, `<C-\><C-m>` is *the* way to ensure Normal mode ("Additionally the command CTRL-\ CTRL-N or <C-\><C-N> can be used to go to Normal mode from any other mode.  This can be used to make sure Vim is in Normal mode, without causing a beep like <Esc> would.").

@treywood, I've just pushed [starter-ensure-normal](https://github.com/echasnovski/mini.nvim/tree/starter-ensure-normal) branch. It has passing test for not triggering `InsertEnter`, but could you test it on your side, please?

treywood:
Ah forgot about command mode! 🤦🏻 

Your branch looks good, I tried it out with my config and it works correctly (no `InsertEnter` event triggered on start)

echasnovski:
Thanks! Should be in `main` now.

PR #198, opened 2023-01-15
(mini.misc) Add support for restoring the cursor

cryptomilk:
## What it does:
* When reopening a file, this will make sure the cursor is placed back to the position where you left before.
* This implements `:help restore-cursor` in a nicer way.
* Creates autocommand for |BufReadPre| and |FileType| event with |MiniMisc.restore_cursor|

This is still work in progress. However I wanted to ask if you would be fine integrating this.


echasnovski:
Hi!

First of all, thanks for taking time to write this. It would be even better if you created an issue to discuss it first (as said in [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)).

Second, could you clarify the purpose and scope of this proposed module? Is it a slightly polished version of the autocommand from `:h last-position-jump` (like open fold, etc.)?

Third, it is a strong opinion of mine that no `ignore_buftype` or `ignore_filetype` will be used as options of 'mini.nvim' modules. It gets complicated to track and not easy to make future proof.

cryptomilk:
I thought before opening an issue I could also create a MR. I wanted to play with mini.nvim anyway :-) So this was an exercise learning the code ...

About the plugin:

By default if you open a file with neovim, the cursor is positioned always at line 1 column 1 (1,1).

If you have the plugin loaded and lets assume you jump to line 60 and move the cursor to column 5 (60, 5). Now you close nvim and open the file again with nvim. With the plugin loaded the cursor will be positioned back on line 60 and column 5 (60, 5). It is last position were you left the file.

The information about the position where you left a file is stored in shada (see `:help shada`)!

The plugin is an enhanced version of the `:h last-position-jump`. I find this extremely useful. Try it.

echasnovski:
> I thought before opening an issue I could also create a MR. I wanted to play with mini.nvim anyway :-) So this was an exercise learning the code ...

Sure, I can understand this motivation :) Would be nice following the guidelines, though.

> By default if you open a file with neovim, the cursor is positioned always at line 1 column 1 (1,1).
> 
> If you have the plugin loaded and lets assume you jump to line 60 and move the cursor to column 5 (60, 5). Now you close nvim and open the file again with nvim. With the plugin loaded the cursor will be positioned back on line 60 and column 5 (60, 5). It is last position were you left the file.
> 
> The information about the position where you left a file is stored in shada (see `:help shada`)!

OK, I see. Having in mind comments about no `ignore_*` options in modules and the common practice of trying to always open just enough folds when moving cursor, it seems too small of a functionality to be a whole module.

I am not opposed to having this as a part of 'mini.misc', though. Along the lines of [`MiniMisc.setup_auto_root()` and `MiniMisc.find_root()`](https://github.com/echasnovski/mini.nvim/commit/bd6d1f3fd2da7741140510c2b37e2de7901b03fb). Say `setup_goto_last_cursor()` and `goto_last_cursor()`. Might be tricky to write tests, though.

> The plugin is an enhanced version of the `:h last-position-jump`. I find this extremely useful. Try it.

I now understand why I've never found this functionality useful... I am using sessions 99% of the time, and they restore this information automatically.

cryptomilk:
You really do not want the 'cursor goto' for `git commit` or the quickfix window. You want to disable it for some buffer/file types. If you do not want a maintain a list, the variables could be empty. However a user might want to be able to set those. It is really annoying if you do a `git commit` and the cursor is in the middle of the comment section. Is it ok to have the options cursor_goto_ignore_filetype but an emtpy table as the default? Then the user could maintain the list ...

echasnovski:
> You really do not want the 'cursor goto' for `git commit` or the quickfix window. You want to disable it for some buffer/file types. If you do not want a maintain a list, the variables could be empty. However a user might want to be able to set those. It is really annoying if you do a `git commit` and the cursor is in the middle of the comment section. Is it ok to have the options cursor_goto_ignore_filetype but an emtpy table as the default? Then the user could maintain the list ...

If we are talking about a module - no, any filetype/buftype related disabling is done only through `vim.b.minixxx_disable` variables. User can then create autocommands to automatically set them up (possibly using a provided common recipes). It is also easier to implement.

If we are talking about 'mini.misc' `setup_*` type of function, then I think these lists can be an optional arguments.

cryptomilk:
I've added it to MiniMisc.setup() but I can move it to `setup_goto_last_cursor()`

lua/mini/misc.lua
```diff
@@ -44,6 +44,10 @@ end
 MiniMisc.config = {
   -- Array of fields to make global (to be used as independent variables)
   make_global = { 'put', 'put_text' },
+  cursor_goto = {
```
echasnovski, line 47: Let's not add it to setup.
It is better suited as fields of `opts` argument for `setup_goto_last_cursor()` and `goto_last_cursor()`.

lua/mini/misc.lua
```diff
@@ -409,6 +413,49 @@ MiniMisc.zoom = function(buf_id, config)
   end
 end
 
+---
+MiniMisc.setup_goto_last_cursor = function()
+  if vim.fn.has("nvim-0.7.0") then
```
echasnovski, line 418: Let's use only one version, which supports Neovim<0.7. No need to do same thing twice.

When deprecating Neovim<0.7, I'll replace all `vim.cmd()` calls with Lua autocommand API.

lua/mini/misc.lua
```diff
@@ -409,6 +413,49 @@ MiniMisc.zoom = function(buf_id, config)
...
+
+    vim.api.nvim_create_augroup(group_name, { clear = true })
+    vim.api.nvim_create_autocmd("BufWinEnter", { callback = MiniMisc.goto_last_cursor, group = group_name })
+    vim.api.nvim_create_autocmd("FileType", { callback = MiniMisc.goto_last_cursor, group = group_name })
+  else
+    vim.api.nvim_command([[augroup MiniMiscGotoLastCursor]])
+    vim.api.nvim_command([[  autocmd!]])
+    vim.api.nvim_command([[  autocmd BufWinEnter * lua require('mini.misc').goto_last_cursor()]])
+    vim.api.nvim_command([[  autocmd FileType * lua require('mini.misc').goto_last_cursor()]])
+    vim.api.nvim_command([[augroup end]])
```
echasnovski, line 429: You can do all this in a single `vim.api.nvim_exec([[...]], false)` call.

Is there a reason to use `BufWinEnter` event? It means cursor will be tried to position every time buffer is displayed in a window.

Is there a reason for not following an example from `:h last-position-jump`?

lua/mini/misc.lua
```diff
@@ -409,6 +413,49 @@ MiniMisc.zoom = function(buf_id, config)
...
+      return
+  end
+
+  -- If a line has already been specified on the command line, we are done
+  --   nvim file +num
+  if vim.fn.line(".") > 1 then
+      return
+  end
+
+  H.cursor_jump()
```
echasnovski, line 456: I think the whole `H.cursor_jump()` can be replaced with `vim.cmd([[normal! g'"zvzz]])`. Not sure if all those condition are vital. Was there a reason why you added them?

lua/mini/misc.lua
```diff
@@ -439,6 +486,15 @@ H.setup_config = function(config)
...
 
+  vim.validate({
+    cursor_goto = { config.cursor_goto, 'table' },
+  })
+
+  vim.validate({
+    ['cursor_goto.ignore_buftype'] = { config.cursor_goto.ignore_buftype, 'table' },
+    ['cursor_goto.ignore_filetype'] = { config.cursor_goto.ignore_filetype, 'table' },
+  })
+
```
echasnovski, line 497: Not needed.

lua/mini/misc.lua
```diff
@@ -450,6 +506,10 @@ H.apply_config = function(config)
   end
 end
 
+H.get_config = function(config)
+  return vim.tbl_deep_extend('force', MiniMisc.config, vim.b.minimisc or {}, config or {})
+end
+
```
echasnovski, line 512: Not needed.

lua/mini/misc.lua
```diff
@@ -465,4 +525,29 @@ H.is_number = function(x) return type(x) == 'number' end
 
 H.is_string = function(x) return type(x) == 'string' end
 
+H.cursor_jump = function()
```
echasnovski, line 528: Possibly also not needed. See comment in `goto_last_cursor()`.

echasnovski (changes requested):

lua/mini/misc.lua
cryptomilk: Done. As you already use nvim_create_autocmd in setup_auto_root I used that

echasnovski: > Done. As you already use nvim_create_autocmd in setup_auto_root I used that

`nvim_create_autocmd()` in `setup_auto_root()` was used because it already relies on Neovim>=0.8. This new function doesn't need to rely on that version and can support Neovim>=0.6. Let's do a more reverse compatible version for now.

lua/mini/misc.lua
cryptomilk: They do not center all the time, only if poosible. If you're at the last line, you don't want to center, you want it to be the last line.

lua/mini/misc.lua
cryptomilk: Done

lua/mini/misc.lua
cryptomilk: Done.

lua/mini/misc.lua
cryptomilk: Done.

cryptomilk:
I've fixed the last issues, added more code comments and improved documentation.

lua/mini/misc.lua
echasnovski: > They do not center all the time, only if possible. If you're at the last line, you don't want to center, you want it to be the last line.

This is debatable. In my mind, this functionality serves two purposes:
- Reduce friction for user to move cursor to its latest place - cursor is already there.
- Reduce friction for user to search visually for cursor - it is in the center of screen and outside of folds.

The second point is not valid under current implementation. I would get it if it would require more work/code to achieve "always centered" behavior, but it is the opposite.

lua/mini/misc.lua
echasnovski: > I've tried to do a one line autcomd but couldn't get it working. This worked just fine. As it is really new functionality why not used the latest API if misc already uses that=

Only `setup_auto_root()` and `find_root()` depend on having `vim.fs` (the same as depending on Neovim>=0.8). All other 'mini.misc' functions support Neovim>=0.6.

lua/mini/misc.lua
```diff
@@ -409,6 +409,102 @@ MiniMisc.zoom = function(buf_id, config)
...
+---   {
+---     ignore_buftype = { "quickfix", "nofile", "help" },
+---     ignore_filetype = { "gitcommit", "gitrebase" },
+---   }
+---   ```
+---   Default: `{}`.
+---
+---@usage >
+---   require('mini.misc').setup_restore_cursor()
+MiniMisc.setup_restore_cursor = function(opts)
```
echasnovski, line 437: Let's follow this as general design:
- `setup_restore_cursor()`:
    - Receives `opts` with at least two possible fields `ignore_buftype` and `ignore_filetype`. Their defaults are as in example (not empty). Use the following approach:
     ```lua
    if not H.is_array_of(x, H.is_string) then
      H.error('In `setup_restore_cursor()` `opts.x` should be an array of strings.')
    end
    ``` 
    - Validates `opts` and ensures defaults are in place.
    - Creates Neovim<0.7 compatible autocommand following `:h restore-cursor`. I think something like this should work:
    ```lua
    local au_command = string.format(
      [[augroup MiniMiscRestoreCursor
          au!
          au BufRead * au FileType <buffer> ++once lua MiniMisc.restore_cursor(%s)
        augroup END]],
      vim.inspect(opts)
    )
    vim.api.nvim_exec(au_command, false)
    ```
- `restore_cursor()`:
    - Also takes `opts` as described earlier, but doesn't validate it.
    - Makes all necessary checks that prohibit restoring cursor.
    - Restores cursor with `vim.cmd([[normal! g'"zvzz]])`.

lua/mini/misc.lua
```diff
@@ -450,6 +546,34 @@ H.apply_config = function(config)
   end
 end
 
+-- Cursor goto default opts
```
echasnovski, line 549: All this should not be needed under design proposed in first comment.

echasnovski (changes requested):

lua/mini/misc.lua
cryptomilk: It is quite strange if you have a 50 lines window, the file has 30 lines and then you center it. Then you don't see the 5 lines on the top and have to scroll.

This is what the first avoids. Does also not center because then you would have half of a blank window.

Currently it shows you always the maximum of the file content.

lua/mini/misc.lua
cryptomilk: The au_command doesn't work:

`misc.lua:465: Vim:E492: Not an editor command:   ignore_buftype = { "quickfix", "nofile", "help" },`

echasnovski: > The au_command doesn't work:
> 
> `misc.lua:465: Vim:E492: Not an editor command: ignore_buftype = { "quickfix", "nofile", "help" },`

Ah, yes. Try replacing `vim.inspect(opts)` with `vim.inspect(opts, { newline = ' ', indent = '' })`.

lua/mini/misc.lua
echasnovski: And yet, always having cursor centered reduces both user surprise (expecting cursor at center) and code for implementation/testing (quite significantly).

lua/mini/misc.lua
echasnovski: Interesting...

Still, current logic of triggering of every `BufRead` and `FileType` is not ideal. For ignored filetypes it means moving cursor twice (to restore and then back to first line).

Further investigation why documented solution doesn't work (when on surface it should) seems like the best course of action here. It can work in Vim (from where it is ported) but not in Neovim, or in both.

lua/mini/misc.lua
```diff
@@ -409,6 +409,130 @@ MiniMisc.zoom = function(buf_id, config)
...
+---   }
+---   ```
+---
+---@usage >
+---   require('mini.misc').setup_restore_cursor()
+MiniMisc.setup_restore_cursor = function(opts)
+  opts = opts or {}
+  opts.ignore_buftype = opts.ignore_buftype or { "quickfix", "nofile", "help" }
+  opts.ignore_filetype = opts.ignore_filetype or { "gitcommit", "gitrebase" }
+  opts.always_center = opts.always_center or true
```
echasnovski, line 441: Center always with `g'"zvzz`, not behind an option.

lua/mini/misc.lua
cryptomilk: It works fine if you use `nvim_create_autocmd`.

lua/mini/misc.lua
cryptomilk: This is the default. We will always center. However I do not want that and also maybe other who used to the behavior of the vim plugin. So you can disable it with that option now.

lua/mini/misc.lua
echasnovski: > It works fine if you use `nvim_create_autocmd`.

With current implementation (triggers on both `BufRead` and `FileType`) or if rewritten to create a `FileType` autocommand on `BufRead`? If former, then see previous comment on it being not ideal. If latter, then it is some weird behavior.

lua/mini/misc.lua
echasnovski: I am fine with `center` option which decided whether to do a simple `zz` or not.

All current if-else-s with their different behavior add too much complexity for little gain. They won't be a part of 'mini.misc'.

cryptomilk: Fixed.

lua/mini/misc.lua
cryptomilk: I've fixed it.

lua/mini/misc.lua
cryptomilk: Always center is the default and you can disable it now.

lua/mini/misc.lua
```diff
@@ -409,6 +409,139 @@ MiniMisc.zoom = function(buf_id, config)
...
+  -- If a line has already been specified on the command line, we are done
+  --   nvim file +num
+  if vim.fn.line(".") > 1 then
+      return
+  end
+
+  local last_line = vim.fn.line([['"]])
+  local buff_last_line = vim.fn.line("$")
+
+  if last_line > 0 and last_line <= buff_last_line then
```
echasnovski, line 531: This can be a separate early return block.

Is there a reason to use `last_line > 0` instead of `last_line > 1` as in docs? The latter seems to ensure that cursor is not restored to the same place it is at right now.

lua/mini/misc.lua
```diff
@@ -409,6 +409,139 @@ MiniMisc.zoom = function(buf_id, config)
...
+    else
+      vim.api.nvim_command([[normal! g`"]])
+    end
+  end
+
+  -- Open fold
+  if vim.fn.foldclosed(".") ~= -1 then
+    vim.api.nvim_command([[normal! zvzz]])
+  end
+end
```
echasnovski, line 543: Maybe all these can be replaced with this?

```lua
vim.cmd('normal! g`"zv')
if opt.center then vim.cmd('normal! zz') end
```

Doing `zv` inside a non-fold seems harmless from experience.

Also, do you have any bad experience with using `` g`" `` instead of `g'"`? I quickly tried couple of days ago the first one and it sometimes gave some weird behavior of off-by-one placement.

lua/mini/misc.lua
```diff
@@ -409,6 +409,139 @@ MiniMisc.zoom = function(buf_id, config)
...
+MiniMisc.restore_cursor = function(opts)
+  -- Check if the buffer should be ignored
+  if vim.tbl_contains(opts.ignore_buftype,
+                      vim.api.nvim_buf_get_option(0, "buftype")) then
+    return
+  end
+
+  -- Check if the filetype should be ignored
+  if vim.tbl_contains(opts.ignore_filetype,
+                      vim.api.nvim_buf_get_option(0, "filetype")) then
```
echasnovski, line 514: Can these use `vim.bo.buftype` and `vim.bo.filetype`?

echasnovski (changes requested):
This starting to look good!
There is only some code polishing and making sure that autocommands work as expected. And then testing can start.

lua/mini/misc.lua
cryptomilk: You're right, and early return is nicer.

Fixed.

lua/mini/misc.lua
cryptomilk: Yep, good idea.

lua/mini/misc.lua
cryptomilk: > Maybe all these can be replaced with this?
> 
> ```lua
> vim.cmd('normal! g`"zv')
> if opt.center then vim.cmd('normal! zz') end
> ```
> 
> Doing `zv` inside a non-fold seems harmless from experience.

I just tried and sometimes the cursor is placed somewhere else if you don't have any folds. However I made the code a bit nicer.

> Also, do you have any bad experience with using `` g`" `` instead of `g'"`? I quickly tried couple of days ago the first one and it sometimes gave some weird behavior of off-by-one placement.

I don't have any problems with ``g`"``, it works just fine.

cryptomilk:
JFYI: https://github.com/neovim/neovim/issues/21923

lua/mini/misc.lua
echasnovski: > I just tried and sometimes the cursor is placed somewhere else if you don't have any folds. However I made the code a bit nicer.

Hmm... I added these lines and tried to open several files:

```lua
_G.restore_cursor = function() vim.cmd([[normal! g`"zvzz]]) end
vim.cmd([[autocmd BufRead * lua _G.restore_cursor()]])
```

Didn't have any issues with folds. Could you isolate the problem and make it reproducible? It is not only for this module, but in case there is an issue I should recheck other modules.

> I don't have any problems with ``g`"``, it works just fine.

Had one issue when opening file with 'telescope.nvim': it places cursor one column to the left where it is needed... Strange. But if it is only in Telescope, then no big deal.

lua/mini/misc.lua
```diff
@@ -409,6 +409,137 @@ MiniMisc.zoom = function(buf_id, config)
...
 end
 
+--- Restore the cursor on file reopen
+---
+--- What it does:
+--- - When reopening a file, this will make sure the cursor is placed back to
+---   the position where you left before.
+--- - This implements :help restore-cursor in a nicer way.
+--- - Creates autocommand for |BufReadPre| and |FileType| event with
+---   |MiniMisc.restore_cursor|
```
echasnovski, line 419: Merge, please, two first points in a single one. No need for implementation details about autocommands.

lua/mini/misc.lua
```diff
@@ -409,6 +409,137 @@ MiniMisc.zoom = function(buf_id, config)
...
+--    is in the middle of the window. If you set it to `false` centering is
+--    turned off.
+---   Default:
+---   ```
+---   {
+---     ignore_buftype = { "quickfix", "nofile", "help" },
+---     ignore_filetype = { "gitcommit", "gitrebase" },
+---     center = true,
+---   }
+---   ```
```
echasnovski, line 439: Could you, please, make it more like other `opts` documentations. Use [this](https://github.com/echasnovski/mini.nvim/blob/91017a96693408ef96efe9a2513c6ace0a87dc8d/lua/mini/ai.lua#L515-L524) as example.

lua/mini/misc.lua
cryptomilk: Done.

lua/mini/misc.lua
cryptomilk: > Had one issue when opening file with 'telescope.nvim': it places cursor one column to the left where it is needed... Strange. But if it is only in Telescope, then no big deal.

I can reproduce it with telescope. However using ``g'"`` doesn't fix it for me. It always moves the cursor. Might be a telescope bug.

cryptomilk:
> Yep, overall this looks nice already. Thanks!
> 
> Two major points to address:
> 
>     * Implement autocommand in the most clean way possible. This means it should have both of these:

Done. Also in lua code without vimscript.

>       * For each buffer `restore_cursor()` should be called exactly once. Based on the `ignore_filetype` option, should be not earlier than `FileType` event. I would really suggest digging into the official documented solution: if it is wrong, suggest a fix upstream (Neovim or even Vim). One more good deed :) If to no avail, maybe can be implemented inside `FileType` event but stopping early if this buffer was already processed (keep track in some `H.restore_cursor_buffers = {}`).

This is also working with lua code, which is commented out, but I tested it.

>       * Support Neovim<0.7. So should be created using Vimscript with `vim.cmd()`.

I've asked. vim.api.nvim_exec(cmd, false) should be used. This is what I'm using now.

>     * Write tests. They should test at least:
>       
>       * General simple work. As indicator that there is at least a basic functionality with default options.
>       * _Every_ `if-else` block for early stop.
>       * _Every_ field of `opts` is respected.

Is there an explanation somewhere how the tests work? Some things are not clear for me.


```
T['restore_cursor_setup()'] = new_set()
```

will this call the `restore_cursor_setup` function form the `MiniMisc` module already or is it called for each test I create then?

I guess nvim is executed and then the function called in it somehow. Can I specify commandline options? I need to create a shada file and copy a test file to /tmp/ with a fixed name, so shada can find the filename and set the cursor.

echasnovski:
> Done. Also in lua code without vimscript.

I might have missed something. So the `au BufReadPre * au FileType <buffer> ++once ...` solution works? If yes, then there is no need for `vim.api.nvim_exec([[normal! gg]], false)` for ignored filetypes.

I also was wondering what were the comments about. I appreciate them, but please remove them, as they are not needed. I'll convert `nvim_exec(...)` solution myself when support for Neovim<0.7 is dropped.

> Is there an explanation somewhere how the tests work?

See [TESTING.md](https://github.com/echasnovski/mini.nvim/blob/main/TESTING.md).

`T` is a [main table](https://github.com/echasnovski/mini.nvim/blob/91017a96693408ef96efe9a2513c6ace0a87dc8d/tests/test_misc.lua#L23-L31) with test cases in form of callbacks. Each callback sets up fresh Neovim process inside `child` and loads `mini.misc` with defaults. After that, each test case usually sets some conditions, executes something, and checks if it performed as expected.

> I guess nvim is executed and then the function called in it somehow. Can I specify commandline options? I need to create a shada file and copy a test file to /tmp/ with a fixed name, so shada can find the filename and set the cursor.

I think using an approach of modeling creation of shada file is better here. Something like "open file" -> position cursor -> `:wshada` in some predetermined file (which will have to be cleaned up) -> `:bwipeout` buffer -> `:rshada!` -> reopen buffer? Might have some pitfalls, not sure.

cryptomilk:
> I might have missed something. So the `au BufReadPre * au FileType <buffer> ++once ...` solution works? If yes, then there is no need for `vim.api.nvim_exec([[normal! gg]], false)` for ignored filetypes.

Yes, that can be dropped now.
 
> I also was wondering what were the comments about. I appreciate them, but please remove them, as they are not needed. I'll convert `nvim_exec(...)` solution myself when support for Neovim<0.7 is dropped.

I can delete the comments for the lua implementation with `vim.api.nvim_create_autocmd` if you want. I took me a while to figure out how to actually do that. I can remove it and keep it in another branch to not loose it. Or you you want to spend quite some time figuring this out again in a few month? :-)

cryptomilk:
Lua implementation of setting up the autocommands: https://github.com/cryptomilk/mini.nvim/tree/mini.restore_cursor_lua

cryptomilk:
Hmm, I create a temporary shada filename and use `:wshada! <file>` to write it. However the shada file doesn't get written. It works fine if I open a `nvim --no-plugin -u NONE`.

echasnovski:
> Hmm, I create a temporary shada filename and use `:wshada! <file>` to write it. However the shada file doesn't get written. It works fine if I open a normal `nvim --no-plugin -u NONE` session.

Yes. Some investigation was needed, but it is because child process in tests are [created with `--clean` flag](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/lua/mini/test.lua#L1077) . In turn, this means that 'shadafile' option is set to "NONE" disabling all shada read and write. Use `child.o.shadafile = ''` and `:wshada!` should work.

While I am here:
- Use existing 'dir-misc', not a new 'dir-cursor' one.
- Do not add '.editorconfig'.
- Use predetermined file path for shada file and clean it up after every case inside of `post_case` hook.

cryptomilk:
> Yes. Some investigation was needed, but it is because child process in tests are [created with `--clean` flag](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/lua/mini/test.lua#L1077) . In turn, this means that 'shadafile' option is set to "NONE" disabling all shada read and write. Use `child.o.shadafile = ''` and `:wshada!` should work.

That works.
 
> While I am here:
> 
>     * Use existing 'dir-misc', not a new 'dir-cursor' one.

Done.

>     * Do not add '.editorconfig'.

I will remove it once everything is ready. However it would be nice to have, also nvim will get support for it in 0.9.

>     * Use predetermined file path for shada file and clean it up after every case inside of `post_case` hook.

I already do.

More tests to follow ...

echasnovski:
> I already do.

You use `os.tmpname()`, which is not predetermined. With this approach it is easy to not detect problems with cleanup.

Please, use approach similar to 'test_sessions.lua': [this](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/tests/test_sessions.lua#L14-L16), [this](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/tests/test_sessions.lua#L33-L35), and [this](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/tests/test_sessions.lua#L129-L158) (but inside `new_set()` for restoring cursor).

cryptomilk:
Looks like we can't test the `center` option. I don't find a way to get the cursor postion in the window.

`vim.api.nvim_win_get_position()` gives you the buffer-relative postion, now the window :-(

cryptomilk:
> > I already do.
> 
> You use `os.tmpname()`, which is not predetermined. With this approach it is easy to not detect problems with cleanup.
> 
> Please, use approach similar to 'test_sessions.lua': [this](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/tests/test_sessions.lua#L14-L16), [this](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/tests/test_sessions.lua#L33-L35), and [this](https://github.com/echasnovski/mini.nvim/blob/1c11d5fdb71c1d766545e42b0dd99eda50f1e349/tests/test_sessions.lua#L129-L158) (but inside `new_set()` for restoring cursor).

Done. Thanks for the pointer :-)

cryptomilk:
I think, I have covered everything with tests now.

tests/dir-misc/cursor_restore.lua
```diff
@@ -0,0 +1,61 @@
+test1 = function()
```
echasnovski, line 1: You certainly shouldn't need this large file for test purposes. Even to test centering.

tests/test_misc.lua
```diff
@@ -1,7 +1,7 @@
 local helpers = dofile('tests/helpers.lua')
 
 local child = helpers.new_child_neovim()
-local expect, eq = helpers.expect, helpers.expect.equality
+local expect, eq, noeq = helpers.expect, helpers.expect.equality, helpers.expect.no_equality
```
echasnovski, line 4: If you use it once, no need to make a helper.

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+local restore_cursor_default_opts = {
+  ignore_buftype = { 'nofile' },
+  ignore_filetype = { 'gitcommit' },
+  center = true,
+}
+
+T['setup_restore_cursor()']['autocmd'] = function()
+  setup_restore_cursor()
+
+  eq(child.lua_get([[#vim.api.nvim_get_autocmds({ group = 'MiniMiscRestoreCursor' })]]) > 0, true)
```
echasnovski, line 558: This will break on Neovim<0.7. Use `eq(child.fn.exists('#MiniMiscRestoreCursor'), 1)`

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
   child.expect_screenshot()
 end
 
+T['setup_restore_cursor()'] = new_set()
+
+local setup_restore_cursor = function(...) child.lua('MiniMisc.setup_restore_cursor(...)', { ... }) end
+
+local restore_cursor_default_opts = {
```
echasnovski, line 549: Most usage of `restore_cursor_default_opts` seems to be to indicate that `restore_cursor()` is called with default options. Cleaner approach is to call it without arguments at all.

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+      cursor_cleanup()
+
+      -- Make all showed messages full width
+      child.o.cmdheight = 10
+
+      -- Load module
+      load_module()
+
+      -- Ensure clear message history
+      child.cmd('messages clear')
```
echasnovski, line 603: Only `cursor_cleanup()` here should be enough.

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+    end,
+    post_once = function()
+      cursor_cleanup()
+      child.stop()
+    end,
```
echasnovski, line 608: This seems to be not needed here. This is a subset of `T`, where child process needs to be stopped.

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+})
+
+local restore_cursor = function(...) child.lua('MiniMisc.restore_cursor(...)', { ... }) end
+local test_file_cursor = make_abspath('tests/dir-misc/cursor_restore.lua')
+
+T['restore_cursor()']['opts'] = function()
+  local opts = restore_cursor_default_opts
+
+  restore_cursor(opts)
+end
```
echasnovski, line 619: What is the purpose of this test?

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+
+local cursor_jump_line = function(line)
+  child.cmd(string.format('%u', line))
+  eq(child.fn.line('.'), line)
+end
```
echasnovski, line 624: This is not necessary.

Make `set_cursor()` and `get_cursor()` helpers (as in most other test files) and use them directly in test cases.

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+
+local write_shada_file = function()
+  child.fn.mkdir(cursor_dir_path)
+  child.o.shadafile = ''
+  child.cmd('wshada! ' .. cursor_shada_file)
+  -- verify that shada file has been written
+  local stat = vim.loop.fs_stat(cursor_shada_file)
+  eq(stat.type, 'file')
+  eq(stat.size > 1, true)
```
echasnovski, line 633: This is more helpful during debugging, not actual tests. Should be safe to remove.

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+end
+
+local read_shada_file = function()
+  -- verify that shada file exists
+  local stat = vim.loop.fs_stat(cursor_shada_file)
+  eq(stat.type, 'file')
+  eq(stat.size > 1, true)
```
echasnovski, line 640: This is not necessary. Even if it is, it can be replaced with `eq(child.fn.filereadable(cursor_shada_file), 1)`.

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+  child.cmd('edit ' .. test_file_cursor)
+
+  local jump_line = 23
+  cursor_jump_line(jump_line)
+  write_shada_file()
+  child.cmd('bdelete')
+
+  read_shada_file()
+  child.cmd('edit ' .. test_file_cursor)
+  eq(child.fn.line('.'), 1)
```
echasnovski, line 657: As implementation uses `` g`" ``, it is a good idea to check for column too. Use `eq(get_cursor(), { x, y })`

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+  child.cmd('quit')
+  read_shada_file()
+  child.cmd('help')
+  eq(child.fn.line('.'), 1)
+  restore_cursor(opts)
+  eq(child.fn.line('.'), jump_line)
+end
+
+T['restore_cursor()']['ignore_buftype'] = function()
+  local opts = {
```
echasnovski, line 680: No need for this helper variable. Use explicitly `restore_cursor({ ignore_buftype = { 'help' } })`

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+  child.cmd('bdelete')
+
+  read_shada_file()
+  child.cmd('edit ' .. test_file_cursor)
+  restore_cursor(opts)
+  eq(child.fn.line('.'), jump_line)
+
+  local win_height = child.api.nvim_win_get_height(0)
+  local win_cursor = child.fn.winline()
+  eq(math.floor(win_height / 2 + 1), win_cursor)
```
echasnovski, line 760: More explicit and robust approach is to set set of child process in the test start (`child.set_size(10, 12)` for example) and then test for top window line (`eq(child.fn.line('w0'), x)`)

tests/test_misc.lua
```diff
@@ -541,4 +542,382 @@ T['zoom()']['respects `config` argument'] = function()
...
+  child.cmd('edit ' .. test_file_cursor)
+  eq(child.fn.line('.'), 1)
+  restore_cursor(opts)
+  eq(child.fn.line('.'), jump_line)
+  eq(child.fn.foldclosed('30'), -1)
+
+  local win_height = child.api.nvim_win_get_height(0)
+  local win_cursor = child.fn.winline()
+  eq(win_height, win_cursor)
+end
```
echasnovski, line 921: Are all of these test cases really necessary? On first glance, these cases should be tested:
- Correctly restores cursor when needed.
- Correctly not restores cursor when not needed (each case for every early return in implementation).
- Correctly opens fold (including correctly centering after, not before).
- Correctly centers when needed (both in the middle of the file and end of file).
- Correctly centers when not needed (`opts.center = false`).

lua/mini/misc.lua
```diff
@@ -409,6 +409,91 @@ MiniMisc.zoom = function(buf_id, config)
...
+  -- Check if the filetype should be ignored
+  if vim.tbl_contains(opts.ignore_filetype, vim.bo.filetype) then
+      return
+  end
+
+  -- If a line has already been specified on the command line, we are done
+  --   nvim file +num
+  if vim.fn.line('.') > 1 then
+      return
+  end
```
echasnovski, line 474: I am not entirely sure, but is there a test for this early return?

echasnovski (changes requested):
Sorry for many comments. This does look very good, only many small optimizations needed.

And general request: please use more explicit test case names describing what should work correctly. Like:
- Instead of `['nocenter']` - `` ['respects `opts.center`'] ``.
- Instead of `['file']` - `['works in general file']`.

tests/dir-misc/cursor_restore.lua
cryptomilk: Fixed.

lua/mini/misc.lua
cryptomilk: How do I restart the child with a file and line specified?

cryptomilk:
I don't see `auto_root()` documented in `readmes/mini-misc.md`. If you could do it I could follow it and document `restore_cursor()`.

tests/test_misc.lua
cryptomilk: Fixed.

echasnovski:
> I don't see `auto_root()` documented in `readmes/mini-misc.md`. If you could do it I could follow it and document `restore_cursor()`.

I'll probably add some kind of summary for all interesting 'mini.misc' functions when this one is done.

tests/test_misc.lua
cryptomilk: Will do, but this is copy & paste from the same file.

lua/mini/misc.lua
echasnovski: Try `child.restart({ '+2', '--', test_file_cursor })`

tests/test_misc.lua
cryptomilk: Fixed.

tests/test_misc.lua
cryptomilk: Fixed.

tests/test_misc.lua
echasnovski: > Will do, but this is copy & paste from the same file.

Yes, because there is doesn't run on Neovim<0.7 . But still, should probably change it there.

tests/test_misc.lua
cryptomilk: Fixed.

tests/test_misc.lua
cryptomilk: Dropped. Was just the first test to check if it works.

echasnovski: > Was just the first test to check if it works.

This is good idea and should be done. But this one didn't actually **test** anything.

tests/test_misc.lua
cryptomilk: Done.

tests/test_misc.lua
cryptomilk: Done.

tests/test_misc.lua
cryptomilk: I replaced it with `eq(child.fn.filereadable(cursor_shada_file), 1)`

tests/test_misc.lua
cryptomilk: I will check it once.

tests/test_misc.lua
cryptomilk: Ok.

lua/mini/misc.lua
cryptomilk: I get:

```
Error executing lua: [string "<nvim>"]:1: attempt to index global 'MiniMisc' (a nil value)
```

for `restore_cursor()`.

tests/test_misc.lua
cryptomilk: I've merged some tests.

lua/mini/misc.lua
echasnovski: Yes. It restarts child process with putting cursor on specific line. You also need to set up 'mini.misc'. Use `load_module()`.

tests/test_misc.lua
cryptomilk: Done.

lua/mini/misc.lua
cryptomilk: `load_module()` and specifying the config is needed to find the module. Works now.

tests/test_misc.lua
cryptomilk: Should be fine now.

echasnovski:
I've taken a quick look. Seems to be nothing major to change. Still some small-ish code style issues, but I think I'll fix them myself during merge. Thanks!

I'll take closer look, probably, the day after tomorrow.

cryptomilk:
Thanks for all the help!

echasnovski (approved):
Thank you for your work!

I'll merge this in separate branch and will polish the rest myself. Of course, you'll be credited as initial author.

echasnovski:
@cryptomilk, I've polished it and merged into `main`.

Couple of differences that I made:
- Removed `ignore_buftype` and made it work only for normal buffers. This seems to be a more reasonable approach. If not, can add those later.
- Decided to act `zv` always and not behind `foldclosed()`, as I didn't see a use case where this goes wrong.
- Rewritten tests to test `setup_restore_cursor()` directly, which required some tricky approach for restarting child process.
- Added note that it won't work if file doesn't have recognized file type. This is a limitation of `FileType` approach, but it doesn't seem to be a huge deal to me right now.

Thanks again for going forward with this!

cryptomilk:
@echasnovski You're welcome.

I've rebased my branch which implements the autocmd using vim.api.nvim_create_augroup() here:

https://github.com/cryptomilk/mini.nvim/tree/mini.restore_cursor_lua

PR #208, opened 2023-01-28
fix(mini.map): winblend description

edr3x:
### 100 makes it opaque:

### DIFF
--- a/doc/mini-map.txt
+++ b/doc/mini-map.txt
@@ -340,8 +340,8 @@ map line. Default: `true`.
 integration count column. Default: 10.
 
 `window.winblend` - value of 'winblend' for map window. Value 0 makes it
-completely non-transparent, 100 - completely transparent (content is still
-visible, but with slightly different highlights).
+completely transparent, 100 makes it completely opaque (content still will
+be visible on transparent mode, but with slightly different highlights).
 
 # Pure scrollbar config ~
 


echasnovski:
> ## DOC: Fix the reverse description of winblend on mini.map
> ### 0 makes it transparent:
> 
> ### 100 makes it opaque:
> 
>     * [x]  I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
> 
>     * [x]  I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

Thanks for PR!

However, the current documentation is correct. Documentation is taken almost verbatim [from corresponding 'winblend' option](https://neovim.io/doc/user/options.html#'winblend'). And [here](https://github.com/echasnovski/mini.nvim/blob/81a575e0c51d4607bf6690f12906c7590d961717/lua/mini/map.lua#L1389-L1390) where it is used directly.

Here is my experience with `winblend = 0`:

map_winblend-0

And this is with `winblend = 100`:

map_winblend-100

I am not sure why you experience this. Assuming no mixup with actual behavior, probably you have your `hi MiniMapNormal` highlight group with `guibg=NONE`. Together with transparent transparent terminal emulator it makes "transparent".

edr3x:
Strange, 
I get same behaviour even if i keep my terminal non transparent and with nvim colorscheme non transparent
maybe this is only for me, 
Well in the end it works for me on reverse, Thanks!!

echasnovski:
> Strange, I get same behaviour even if i keep my terminal non transparent and with nvim colorscheme non transparent maybe this is only for me, Well in the end it works for me on reverse, Thanks!!

Strange indeed.

Are you sure that 'winblend' is actually set to what you think it is set? If you execute `:lua print(MiniMap.config.window.winblend)`, does it show the value you expect it to?

If the answer is yes, what is the output of `:hi MiniMapNormal`?

edr3x:
Yes, doing `:lua print(MiniMap.config.window.winblend)` outputs 0 which is what i set 
> If the answer is yes, what is the output of :hi MiniMapNormal?

result

echasnovski:
Very strange... No clue at this point as to why you experience this. Good thing you can work around this.

If in future you detect a change in behavior (maybe after changing theme or terminal emulator), would you mind writing here about this?

edr3x:
> If in future you detect a change in behavior (maybe after changing theme or terminal emulator), would you mind writing here about this?
 
Currently i am using Tokyo-night color scheme and Alacritty terminal emulator.
I will write about this if the behavior changed..

PR #215, opened 2023-02-01
Annotate config setup tables as optional

tobealive:

Currently, not all mini modules have their config setup tables annotated as optional even though they have noted:
> -- No need to copy this inside `setup()`. Will be used automatically.

E.g.: https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-trailspace.md#installation

For those modules it results in a type error.

Of course a user can just pass an empty `{}` to silent the error.
But I feel like the right solution would be to annotate them correctly as optional.

Usually marking a param as optional is done by adding a question mark, `<type>?`. But I followed the convention in the modules that have marked it's config tables as optional by addling `|nil`. If this should changed, please let me know. I think it could be nicely done in one wash with this PR.

### DIFF
--- a/lua/mini/base16.lua
+++ b/lua/mini/base16.lua
@@ -143,7 +143,7 @@ local H = {}
 --- - Table with similar structure to `palette` but having terminal colors
 ---   (integers from 0 to 255) instead of hex strings.
 ---
----@param config table Module config table. See |MiniBase16.config|.
+---@param config table|nil Module config table. See |MiniBase16.config|.
 ---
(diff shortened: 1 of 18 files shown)


echasnovski:
Thank you for PR!

Yeah, I've been meaning to do this for while. Not only in `setup()` but also for all functions which have default values. I'll try to make time for this in the near future.

echasnovski:
This should be in `main` now. Along with the massive update of all annotations. Thanks for the push!

Issue #146, opened 2022-10-20
Consistent `verbose` feature


echasnovski:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

all

### Description

Consider implementing `verbose` option for all modules which will control emitted helper messages. Like helper message in 'mini.ai'/'mini.surround' with information to press a key or that no object is found.

Also consider whether it should be option or `vim.{b,g}.mini***_verbose`.

PR #224, opened 2023-02-08
[mini.ai]: add 'verbose' option to suppress messages

mawkler:
Adds the option `verbose` which (when set to `false`) disables messages like this:

```
(mini.ai) No textobject "ib" found covering region within 50 lines and `search_method = 'next'`.
```

Neovim doesn't print any message if you try to access built-in text-objects like `i"`, `a)`, that don't exist. Therefore I think it's reasonable to have an option to disable mini.ai's messages.

### DIFF
--- a/lua/mini/ai.lua
+++ b/lua/mini/ai.lua
@@ -501,6 +501,9 @@ MiniAi.config = {
   -- neighborhood). One of 'cover', 'cover_or_next', 'cover_or_prev',
   -- 'cover_or_nearest', 'next', 'prev', 'nearest'.
   search_method = 'cover_or_next',
+
+  -- Print message if a textobject can't be found
+  verbose = true,
 }
 --minidoc_afterlines_end
 
@@ -538,7 +541,7 @@ MiniAi.find_textobject = function(ai_type, id, opts)
   -- Find region
   local res = H.find_textobject_region(tobj_spec, ai_type, opts)
 
-  if res == nil then
+  if H.get_config().verbose and res == nil then
     local msg = string.format(
       [[No textobject %s found covering region%s within %d line%s and `search_method = '%s'`.]],
       vim.inspect(ai_type .. id),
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for PR! 

> Neovim doesn't print any message if you try to access built-in text-objects like `i"`, `a)`, that don't exist. Therefore I think it's reasonable to have an option to disable mini.ai's messages.

The reason is explained in [this comment](https://github.com/echasnovski/mini.nvim/issues/135#issuecomment-1263305433) of the issue for this specific functionality.

------

Adding `verbose` option is tracked in #146. I plan to use it a little bit more broader than suppressing these "not found messages".

It will be a part of work scope for after the release.

mawkler:
@echasnovski  Ok, I see!

Btw, I noticed that you use `echo` to print the messages. Is there any reason why you're not using `vim.notify` which allow users to use plugins like [nvim-notify](https://github.com/rcarriga/nvim-notify) to display the messages in a floating popup window instead of at the bottom? That way you're not forced to "press any button" when the message is multiple lines long.

echasnovski:
> Btw, I noticed that you use `echo` to print the messages. Is there any reason why you're not using `vim.notify` which allow users to use plugins like [nvim-notify](https://github.com/rcarriga/nvim-notify) to display the messages in a floating popup window instead of at the bottom?

Well, that is exactly the reason. There are many messages that I find helpful (especially for new users), but which are issued quite frequently. Like "Enter character" reminder after one second of not pressing anything, etc. Initially it did use `vim.notify`, but stopped in 1764bcfb.

> That way you're not forced to "press any button" when the message is multiple lines long.

You shouldn't have the need to "press any button" when this type of message is long. It was handled in b8e9fccd. If you experience any hit-enter-prompt during any type of "information messages" (i.e. not errors or warning-like messages dedicated to grab user's attention).

mawkler:
> Initially it did use vim.notify, but stopped in 1764bcf

Ok that makes sense.

> You shouldn't have the need to "press any button"

I realize now that I'm seeing the message because I'm using `vim.o.cmdheight = 0`. I'll have to figure out some way to display messages in my statusline instead.

echasnovski:
I've decided to go the route of global/buffer-local variables for this one, as it more cleanly indicates that this approach is universal for all modules.

In latest `main` branch you should be able to set `vim.g.miniai_silence = true` in your Lua config file and no messages should be shown (neither helper message after one idle second, nor "No textobject ... found"). Sorry, not exactly what you were suggesting here, but this is the way now.

Thanks for the final push to make me implement this.

PR #264, opened 2023-03-11
vimdoc format patches

osamuaoki:
- [YES] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [ YES] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

This pull request fixes 2 trivial vimdoc format issues as below:

 (doc) Fix format of vimdoc for LOCAL ADDITIONS use
    
    In order for nvim to list entries under "LOCAL ADDITIONS:", vimdoc
    should start with particular format.
     * The leading line should contain module name.
     * There should be no leading space in it.
    
    I understand these documentation links under "LOCAL ADDITION:" for many
    other packages are also missing under the modern delayed loading
    situation under lazy.nvim. This makes it easy for many users to overlook
    this kind of bug. Such issue of missing documentation links can be
    addressed by, e.g.,  creating symlinks from ~/.config/nvim/doc/
    directory. But this vimdoc format issuees of mini* were preventing to
    list links for these documents while many other packages from other
    author didn't suffer such issue since they followed this initial line
    format convention.
    
    I (or nvim auto-formatter) also fixed missing newline at the
    end-of-file.  This is generally good practice for text files.

 (doc) fix format for mini.txt for conceal
    
    In order for the conceal feature of nvim to work nicely, there should be
    a space before leading "|" (bar) character. (I suppose this is dictated
    by some treesitter logic used for this feature.)
    
    I also fixed one line with 2 leading spaces to match other lines.


osamuaoki:
As for the first patch, the first line may be:
```
*filename.txt*  explanation
```

echasnovski:
Thank you for the PR!

I would be better if you opened issue to discuss this. The problem is that all documentation is auto-generated using 'mini.doc' module. I am currently looking for the best way to manage that first line for local additions.

Modeline (last line of help files) seems to be formatted in the same way as Neovim's built-in help files are. So I don't thing it should be changed.

Adding space is a good catch, thanks! I personally don't yet use tree-sitter for help files, so didn't see that.

echasnovski:
Thank you for taking your time and making this PR! I've learned about `:h local-additions` thanks to it :)

I've decided to make a long overdue rework of headers for help files and made sure to have new files appear in `local-additiions`.

I also update table of contents in main help file (with you as commit co-author), so it should now be displayed correctly with tree-sitter for help files enabled.

osamuaoki:
ThanksYah, I should have made this PR as proof-of-concept/wip…Sent from iPhone 2023/03/12 23:06、Evgeni Chasnovski ***@***.***>のメール:﻿
Thank you for the PR!
I would be better if you opened issue to discuss this. The problem is that all documentation is auto-generated using 'mini.doc' module. I am currently looking for the best way to manage that first line for local additions.
Modeline (last line of help files) seems to be formatted in the same way as Neovim's built-in help files are. So I don't thing it should be changed.
Adding space is a good catch, thanks! I personally don't yet use tree-sitter for help files, so didn't see that.

PR #277, opened 2023-03-22
(mini.basics) Keep default nohl mapping in Neovim

joelazar:

Neovim's default nohl mapping is `C-l` instead of `C-L`.

### DIFF
--- a/lua/mini/basics.lua
+++ b/lua/mini/basics.lua
@@ -671,7 +671,7 @@ H.keymap_set = function(modes, lhs, rhs, opts)
     local map_info = H.get_map_info(mode, lhs)
     local is_default = map_info == nil
       -- Some mappings are set by default in Neovim
-      or (mode == 'n' and lhs == '<C-L>' and map_info.rhs:find('nohl') ~= nil)
+      or (mode == 'n' and lhs == '<C-l>' and map_info.rhs:find('nohl') ~= nil)
       or (mode == 'x' and lhs == '*' and map_info.rhs == [[y/\V<C-R>"<CR>]])
       or (mode == 'x' and lhs == '#' and map_info.rhs == [[y?\V<C-R>"<CR>]])
     if not is_default then return end


echasnovski:
Does this PR fixes something for you which doesn't work right now?

I think it is correct as is, mainly for two reasons:
- Case of second key in ctrl mappings doesn't matter for creating keymap. Like `<C-l>` and `<C-L>` both represent same combination of keys.
- Usage of `<C-L>` instead of `<C-l>` is prompted by the fact that this is how `vim.api.nvim_get_keymap()` returns this key combo.

joelazar:
> Does this PR fixes something for you which doesn't work right now?
> 
> I think it is correct as is, mainly for two reasons:
> 
>     * Case of second key in ctrl mappings doesn't matter for creating keymap. Like `<C-l>` and `<C-L>` both represent same combination of keys.
> 
>     * Usage of `<C-L>` instead of `<C-l>` is prompted by the fact that this is how `vim.api.nvim_get_keymap()` returns this key combo.

Mhm, for me without this change the `C-l` mapping didn't work when I wanted to hide the search highlights after a search. However, I see your point as well and it might be a side-effect of some weird combinations from my local setup now (kitty, neovim 0.8.3 or some other plugins) I'll try to come up with a minimal reproducible example later today.

echasnovski:
What does `:nmap <C-l>` show without this PR's change?

If you enabled `mappings.windows` then `<C-l>` is overwritten as `<C-w>l` for more coherent window navigation. So If you want to use window navigation mappings, you can not enable them in the `setup()` and instead put modified [version of the source code](https://github.com/echasnovski/mini.nvim/blob/main/lua/mini/basics.lua#L627) in your config.

joelazar:
yes, sorry, that was it. I've enabled `mapping.windows` in my config and forgot about it. Closing this one now. Thank you for your help :pray:

Issue #86, opened 2022-05-28
Add Module: Better Escape


Sduby22:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

https://github.com/max397574/better-escape.nvim

### Description

a better escaping module for those who prefer 'jk' or 'jj' to normal esc or Ctrl-C

This module provides better experience than simply `imap jk <esc>`, such as:

- Customizable delay which is independent from `timeoutlen`
- `j` will be immediately echoed back on screen when pressed. (if set `imap jk <esc>`, `j` wont be echoed until `timeoutlen` expired or you pressed something other than `k`. )

See: [Better Escape](https://github.com/max397574/better-escape.nvim)

echasnovski:
That is an interesting suggestion. Thanks!

I don't think just mimicking this (already Lua based) plugin is the way to go, though. This functionality might be a part of some more general module that unobtrusively tracks keys and does something when certain key sequence is seen. I'll think about it.

matu3ba:
Better alternatives are 1. https://github.com/anuvyklack/hydra.nvim, 2. https://github.com/Iron-E/nvim-libmodal and both dont rely on hacky things like `timeoutlen`.

echasnovski:
> Better alternatives are 1. https://github.com/anuvyklack/hydra.nvim, 2. https://github.com/Iron-E/nvim-libmodal and both dont rely on hacky things like `timeoutlen`.

Yeah, I'll look more closely at them when attempting ['mini.clue'](https://github.com/echasnovski/mini.nvim#planned-modules).

echasnovski:
This is now possible with [`MiniKeymap.map_combo()`](https://github.com/echasnovski/mini.nvim/blob/ee23e1abc206efc6d6cce19ec8c0a3da7a897035/doc/mini-keymap.txt#L370). Here are [examples of how to achieve "better escape" like behavior](https://github.com/echasnovski/mini.nvim/blob/ee23e1abc206efc6d6cce19ec8c0a3da7a897035/doc/mini-keymap.txt#L136-L151).

For any questions/recommendations use #1780.

PR #279, opened 2023-03-23
(mini.basics) Map 'jk' as <ESC>

BreakingLead:
- [ ] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)

There is a line of document says:"    -- Basic mappings (better 'jk', save with Ctrl+S, ...)"
But I didn't find the mapping of "better jk" so I added it.
I'm not a native English speaker so I'm sorry if my words are confusing.

### DIFF
--- a/lua/mini/basics.lua
+++ b/lua/mini/basics.lua
@@ -564,8 +564,15 @@ H.apply_mappings = function(config)
     -- Copy/paste with system clipboard
     map({ 'n', 'x' }, 'gy', '"+y', { desc = 'Copy to system clipboard' })
     map(  'n',        'gp', '"+p', { desc = 'Paste from system clipboard' })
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

By "better 'jk'" it meant that `j` and `k` [move by visible lines when no `[count]` is set](https://github.com/echasnovski/mini.nvim/blob/9f8b92998a9964e03535c2c7e0708958e23e261a/doc/mini-basics.txt#L233-L234).

This type of Insert mode exit introduces delay when typing `j` and/or `k` in Insert mode. Which is not good. A ["better-escape"](https://github.com/max397574/better-escape.nvim) approach is planned for this (see #86), just not as separate module but as a function in 'mini.misc'. So, sorry, I'll close this.

PR #280, opened 2023-03-23
(mini.misc) Use neovim API for MiniMisc.restore_cursor

cryptomilk:
There is also an API for setting marks but not for opening folds. So it doesn't make sense to use `vim.api.nvim_win_set_cursor()` yet.


cryptomilk:
The failing test looks unrelated to this change.

echasnovski:
These changes are not that necessary, as they introduce neither much speed up nor concise code. But I think I'll merge it with slight tweaks for a little bit more consistency with other code base. Thanks!

PR #295, opened 2023-04-11
Add instructions for NVIM_APPNAME installation

craigmac:


echasnovski:
Thanks for suggestion!

I don't think `NVIM_APPNAME` needs a separate mention in installation instructions. Besides, suggesting to try config with **only** 'mini.nvim' doesn't seem like a good idea to me.

PR #315, opened 2023-05-01
(mini.misc) Don't run filetype autocmd if the buffer is not a normal buffer

cryptomilk:
We can already leave earlier ...


echasnovski:
Let's just leave this as is. I am not sure there is a guarantee that `buftype` can't change between `BufReadPre` and `FileType` events. Besides, it doesn't gain much performance too.

PR #362, opened 2023-06-09
(mini.indentscope) Enable only on normal buffers

tssm:

Hey, Indentscope is enabled automatically unless `miniindentscope_disable` is explicitly set, which makes it a bit annoying on buffers you will never edit. I could put this under a flag if you think that's better, however IMO the feature only makes sense on normal buffers. What do you think?

### DIFF
--- a/lua/mini/indentscope.lua
+++ b/lua/mini/indentscope.lua
@@ -678,7 +678,7 @@ end
 
 -- Autocommands ---------------------------------------------------------------
 H.auto_draw = function(opts)
-  if H.is_disabled() then
+  if vim.bo.buftype ~= '' or H.is_disabled() then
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

I don't think this is a good idea for at least these reasons:
- There are non-normal buffers which are still modifiable, like scratch buffers.
- It still can be useful to see indent scope as reference even if users can't modify current buffer. Having this rule inside 'mini.indentscope' makes it harder to configure.

If users want to only show indent scope inside normal buffers, adding this autocommand is a suggested solution:
`au BufEnter * lua vim.b.miniindentscope_disable = vim.bo.buftype ~= ''`

PR #390, opened 2023-06-27
(mini.files) FEATURE: Add callbacks.fs_delete

mrshmllow:
This option allows users to define the way they would like files to be deleted. An example would be:

```lua
opts = {
	callbacks = {
		fs_delete = function(path)
			local obj = vim.system({
				"gio",
				"trash",
				path,
			}):wait()

			return obj.code == 0
		end,
	}
}
```

Feedback on documentation and implementation details would be appreciated as always. I just wanna see this useful feature happen.


echasnovski:
Thanks for taking your time writing this.
At the moment I don't think this extra option is a good fit for "usefulness vs. complexity/maintenance" traid-off. Also having a way to customize deleting but not other actions seems a bit off, which only adds to the complexity/maintenance.

Am I correct understanding that your main issue with current `vim.fn.delete()` approach is that it deletes entirely while you want to move into trash?

I'll think about the best way to handle this, but my hopes are not high for this to end up implemented.

mrshmllow:
I have updated this PR to greatly reduce complexity and scrap callbacks idea

echasnovski:
I am still thinking about whether this is a good fit to 'mini.files', to be honest. Really tempted to say "this is not meant as fully functional file explorer". Not just there yet, though, as I see how this can be an important to some people. May still take a while for me to decide.

mrshmllow:
🎉

echasnovski:
@mrshmllow, I've opted to adding module-specific "trash" functionality behind a new `options.permanent_delete` option. If set to `false`, it will move deleted entries to 'mini.files/trash' directory of Neovim's standard `data` path ('~/.local/share/nvim' on Unix). Here is [full info](https://github.com/echasnovski/mini.nvim/blob/f3648ab6ce4a5a44f632f98d5b6181a0463a5efc/doc/mini-files.txt#L587-L591).

My main reason for not going with `delete_cmd` is mostly because it is virtually Unix-only; Windows users probably won't install any CLI apps. 

Although I did not accept these changes, thanks for your enthusiasm to get this type of feature added.

PR #403, opened 2023-07-05
refactor!: remove nvim-ts dependency

amaanq:

The new nvim-ts main branch removes a lot of the functionality and brings many performance improvements, but it is not the default branch for now, and won't be for a while. Though, after 0.10 master is pretty much deprecated, and plugins should not depend on nvim-ts utilities besides locals...this will also allow users to use the main branch without these plugins breaking. Thanks!

### DIFF
--- a/lua/mini/ai.lua
+++ b/lua/mini/ai.lua
@@ -1151,8 +1151,9 @@ end
 
 H.is_disabled = function() return vim.g.miniai_disable == true or vim.b.miniai_disable == true end
 
-H.get_config =
-  function(config) return vim.tbl_deep_extend('force', MiniAi.config, vim.b.miniai_config or {}, config or {}) end
+H.get_config = function(config)
+  return vim.tbl_deep_extend('force', MiniAi.config, vim.b.miniai_config or {}, config or {})
+end
 
 H.is_search_method = function(x, x_name)
   x = x or H.get_config().search_method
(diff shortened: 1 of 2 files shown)


clason:
Note: `locals.lua` is also not guaranteed to be kept in nvim-treesitter 1.0; only the queries will (most likely) be maintained.

If there are (widely) useful utility functions missing, they should be considered for inclusion in the core API.

echasnovski:
> The new nvim-ts main branch removes a lot of the functionality and brings many performance improvements, but it is not the default branch for now, and won't be for a while. Though, after 0.10 master is pretty much deprecated, and plugins should not depend on nvim-ts utilities besides locals...this will also allow users to use the main branch without these plugins breaking. Thanks!

Thanks the PR!

To be honest, I am not keen on making braking changes for the not stable work-in-progress branch which is not the default yet. Besides, all proposed changes address tree-sitter based functionality which can choose what approach to use through `opts.use_nvim_treesitter`. If users want/need to use `main` 'nvim-treesitter' branch, I'd recommend using `opts.use_nvim_treesitter = false` for now.

@clason, the one utility function used in 'mini.ai' and 'mini.surround' which I can not find in Nightly is `require('nvim-treesitter.query').get_capture_matches_recursively()`. Is this considered to be a (widely) used utility function?

clason:
Is this a breaking change though? The user-facing behavior should not change, except that users can (but don't have to) drop nvim-treesitter if the "bring their own parser".

> require('nvim-treesitter.query').get_capture_matches_recursively(). Is this considered to be a (widely) used utility function?

Not sure; if someone were to make a PR, a quick GH search would tell them. It would also depend on whether this can be easily replicated using already provided functions.

echasnovski:
> Is this a breaking change though? The user-facing behavior should not change, except that users can (but don't have to) drop nvim-treesitter if they "bring their own parser".

I would assume that such change would require new tree-sitter utilities ported from 'nvim-treesitter' into Neovim core, which makes it a hard dependence on nightly (hence breaking for others Neovim versions). The code in PR seems like a direct copy-paste from 'nvim-treesitter' source code, so can't judge beyond that.

amaanq:
It is just copy-pasted, so it wouldn't be breaking for now

clason:
That being said, there's no particular urgency; the deprecation of the legacy nvim-treesitter won't be happening until after the Nvim 0.10 release, probably sometime (late) this summer.

clason:
> It is just copy-pasted (slightly adjusted), so it wouldn't be breaking for now

Yeah, that's not good enough; you need an actual refactor of the calling code based on current Neovim API. I'd recommend closing this PR and trying again later.

amaanq:
the adjusted bit is adjusted for the current Neovim API

clason:
You did not adjust the actual code at all (or sufficiently, it looks like); just a copy-paste leaving references to internal nvim-treesitter code...

amaanq:
oh yeah...I relied on lsp too much, will close for now, my bad

clason:
@amaanq If you permit me to be blunt: this PR is a bit of an ill-considered hip shot. The first step should be to look at our own `nvim-treesitter-textobjects` and do a proper refactor of this (in a separate `main` branch). After this is done (possibly including upstreaming utility functions), we can take the lessons learned to the broader ecosystem.

amaanq:
yeah that's a good idea, that'd also give me a better sense of how to refactor for nvim-ts main going forward

PR #417, opened 2023-07-14
(mini.basics) Use a more specific 'TermOpen' autocomand pattern

LkeMitchll:

***

The pattern that the “Start builtin terminal in Insert mode” autocommand is using is not specific enough, which seems to cause [Neogit](https://github.com/NeogitOrg/neogit) to open in insert mode which shouldn’t be possible, causing an error that stops Neogit being used until the `esc` key is hit:

`E21: Cannot make changes, ‘modifiable’ is off`

Minimal setup example (using lazy as a plugin manager):

```lua
require("lazy").setup({
  {
    "echasnovski/mini.nvim",
    init = function()
      require("mini.basics").setup({})
    end
  },
  {
    "NeogitOrg/neogit",
    dependencies = {
      "nvim-lua/plenary.nvim"
    },
    keys = { { "<C-G>", ":Neogit" } }
  }
})
```

Disabling the basic autocommands solves the issue, but I’d like to still use them.

[Others have found](https://github.com/NeogitOrg/neogit/issues/426#issuecomment-1374833515) that changing the pattern to `term://*` stops the Neogit problem, and keeps the autocommand working, which is the method that this PR uses.

### DIFF
--- a/lua/mini/basics.lua
+++ b/lua/mini/basics.lua
@@ -721,7 +721,7 @@ H.apply_autocommands = function(config)
 
   if config.autocommands.basic then
     au('TextYankPost', '*', function() vim.highlight.on_yank() end, 'Highlight yanked text')
-    au('TermOpen', '*', function() vim.cmd('startinsert') end, 'Start builtin terminal in Insert mode')
+    au('TermOpen', 'term://*', function() vim.cmd('startinsert') end, 'Start builtin terminal in Insert mode')
   end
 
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

This looks reasonable. I'll investigate a bit to see if there are some other unwanted consequences.

echasnovski:
Thanks again for the PR!

I'll added your patch along with some tests and fix for other issue. Should work on current `main` branch.

PR #426, opened 2023-07-23
(mini.files) FEATURE: Add MiniFiles.toggle()

David-Kunz:

Hi @echasnovski ,

Thanks for this genius plugin! Do you think it makes sense to include a `toggle` functionality? This way, we'd only need one binding to open/close the explorer (and avoid the risk of recording a macro when pressing `q` too much).

Please let me know what you think, feel free to close this PR if not needed.

Thanks and best regards,
David

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -683,16 +683,7 @@ Return~
                 `MiniFiles.toggle`({path}, {use_latest}, {opts})
 Toggle file explorer
 
-Parameters~
-{path} `(string|nil)` A valid file system path used as anchor.
-  If it is a path to directory, used directly.
-  If it is a path to file, its parent directory is used as anchor while
-  explorer will focus on the supplied file.
-  Default: path of |current-directory|.
-{use_latest} `(boolean|nil)` Whether to load explorer state from history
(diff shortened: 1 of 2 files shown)


echasnovski:
Hi, David! Thanks for the PR!

I've been thinking about adding `toggle()` for a while now. The main reason stopping me from doing so is that for users to write their own function for that is basically a one-liner:

```lua
require('mini.files').setup()
local minifiles_toggle = function(...)
  if not MiniFiles.close() then MiniFiles.open(...) end
end
-- Use different left hand side for your liking
vim.keymap.set('n', '<Leader>et', minifiles_toggle)
```

I'll think about this a bit more. Maybe adding this code to examples is a better solution.

David-Kunz:
Hi @echasnovski ,

Thank you for your comment and thoughts!

The approach with `if not MiniFiles.close() then MiniFiles.open() end` is neat, I'll use that in my config, beautiful API!

echasnovski:
After thinking about this for a bit, I am still not convinced that adding separate ``toggle()` is worth it over having a documented one-liner. So I added it to examples.

Anyway, thanks for the contribution, @David-Kunz.

PR #434, opened 2023-08-07
(mini.pair) fix: prevent `'` from inserting after `&` character

MinaSaad47:


MinaSaad47:
(mini.pair) prevent ' from inserting after & character (for rust life times case)

echasnovski:
I don't think this is a general enough case to be included by default. Users can customize this themselves if they want to.

PR #435, opened 2023-08-09
(mini.starter) Feature: exposed path string configuration via path_modifier parameter

benjaminchristie:
Feature:

This patch adds a `path_modifier` parameter to MiniStarter.sections.recent_files which changes the way `path` is displayed. This defaults to `nil`, which is redefined as `function(x) return x end`.

By exposing the `path` variable via the `path_modifier` parameter, one can change how the `path` for each file is displayed, such as shortening the path. See below for a potential use case:

```
local function path_modifier(str, max_count)
    -- can use vim.regex() for this...
    local tail = vim.fn.fnamemodify(str, ":t")
    local head = vim.fn.fnamemodify(str, ":h")
    local new_str = ""
    local count = 0
    if head == nil or #head <= max_count then
        return str
    else
        for c in string.gmatch(head, '.') do
            if c == "/" then
                count = 0
            end
            if count < max_count then
                new_str = new_str .. c
                count = count + 1
            end
        end
    end
    new_str = new_str .. "/"
    return new_str .. tail
end
require("mini.starter").setup({
    items = {
        starter.sections.recent_files(.., .., .., function(x) return path_modifier(x, 4) end),
    },
})

```

### DIFF
--- a/lua/mini/starter.lua
+++ b/lua/mini/starter.lua
@@ -490,12 +490,16 @@ end
 ---   directory and its subdirectories. Default: `false`.
 ---@param show_path boolean|nil Whether to append file name with its full path.
 ---   Default: `true`.
+---@param path_modifier function|nil Modifier to apply to full path of file name.
+---   Expected to be function that accepts a string and returns a string
+---   Default: 'nil'.
 ---
 ---@return __starter_section_fun
-MiniStarter.sections.recent_files = function(n, current_dir, show_path)
+MiniStarter.sections.recent_files = function(n, current_dir, show_path, path_modifier)
   n = n or 5
   if current_dir == nil then current_dir = false end
   if show_path == nil then show_path = true end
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

I am not sure if this fits nicely with current design. I'll take some time to think about it and get back to you.

benjaminchristie:
Understood. A more friendly API may be to use a gen_hook. If this is desired, I can modify this PR / submit a new one.

echasnovski:
Hi again after a long time! I've finally made an effort to revisit the backlog.

What this PR achieves with new `path_modifier` is now possible on `main` branch with callable `show_path`. See [this part of help](https://github.com/echasnovski/mini.nvim/blob/b94d010c61ba46474af78536873db89492b251a4/doc/mini-starter.txt#L370-L371).

Nevertheless, thanks for your effort doing this PR!

PR #465, opened 2023-09-05
(mini.bufremove) Feature: Configurable fallback on delete last buffer

MikaelElkiaer:

I took a default and made it configurable.

I utilize this in my own config of mini.bufremove:

```lua
{
  "echasnovski/mini.bufremove",
  dev = true, -- Remove once PR is accepted
  opts = {
    on_delete_last_fallback = function(_)
      local orig_cwd = os.getenv("PWD")
      vim.fn.chdir(orig_cwd)
      vim.cmd("SessionDelete")
      vim.cmd("Alpha")
    end,
  },
}
```

### DIFF
--- a/doc/mini-bufremove.txt
+++ b/doc/mini-bufremove.txt
@@ -29,7 +29,8 @@ To stop module from showing non-error feedback, set `config.silent = true`.
    decided by the algorithm:
    - If alternate buffer (see |CTRL-^|) is listed (see |buflisted()|), use it.
    - If previous listed buffer (see |bprevious|) is different, use it.
-   - Otherwise create a new one with `nvim_create_buf(true, false)` and use it.
+   - Otherwise use fallback function `config.on_delete_last_fallback(win_id)`.
+     (Default is to create a scratch one with `nvim_create_buf(true, false)`)
 
 # Disabling~
 
@@ -58,6 +59,13 @@ Module config
 Default values:
 >
   MiniBufremove.config = {
+    -- Fallback function when deleting last buffer
+    on_delete_last_fallback = function(win_id)
+      -- Create new listed scratch buffer
+      local new_buf = vim.api.nvim_create_buf(true, false)
+      vim.api.nvim_win_set_buf(win_id, new_buf)
+    end,
+
     -- Whether to set Vim's settings for buffers (allow hidden buffers)
     set_vim_settings = true,
 
--- a/lua/mini/bufremove.lua
+++ b/lua/mini/bufremove.lua
@@ -29,7 +29,8 @@
 ---    decided by the algorithm:
 ---    - If alternate buffer (see |CTRL-^|) is listed (see |buflisted()|), use it.
 ---    - If previous listed buffer (see |bprevious|) is different, use it.
----    - Otherwise create a new one with `nvim_create_buf(true, false)` and use it.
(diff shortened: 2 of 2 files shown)


echasnovski:
Thanks for the PR!

I'd rather not introduce this functionality instead suggesting to write own wrappers using exported functions. Something [like this](https://github.com/echasnovski/mini.nvim/issues/249#issuecomment-1452431730).

In your case it can be something like this:

```lua
local open_starter_if_empty_buffer = function()
  local is_empty = vim.api.nvim_buf_get_name(0) == "" and vim.bo.filetype == ''
  if not is_empty then return end

  local orig_cwd = os.getenv("PWD")
  vim.fn.chdir(orig_cwd)
  vim.cmd("SessionDelete")
  require('mini.starter').open()
end

_G.my_bufdelete = function(...)
  require('mini.bufremove').delete(...)
  open_starter_if_empty_buffer()
end

_G.my_bufwipeout = function(...)
  require('mini.bufremove').wipeout(...)
  open_starter_if_empty_buffer()
end
```

MikaelElkiaer:
Oh damn, was not thorough enough in my search.

Thanks for the suggestion. :)

echasnovski:
> Oh damn, was not thorough enough in my search.
> 
> Thanks for the suggestion. :)

No problem. It *was* hidden somewhat deep.

MikaelElkiaer:
I still have a problem with a dangling '[No Name]' buffer, created by `require("mini.bufremove").delete(...)`.

I tried out the following:

```lua
{
  "echasnovski/mini.bufremove",
  keys = {
    {
      "<leader>bd",
      function(...)
        require("mini.bufremove").delete(...)
        local is_empty = vim.api.nvim_buf_get_name(0) == "" and vim.bo.filetype == ""
        if not is_empty then
          return
        end

        -- vim.api.nvim_buf_delete(0, {}) -- desperate times...
        local orig_cwd = os.getenv("PWD")
        vim.fn.chdir(orig_cwd)
        vim.cmd("SessionDelete")
        vim.cmd("Alpha")
      end,
      desc = "Delete Buffer",
    },
  },
}
```

echasnovski:
> I still have a problem with a dangling '[No Name]' buffer, created by `require("mini.bufremove").delete(...)`.
> 
> I tried out the following:

I'd say you started on a good route here. Try something like this:

```lua
local open_starter_if_empty_buffer = function()
  local buf_id = vim.api.nvim_get_current_buf()
  local is_empty = vim.api.nvim_buf_get_name(buf_id) == '' and vim.bo[buf_id].filetype == ''
  if not is_empty then return end

  local orig_cwd = os.getenv('PWD')
  vim.fn.chdir(orig_cwd)
  vim.cmd('SessionDelete')
  require('mini.starter').open()
  vim.api.nvim_buf_delete(buf_id, { force = true })
end
```

MikaelElkiaer:
@echasnovski TYVM!

Not 100 % sure why my own example didn't work.
Perhaps I was not allowed to delete the empty buffer since it was the last one?

echasnovski:
> Not 100 % sure why my own example didn't work.
> Perhaps I was not allowed to delete the empty buffer since it was the last one?

it essentially didn't do anything (even if uncommented). Mostly because Neovim *needs* at least some buffer to show. When you execute other commands and give Neovim something else to show, it can now delete that buffer.

PR #492, opened 2023-09-26
(mini.files) FEATURE: do not start syntax highlighting if preview buffer size exceeds limit.

### LINKED ISSUE #489: (mini.files) preview is unusable with large minified files
### Contributing guidelines

- [X] I have updated 'mini.nvim' to latest version

### Module(s)

mini.files

### Description

Plugin explicitly starts treesitter causing neovim to freeze

### Neovim version

0.9.1

### Steps to reproduce

1. install treesitter syntax for js (doesn't matter if it's js/css or any other language)
2. enable mini.files preview
3. create following structure (use at least ~500kB minified js file) i.e.:
```bash
mkdir some-dir
touch some-dir/a.js some-dir/b.js some-dir/d.js
curl https://cdn.jsdelivr.net/npm/angular2@2.0.0-beta.21/bundles/angular2-all.umd.min.js > some-dir/c-slow.js
```
4. browse this structure from `some-dir/a.js` to `some-dir/d.js` with `j`

### Expected behavior

It should be possible to preview a file without starting treesitter / enabling syntax highlighting or decide whenever to show preview for specific file. 

There are few options for current implementation

1. `boolean` option to disable syntax highlighting globally for mini.files
2. `function` option to disable preview completely per file (similarly how non text files work)
3. `function` option to disable syntax highlighting per file

### Actual behavior

Neovim freezes when large minified file is previewed.

It is possible to use `MiniFilesBufferUpdate` to disable syntax and stop treesitter, but you will still experience little freeze first time you preview file (caused by already started treesitter).

```lua
-- current workaround
    group = vim.api.nvim_create_autocmd("mygroup", { clear = true }),
    pattern = "MiniFilesBufferUpdate",
    callback = function(args)
        local buffer = args.data.buf_id

        vim.bo[buffer].syntax = nil
        vim.treesitter.stop(buffer)
    end,
})
```

fbuchlak:
Details
- Resolves #489

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -531,6 +531,8 @@ Default values:
     windows = {
       -- Maximum number of windows to show side by side
       max_number = math.huge,
+      -- Maximum size of preview buffer in bytes to start syntax highlighting
+      max_preview_bytes = 50 * 1024,
       -- Whether to show preview of file/directory under cursor
       preview = false,
       -- Width of focused window
@@ -609,6 +611,10 @@ directory (execute `:echo stdpath('data')` to see its path in your case).
 `windows.max_number` is a maximum number of windows allowed to be open
 simultaneously. For example, use value 1 to always show single window.
 There is no constraint by default.
+
+`windows.max_preview_bytes ` is a maximum size in bytes that is preview buffer
+allowed to have in order to start syntax highlighting. For example, use value
+50 * 1024 start syntax highlighting only in buffers that have less than 50 KiB.
 
 `windows.preview` is a boolean indicating whether to show preview of
 file/directory under cursor.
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for making this!

I am still not sure if this should be a separate configurable option or something that is computed based on the Neovim state. Like `1000 * vim.o.lines` or something else.

fbuchlak:
> Thanks for making this!
> 
> I am still not sure if this should be a separate configurable option or something that is computed based on the Neovim state. Like `1000 * vim.o.lines` or something else.

That might be enough if you don't want to introduce it as option. If someone wants to start syntax highlighting, `MiniFilesBufferUpdate` can be used.

echasnovski:
Thanks again for making this PR!

I've decided to go slightly different approach and not highlight if either average line length is more than 1K bytes or total size is more than 1M bytes. Thanks for bringing this up!

PR #520, opened 2023-10-16
(mini.map) FEATURE: Adds zindex as window option

nolantait:

Needed this to avoid orchestrating my other floats around the default 10 z-index. [nvim-treesitter-context](https://github.com/nvim-treesitter/nvim-treesitter-context) has a default of 20 so anyone installing it would run into this issue.

I'm currently using a [fork](https://github.com/nolantait/mini.map) of this on my [configs](https://github.com/nolantait/dotfiles/blob/a0c754859046d513ea6c0ce7ee89b74e4b839f8c/config/nvim/lua/plugins/configs/mini-map.lua#L43) and its working great.

### DIFF
--- a/doc/mini-map.txt
+++ b/doc/mini-map.txt
@@ -218,6 +218,9 @@ Default values:
 
       -- Value of 'winblend' option
       winblend = 25,
+
+      -- Z-index of map window
+      zindex = 10,
     },
   }
 <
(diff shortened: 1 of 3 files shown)


echasnovski:
OK, I think that adding `zindex` is indeed the most reasonable solution here. Allowing full control over window config might be harmless as allowing full control over window config will also need several adjustments in other places (that also need tests).

I've merged it in separate branch for to add a few touches. Will merge in `main` soon.

echasnovski:
This now should be in `main` branch. Thanks again for the PR!

PR #539, opened 2023-10-24
Basics: enhance window resize logic

wroyca:
Window resizing was a bit confusing, with left sometimes behaving like right and right sometimes behaving like left. This was not in line with user expectations. This PR enhance the logic behind, ensuring that left remains left and right remains right, similar to how it works for "nN" and "Nn" and/or "gj" and "gk".

[Screencast from 2023-10-24 09-02-29.webm](https://github.com/echasnovski/mini.nvim/assets/91024200/a029c13a-e495-4ba6-8ea6-1d30f952446e)

>  The best way to do it is to receive a positive feedback from maintainer on your initiative in one of the GitHub issues (existing one or created by you otherwise). 

Sorry about that, I read this after the fact ^

### DIFF
--- a/lua/mini/basics.lua
+++ b/lua/mini/basics.lua
@@ -646,10 +646,10 @@ H.apply_mappings = function(config)
     map('n', '<C-L>', '<C-w>l', { desc = 'Focus on right window' })
 
     -- Window resize (respecting `v:count`)
-    map('n', '<C-Left>',  '"<Cmd>vertical resize -" . v:count1 . "<CR>"', { expr = true, replace_keycodes = false, desc = 'Decrease window width' })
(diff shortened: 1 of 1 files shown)


wroyca:
I will need some free time available to figure out the whole testing machinery and update it accordingly. If anyone wants to beat me to it, feel free to go ahead!

echasnovski:
Thanks for taking time to make this!

> > The best way to do it is to receive a positive feedback from maintainer on your initiative in one of the GitHub issues (existing one or created by you otherwise).
> 
> Sorry about that, I read this after the fact ^

Yeah, that is somewhat of a crucial part here. I did notice this behavior when adding these keymaps and intentionally left them as is. The point of 'mini.basics' mappings is to be easy to understand and copy-pasteable (with some exceptions).

Default commands work as they are described: they increase/decrease width/height. This is usually the same as "move edge right/left/up/down" but not for right/bottom windows (which move left/top edge in opposite direction). This needs some getting used to, but works as intended.

Sorry, but I am closing this one as not planned.

PR #554, opened 2023-11-04
(mini.completion) Update `set_vim_settings`'s `shortmess` to hide completion messages

yamin-shihab:
- [x] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

Right now, mini.completion adds "c" to `shortmess` when `set_vim_settings` is enabled to reduce the amount of command line messages printed when completing. This PR also adds "C" to `shortmess` to further silence unnecessary messages. This has also already been referenced in https://github.com/echasnovski/mini.nvim/issues/5#issuecomment-1278554178 before. I've changed the code so that it detects to see if the user is using Neovim 0.9.0 or newer before adding the flag, and I also edited the appropriate tests.

PR #555, opened 2023-11-06
(mini.misc) FEATURE: implement `setup_mkdir_missing()`.

yamin-shihab:

If this PR were to be merged, mini.misc will have the `setup_mkdir_missing()` function which will automatically create any missing directories needed for saving a file. This is similar to the functionality provided by [tidy.nvim](https://github.com/mcauley-penney/tidy.nvim). I also  made sure it didn't break with netrw URLs, and I wrote it some documentation and tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 
 ## mini.completion
 
-- FEATURE: Start adding `C` flag to `shortmess` option on Neovim>=0.9.
+- FEATURE: Start adding `C` flag to `shortmess` option on Neovim>=0.9. (@yamin-shihab, PR #554)
 
 ## mini.extra
 
@@ -50,6 +50,7 @@ Introduction of a new module.
 
 ## mini.misc
 
+- FEATURE: Add `MiniMisc.setup_mkdir_missing()` for automatically creating missing directories before every save. (@yamin-shihab, PR #555)
 - FEATURE: `setup_auto_root()` and `find_root()` now have `fallback` argument to be applied when no root is found with `vim.fn.find()`.
 
 ## mini.pick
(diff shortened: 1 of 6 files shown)


echasnovski:
Thanks for the PR!

Sorry, but I don't think this is a good functionality to have. Both in terms of what it provides (performing the check on every `BufWrite` is costly and redundant as it can be done several times per minute) and adding it to 'mini.misc' (this is a rather trivial autocommand with almost a one line function). I would suggest instead create a file before starting work in it (for example, with 'mini.files' it is very straightforward).

Please, open an issue to discuss a feature first prior to creating a PR. This way it might lead to less time wasted on your side.

Closing as not planned.

echasnovski:
I will update 'CHANGELOG.md' to proper mention you for #554, though. Sorry about that.

yamin-shihab:
@echasnovski What if I were to get the autocommand to run only once using the `once` `opts` flag for `nvim_create_autocmd` for better performance? And as for it not being suitable for `mini.misc` for how short it is, wouldn't the same apply for `setup_restore_cursor()` because it can mostly just be done with `'"` in normal mode?

echasnovski:
> @echasnovski What if I were to get the autocommand to run only once using the `once` `opts` flag for `nvim_create_autocmd` for better performance?

It won't work with a simple `once` as it should create the autocommand on every `BufCreate` event. Besides, it doesn't address the fact that I don't really want to enable this kind of workflow.

 >  And as for it not being suitable for `mini.misc` for how short it is, wouldn't the same apply for `setup_restore_cursor()` because it can mostly just be done with `'"` in normal mode?

`setup_restore_cursor()` does a bit more than that (checks if cursor was already placed, checks for proper `"` register, opens folds, centers window) and does it smartly.

PR #556, opened 2023-11-06
(mini.completion) Fix reusing non existing buffer for floating windows.

ibutra:

I regularly switch sessions and as part of this wipe all buffers. Because completion doesn't check if the buffer used for floating windows is still present, this gives an error message. This pull request fixes this.


echasnovski:
Thanks for the PR!

I've slightly tweaked the condition and added tests myself. Should now be fixed on the `main` branch.

PR #562, opened 2023-11-08
(mini.extra) Fix git pickers not working when repository path contains spaces

habeebweeb:

Some projects I work on are contained in a directory called 'My Projects'. Since the path is concatenated without  quotes, the `git -C` command fails.

### DIFF
--- a/lua/mini/extra.lua
+++ b/lua/mini/extra.lua
@@ -1588,7 +1588,8 @@ end
 
 H.git_get_repo_dir = function(path, path_type, picker_name)
   local path_dir = path_type == 'directory' and path or vim.fn.fnamemodify(path, ':h')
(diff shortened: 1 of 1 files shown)


echasnovski:
Oh, indeed, thanks for noticing this!

I'll take a look tomorrow for maybe this (respecting paths with spaces) needs a more involvement.

habeebweeb:
I've tried out the other pickers and found that they work fine, including the file explorer. It seems like only the git pickers do not work and they all break at the same point but of course I could be missing something.

echasnovski:
I've also looked through, and this indeed seems like the only place needs to adjusting. This should now be in `main` branch.

Thanks again for the PR!

PR #616, opened 2023-12-15
(mini.files) Expose `go_in_plus()` and `go_out_plus()`

gpanders:
Expose the "plus" variants of `go_in` and `go_out` so that users may
create multiple mappings to these actions.

It is not currently possible to map more than one key to these actions
without implementing their behavior manually, since the only way to map
a key to these actions is through the `mappings` table of the `setup()`
call, which does not support mapping multiple keys to a single action.

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -700,11 +700,21 @@ Depends on entry under cursor:
   Explorer is not closed after that.
 
 ------------------------------------------------------------------------------
+                                                        *MiniFiles.go_in_plus()*
+                            `MiniFiles.go_in_plus`()
+Like |MiniFiles.go_in()|, but closes explorer after opening a file.
+
+------------------------------------------------------------------------------
                                                             *MiniFiles.go_out()*
                               `MiniFiles.go_out`()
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for creating this!

The current suggestion for multiple mappings in `go_in_plus` and `go_out_plus` is to make custom wrappers. Like ones described in #409.

But I think I'll rethink this decision. Either with the approach from this PR or by adding options to `go_in()` and `go_out()`.

I'll think about it.

echasnovski:
After 40086c0a646100c766e8e05cd6e7d75bb1ca37de, `go_in()` now has the `close_on_file` option which can be used in custom mappings. If used `go_in()` directly, it won't support `count`, but I think it is a reasonable compromise.

Also decided to not add any option for `go_out()` to emulate `go_out_plus` as it would quite literally be a call to `trim_right()` inside an `if`. I think it is reasonable to leave this kind of functionality to user wrappers.

All in all, I am going to close this as there are now reasonably good tools to solve the actual problem.

Thanks again for your time and interest!

PR #644, opened 2024-01-06
(mini.bracketed) FEATURE: Add MiniBracketed.tabpage()

augtheo:

This adds support for intuitively navigating through tabpages using the same idea of `[` and `]` mappings as part of mini.bracketed. The suffix for tabpages that I've considered is the `<tab>` key itself.

### DIFF
--- a/doc/mini-bracketed.txt
+++ b/doc/mini-bracketed.txt
@@ -31,38 +31,40 @@ Features:
 - Supported targets (for more information see help for corresponding Lua
   function):
 
-  `Target`                           `Mappings`         `Lua function`
+  `Target`                           `Mappings`                               `Lua function`
 
-  Buffer.......................... `[B` `[b` `]b` `]B` .... |MiniBracketed.buffer()|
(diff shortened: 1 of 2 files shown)


echasnovski:
Thank you for your time doing this!

Unfortunately, this is out of scope because there already are built-in mappings for navigating tabpages: `gt` and `gT`. If user wants to have 'mini.bracketed' style of navigation, there is this set of mappings:

```lua
vim.keymap.set('n', '[T', '<Cmd>tabfirst<CR>')
vim.keymap.set('n', '[t', '<Cmd>tabprevious<CR>')
vim.keymap.set('n', ']t', '<Cmd>tabnext<CR>')
vim.keymap.set('n', ']T', '<Cmd>tablast<CR>')
```

Those don't really support `{count}` but should be good enough alternative.

Besides, `t` suffix is reserved for tree-sitter.

Closing as not planned.

------

Please, open an issue to discuss the change first. This might save you much time.

PR #724, opened 2024-02-29
(mini.basics) Check diagnostic status with vim.diagnostic.is_disabled().

star-szr:
[Details:](url)
- Neovim>=0.9 provides this function, so when possible we can use this rather than having to assume that diagnostics are enabled.

Reference: https://github.com/neovim/neovim/releases/tag/v0.9.0

Just an idea I had, I believe tests will fail but I'm not clear on how to resolve.

A use case I had is wanting to load buffers with diagnostics initially disabled. When doing so, the toggle function doesn't work as expected because it assumes diagnostics will initially be enabled.

Example autocmd:
```lua
vim.api.nvim_create_autocmd("BufReadPost", {
  pattern = "*",
  callback = function()
    vim.diagnostic.disable(0)
  end,
})
```

### DIFF
--- a/lua/mini/basics.lua
+++ b/lua/mini/basics.lua
@@ -361,7 +361,12 @@ MiniBasics.config = {
 ---@return string String indicator for new state. Similar to what |:set| `{option}?` shows.
 MiniBasics.toggle_diagnostic = function()
   local buf_id = vim.api.nvim_get_current_buf()
-  local buf_state = H.buffer_diagnostic_state[buf_id]
+  local buf_state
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Yeah, indeed, it is worth using it. I'll refine it and make it work with tests later (and mention you as co-author).

echasnovski:
Thanks again for bringing my attention to the new `vim.diagnostic.is_disabled()`!
It should now be used on latest `main`.

PR #766, opened 2024-03-23
fix(pick): use `opts.kind` in `ui_select()` as fallback…

AlexKurisu:
… instead of the first thing.
Fixes `vim.ui.select()` prompt in some plugins (such as nvim-tree).

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -1208,7 +1208,7 @@ MiniPick.ui_select = function(items, opts, on_choice)
     end)
   end
 
-  local source = { items = items_ext, name = opts.kind or opts.prompt, preview = preview, choose = choose }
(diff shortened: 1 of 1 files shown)


AlexKurisu:
Was too lazy to edit properly, so used github's editor (this is a one-liner anyway). Feel free to close if i shouldn't have done this.

echasnovski:
Thanks for the PR!

Is there any use case where you think the current source name preference is worse than the one you suggest?

AlexKurisu:
@echasnovski well, for example nvim-tree sets both opts.kind and opts.prompt, so it's prompt looks like this:
image
IMHO, this is not what user expects to see.

AlexKurisu:
With this PR applied it looks like this:
image

echasnovski:
Thanks again for the PR!

Yeah, I agree that preferring `opts.prompt` looks more reasonable here. This should be part of latest `main` now.

PR #780, opened 2024-04-03
feat(files): add support for dynamic buffer width

smilovanovic:
Having just the focused buffer and a preview is my ideal setup but some file names tend to be long and can't be seen with fixed width limitation that currently exists.

This change introduces the ability to have dynamic buffer width which will be determined based on the longest line of that buffer.

With this feature added, my windows config is:
```lua
    windows = {
      -- Maximum number of windows to show side by side
      max_number = 2,
      -- Whether to show preview of file/directory under cursor
      preview = true,
      -- Width of focused window
      width_focus = 0,
      -- Width of non-focused window
      width_nofocus = 0,
      -- Width of preview window
      width_preview = 0,
    }
```

Screenshot 2024-04-04 at 00 14 56

As I'm using Mac, I couldn't add necessary tests

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -535,10 +535,13 @@ Default values:
       -- Whether to show preview of file/directory under cursor
       preview = false,
       -- Width of focused window
+      -- If set to 0 width will be calculated based on the longest line in the buffer
       width_focus = 50,
       -- Width of non-focused window
+      -- If set to 0 width will be calculated based on the longest line in the buffer
       width_nofocus = 15,
       -- Width of preview window
+      -- If set to 0 width will be calculated based on the longest line in the buffer
       width_preview = 25,
     },
   }
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -627,10 +627,13 @@ MiniFiles.config = {
     -- Whether to show preview of file/directory under cursor
     preview = false,
     -- Width of focused window
+    -- If set to 0 width will be calculated based on the longest line in the buffer
     width_focus = 50,
     -- Width of non-focused window
+    -- If set to 0 width will be calculated based on the longest line in the buffer
     width_nofocus = 15,
     -- Width of preview window
+    -- If set to 0 width will be calculated based on the longest line in the buffer
     width_preview = 25,
   },
 }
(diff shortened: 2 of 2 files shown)


echasnovski:
Thanks for the PR!

I generally tried to avoid this type of dynamic window width based on the file content. For at least the following reasons:
- Having as small width values as is reasonably possible is definitely the goal here, as it mean more windows can fit side-by-side on the screen.
- Even single wide file name in the current buffer results into a wide window. Although that file might be invisible in the current window view.
- Users can scroll horizontally with `zL` and `zH`, with special care put to ensure that this scrolled view is preserved during vertical scroll.

I'll take time to think about this, but at the moment I have doubts this will be included.

smilovanovic:
Backwards compatibility is there, whoever wants to have fixed small width value can do so.
There aren't many people who are aware of zL and zH.
Maybe we can get to a compromise regarding the way this feature is enabled?
Instead of setting individual width props to 0 we can have a separate option like `dynamic_width: boolean` which will be applied to all three window types (nofocus, focus and preview)?

Anyway, I love your plugin, great job and an idea to edit things like it's just another text buffer

echasnovski:
> Backwards compatibility is there, whoever wants to have fixed small width value can do so.
> There aren't many people who are aware of zL and zH.
> Maybe we can get to a compromise regarding the way this feature is enabled?
> Instead of setting individual width props to 0 we can have a separate option like `dynamic_width: boolean` which will be applied to all three window types (nofocus, focus and preview)?

Backwards compatibility is always a concern and it is indeed present here.

What also is a primary concern is a "functionality - maintainability" balance. The 'mini.files' module is already quite big, both in terms of its functionality and lines of code. Adding more will mean more long-term maintenance, which also includes documentation and tests (which are needed plenty as there are quite a few edge cases to test).

Together with some usability concerns I expressed earlier, this currently makes me hesitant to add this.

> One more question. Are contributions to this repo automatically applied to module specific repos as well? https://github.com/echasnovski/mini.files in this case

If merged, it will be applied later during (manual) synchronization.

echasnovski:
Thanks again for taking your time and implementing this PR!

However, after trying this out I still concur the reasoning in [this comment](https://github.com/echasnovski/mini.nvim/pull/780#issuecomment-2036878883). Dynamic width makes navigation a lot less visually consistent than I'd like it to be. It might be OK with only two windows (although I still don't really like it) but it is definitely not OK with bigger number of windows.

Closing as not planned

------

Couple of suggestions in case you'd want to have this in your own fork:
- I'd use `max = math.max(max, vim.fn.strdisplaywidth(v))` to compute maximum line width inside the cycle, as it gets a better width in presence of icons.
- The resulting width in non-preview windows will always be a bit greater than needed due to how 'mini.files' tracks its content by [prepending each line with concealed path identifier](https://github.com/echasnovski/mini.nvim/blob/32b2db5ef9066f3d7ca375ea1ccc0fc44090727d/doc/mini-files.txt#L236-L238). This can also be taken into account with addition `string.match()`, though.

Issue #787, opened 2024-04-07
[mini.starter] sessions are broken when a file named Starter is in the session


echasnovski:
### Discussed in https://github.com/echasnovski/mini.nvim/discussions/786

<div type='discussions-op-text'>

<sup>Originally posted by **ldelossa** April  7, 2024</sup>
I'm wondering if this is an issue.

I have auto-session working as a session manager, and I use it through mini.starter.
I don't think this has to do with what session manager is in use tho. 

I think if you're using mini.starter and you do the following with a session manager:
1. Open nvim
2. Open one or two other files
3. Create a source file in your repo called Starter
4. Save your session

When you go to restore this session in mini.starter I get this:

```
Error detected while processing CursorMoved Autocommands for "<buffer=1>":                              
Error executing lua callback: ...nvim/site/pack/deps/start/mini.nvim/lua/mini/starter.lua:1251: Cursor p
osition outside buffer                                                                                  
stack traceback:                                                                                        
        [C]: in function 'nvim_win_set_cursor'                                                          
        ...nvim/site/pack/deps/start/mini.nvim/lua/mini/starter.lua:1251: in function 'position_cursor_o
n_current_item'                                                                                         
        ...nvim/site/pack/deps/start/mini.nvim/lua/mini/starter.lua:1319: in function <...nvim/site/pack
/deps/start/mini.nvim/lua/mini/starter.lua:1319>    
``` 

I think mini.starter may get confused in this case. Would it make more sense to use a buffer name which starts with ```starter:///``` to ensure the starter buffer does not overlap with any files?</div>

echasnovski:
Yes, indeed 'mini.starter' does not check existing buffers for possibly duplicating names. I can reproduce an error related to having the buffer named the same as the one in 'mini.starter', but the error is different then what what you (@idelossa) report.

Does replacing [these lines](https://github.com/echasnovski/mini.nvim/blob/d0b297ab233a57bed00ab69bd66135dee111967f/lua/mini/starter.lua#L1339-L1340) with the following one line solve the issue for you?

```lua
  vim.api.nvim_buf_set_name(buf_id, 'starter://' .. H.buffer_number)
```

ldelossa:
Yup that works.

echasnovski:
This should be fixed on latest `main`.

PR #788, opened 2024-04-07
fix(mini-starter): use scheme name for Starter buffer

ldelossa:
When using a session manager and in the off-chance that there's a real souce file named 'Starter' in the session, restoring the session with this file from mini.starter causes an [issue](https://github.com/echasnovski/mini.nvim/discussions/786)

This is because mini.starter buffer name conflicts with the real source file named Starter.

This can be avoided by using a scheme-like name for the starter buffer which is a common practice for buffers created by plugins.

### DIFF
--- a/lua/mini/starter.lua
+++ b/lua/mini/starter.lua
@@ -1336,7 +1336,7 @@ H.apply_buffer_options = function(buf_id)
 
   -- Set buffer name
   H.buffer_number = H.buffer_number + 1
-  local name = H.buffer_number <= 1 and 'Starter' or ('Starter_' .. H.buffer_number)
+  local name = H.buffer_number <= 1 and 'mini.starter://Starter' or ('mini.starter://Starter_' .. H.buffer_number)
   vim.api.nvim_buf_set_name(buf_id, name)
 
   -- Having `noautocmd` is crucial for performance: ~9ms without it, ~1.6ms with it


echasnovski:
Thanks for the PR!

Would you mind answering the #787 first?

ldelossa:
Sure thing, answered. And yes it works.

echasnovski:
OK, thanks!

I quite like how "Starter" name is shown in statusline, while "starter://1" does not look too good (especially for new users). I'll take a moment to think if preserving current naming is important or not.

ldelossa:
I think the PR i have here is a good compromise no? 

This is what other plugins do "[plugins]://[BufferName]" 

This pr has that format ^, FWIW.

echasnovski:
Thanks again for the PR!

I decided to go a slightly more complicated route of trying to use current 'Starter_x' format as it is more user friendly to see in the statusline.

echasnovski:
> It relies on finding "Starter" in the name, can i safely move this to looking for "starter://" string in buffername everytime?

No, as 'mini.starter' still prefers "Starter" as the buffer name in most situations. The most reliable way to tell if the buffer is from 'mini.starter' is to check that it has 'starter' filetype. So like `vim.bo.filetype == 'starter'`.

ldelossa:
Ah I see your solution there. 

Ugh, from a integration point of view, this is kinda bad. 

Plugins or scripts which want to look at the starter page now has to check both for plane "Starter" and "starter://" in buf name, and it means if Starter is a real file in the repo, matching becomes inconvenient :/

it would be really helpful to just stick to a scheme so users (especially with regard to session management) who write scripts against mini.starter can always identify the starter buffer

echasnovski:
> The most reliable way to tell if the buffer is from 'mini.starter' is to check that it has 'starter' filetype. So like `vim.bo.filetype == 'starter'`.

echasnovski:
@ldelossa, just an FYI. Resolving #1455 indeed moved to a consistent naming for buffers that are created by modules. I still don't like the fact that statusline will not show `Starter` as buffer name (it is `ministarter://1/welcome` now), but that is the price for a consistent application of a general principle.

The `vim.bo.filetype == 'ministarter'` is still the most reliable way to identify the buffer created by 'mini.starter', though.

ldelossa:
Awesome! While its a bit ugly, it seems to be the canonical way that plugins are moving too. Thanks for the heads up. Tho, i did think using the filetype to inform if a special buffer was adequate as well.

PR #795, opened 2024-04-10
ci(lintcommit): check out HEAD instead of merge commit

dundargoc:
Tasks that depend on git history such as lintcommit cannot be used with
the regular checkout as that defaults to github's own merge branch which
doesn't reflect the real history. Instead we need to check out the
actual branch of the users PR.


echasnovski:
Thanks for the PR!

These GitHub merge branches are often more confusing than useful for me. Thankfully, I've got an expert contributor watching my back here :)

Issue #798, opened 2024-04-10
mini.pick doesn't open URI paths generated by the jdtls language server for lsp actions (definition, implementation etc.)


zydras07:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

mini.pick

### Description

Potentially following from #714, mini.pick will not open URI paths generated by jdtls for decompiled class files from lsp actions. I am not sure whether this is specific to jdtls or more general.

### Neovim version

0.10.0-dev-2826+gf6dcc464f

### Steps to reproduce

1. Ensure mini.pick, mini.extra and the jdtls language server are installed and configured.
2. Create a new Java project.
3. In `App.java`, write
```java
package org.example;  // relevant package name

public class App {
    public static void main(String[] args) {
        System.out.println("-");
    }
}
```
4. Move the cursor to `System` and call `require("mini.extra").pickers.lsp({ scope = "definition" })`.
5. Select the first and only option in the picker.

### Expected behavior

The URI should be opened in Neovim.

### Actual behavior

The URI is not opened, and `}` is printed below the statusline.

echasnovski:
Thanks for the issue!

I don't have Java stack installed (and don't plan to at the moment), so would you mind helping me debug this? As `MiniExtra.pickers.lsp()` uses corresponding `vim.lsp.buf` methods, could execute the following command in the same conditions as when you'd call `lsp` picker:

```
-- Execute `vim.lsp.buf.definition()` but save the result instead of acting on it
:lua vim.lsp.buf.definition({ on_list = function(...) _G.args = { ... } end })
```

Then follow up with `:lua vim.fn.append(0, vim.split(vim.inspect(args), '\n'))` which will paste `args` as text at the top of current buffer. Copy those lines and paste them here.

zydras07:
```
{ {
    items = { {
        col = 20,
[...]

PR #803, opened 2024-04-11
fix(pick): added 'uri' destination type to open non-local destinations

zydras07:

Used to open buffers associated with URIs that do not exist locally. Removed check for buffer being a listed buffer, may have unintended consequences. Tested with jdtls lsp actions (go to definition, go to reference etc.).

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -3105,7 +3105,7 @@ H.choose_path = function(win_target, item_data)
   -- `:edit` call and avoids some problems with auto-root from 'mini.misc'.
   local path, path_buf_id = item_data.path, nil
   for _, buf_id in ipairs(vim.api.nvim_list_bufs()) do
-    local is_target = H.is_valid_buf(buf_id) and vim.api.nvim_buf_get_name(buf_id) == path and (vim.bo[buf_id].buflisted or pcall(vim.uri_to_fname, buf_id))
+    local is_target = H.is_valid_buf(buf_id) and vim.bo[buf_id].buflisted and vim.api.nvim_buf_get_name(buf_id) == path
     if is_target then path_buf_id = buf_id end
   end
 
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Is it related to #798? Does this PR solve that issue?

zydras07:
Yes, it was written to solve that.

echasnovski:
> Yes, it was written to solve that.

Sorry to nit pick, but *does* it solve that issue? As in did you try the steps from the issue and everything works as expected?

My main reason for asking this is that it uses `filename` with URI of the format `jdt:xxx`. While `vim.uri_to_fname()` [works only for `file:xxx` format](https://github.com/neovim/neovim/blob/4459e0cee8b6d043ab2b06cbd89545c45a76a612/runtime/lua/vim/uri.lua#L104-L106).

zydras07:
Everything works as expected on following the steps from the issue. Individually testing the `vim.uri_to_fname` function, it only throws when the URI regex doesn't match, which includes strings from any URI format.

echasnovski:
> Everything works as expected on following the steps from the issue. Individually testing the `vim.uri_to_fname` function, it only throws when the URI regex doesn't match, which includes strings from any URI format.

Ah, I see.

My main concern right now is that this introduces a lot of new code lines when overall 'mini.pick' line count is taken into account. If it were significantly less *while* not removing `vim.bo[buf_id].buflisted` check, I'd consider adding this.

For example, could `H.preview_uri` utilize `vim.uri_to_bufnr()`?

echasnovski:
@zydras07, as I don't have a way to check if this works, can you confirm that with the current state of PR (e3473a205fdb4c8e7eb757e90395f5ed01e3eebc) everything works as expected? Looking at code, I still have some doubts (like `pcall(vim.uri_to_fname, buf_id)` check should always by `false` and thus useless).

zydras07:
Apologies, the check with `buflisted` was indeed unneeded. The intention was to check whether the buf_id was a URI if it were unlisted (which should have used `vim.uri_from_bufnr` regardless, my bad) but, at least in jdtls, the URI becomes a listed buffer prior to jumping, so the original buflisted check can remain.

echasnovski:
Thanks for the PR!

I've polished the code and added tests. This should now be in current `main` branch.

PR #847, opened 2024-05-01
docs(map): add a "scrollbar with signs" config example

dimhatz:
- [x ] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [x ] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

First of all, thank you for this nice set of libraries! I really enjoy using them and also greatly appreciate your goal to keep the code minimal and maintainable. So many plugins today are too complex, resulting in weird bugs and bad performance. Yours are a pleasant exception. I am hoping they will stay that way :)

This PR adds a configuration example for `mini.map`, that makes it look like a "pure scrollbar", but with signs (see screenshot).
I consider it to be a common case, where someone may want a scrollbar (without the map), but with signs that show errors, git added / removed, search hits.

This config may seem a bit "creative", but IMHO it's better to use this, than adding another config option (and complexity).

Cheers!

P.S. regarding when to trigger `MiniMap.open()`: I have added `UIEnter` next to `VimEnter` event, since this one is that worked for me on windows with alacritty.

minimap-scrollbar-fancy

### DIFF
--- a/doc/mini-map.txt
+++ b/doc/mini-map.txt
@@ -57,7 +57,8 @@ What it doesn't do:
   |MiniMap.toggle_side()|. Works best with global statusline.
 - Provide autoopen functionality. Due to vast differences in user preference
   of when map window should be shown, set up of automatic opening is left to
-  user. A common approach would be to call `MiniMap.open()` on |VimEnter| event.
+  user. A common approach would be to call `MiniMap.open()` on |VimEnter| or
+  |UIEnter| event.
   If you use |MiniStarter|, you can modify `<CR>` buffer mapping: >
 
   vim.cmd([[autocmd User MiniStarterOpened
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

Although visually it looks interesting, this looks like too much text for something that is not a primary use case of 'mini.map'. There is already a "pure scrollbar config" section, which I think should be enough.

------

Couple of notes:
- Overriding the module's function should never be the problem solution. I believe the same result can be achieved by using `config.symbols.encode`:
  ```
  ...
        encode = { '▐', '▐', resolution = { row = 1, col = 1 } },
  ...
  ```
- I am not sure why `UIEnter` might work but `VimEnter` might not. I'd rather have it a single advice until there are more details.

dimhatz:
Thank you for your feedback! 
The snippet you provided resulted in exactly the same visual result.
I somehow missed the point of `config.symbols.encode` and could not see it as the solution.

Regarding `VimEnter` vs `UIEnter`, I will investigate further and open a ticket if needed.

PR #852, opened 2024-05-02
fix(diff): update for breaking change on nightly

gpanders:
https://github.com/neovim/neovim/pull/28101 renamed nvim__buf_redraw_range to nvim__redraw.


echasnovski:
Thanks for the PR!

Just noticed it couple of minutes ago :) In current form it will break earlier Neovim versions (as seen in CI). I would rather have a separate helper for the whole buffer (which is the only way it is used in 'mini.diff' right now) which uses the new `vim.api.nvim__redraw()` if it is present.

gpanders:
Pulled out into a `redraw_buffer` helper. Let me know if any other changes are needed.

echasnovski:
> Pulled out into a `redraw_buffer` helper. Let me know if any other changes are needed.

Only style nit picking which I won't bother you with and will do myself. Thanks for the quick reaction!

PR #853, opened 2024-05-02
fix(diff): redraw statusline with nvim__redraw if available

gpanders:
Redrawing the statusline with :redrawstatus causes visible flicker when text changes are made rapidly (e.g. writing in Insert mode) and when the `delay.text_change` config option is set to a low value (e.g. 100). Instead, we can redraw the statusline at the same time that we redraw the buffer with nvim__redraw.


echasnovski:
Thanks for the PR! Looks good.

PR #863, opened 2024-05-08
WIP feat(clue): custom key labels

rafamadriz:

Hello, thank you for this project it's amazing. I'm trying to add custom key labels. Reason is that some keys take to much
width and make the popup window look weird if all of the other keys are a single character, which is generally the case except for `<Space>` and similar keys.

This is in the same spirit of:
https://github.com/folke/which-key.nvim/issues/77
https://github.com/folke/which-key.nvim/commit/2be929e34b2f2b982e6b978c0bd94cd2e1d500e6  

Before:
Screenshot from 2024-05-07 21-56-31

After: 
Screenshot from 2024-05-07 21-55-34

This PR is not complete, I did manage to make it work in 1ea0b267f7a6242f8f3976aea2ceb0e087d6ee16 but can't figure out how to add the option to `MiniClue.config` and make it work in `setup`, this is my attempt 4be9cdf76d3529f800761a0fe6b5e113fe3d54ab but I'm really not familiar with how this works and had quite a hard time just to get there. So I decided to stop and first see if you're open to this. If you think this fits for the project and can point me in the right direction then I'd be more than happy to finish it.

### DIFF
--- a/lua/mini/clue.lua
+++ b/lua/mini/clue.lua
@@ -577,6 +577,9 @@ MiniClue.config = {
   -- Array of extra clues to show
   clues = {},
 
+  -- Custom key labels
+  key_labels = {},
+
   -- Array of opt-in triggers which start custom key query process.
   -- **Needs to have something in order to show clues**.
   triggers = {},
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

I remember having this problem during development and was contemplating to adding this kind of functionality. In the end I decided to consistently go with the same key descriptions as are officially used in Neovim (see `:h key-codes`) and have a handy helper function to achieve that (`:h keytrans()`). Besides, 'mini.clue' setup is already quite verbose, so adding new not ground breaking options does not feel right.

The resulting extra width is indeed unfortunate, but I think it is worth it in the long term.

Closing as not planned.

PR #889, opened 2024-05-18
feat(base16): allow to only use cterm colors

### LINKED ISSUE #456: mini.base16: support default cterm color map when no termguicolors
### Contributing guidelines

### Module(s)

base16

### Description

I'm currently using the base16 mini module because it has a fairly comprehensive set of highlight rules for plugins that I rely on. However, I am using a base16 (256 color) theme for my terminal with termguicolors turned off, this gives me consistent colors accross all terminal applications.

The current base16 module however (at least as I understood the docs) is geared towards termguicolors and only attempts to choose close color codes for cterm fallback, rather than providing the actual base16 color codes for a properly configured base16 terminal.

I am working around this with something that is a bit of a hack:

```
                base16 = {
                    plugins = { default = true },

                    -- No effect in termguicolors so values don't matter, but must be set.
                    palette   = {
                        base00 = "#000000",
                        base01 = "#000000",
                        base02 = "#000000",
                        base03 = "#000000",
                        base04 = "#000000",
                        base05 = "#000000",
                        base06 = "#000000",
                        base07 = "#000000",
                        base08 = "#000000",
                        base09 = "#000000",
                        base0A = "#000000",
                        base0B = "#000000",
                        base0C = "#000000",
                        base0D = "#000000",
                        base0E = "#000000",
                        base0F = "#000000",
                    },

                    -- Assign bases to the official extended base16 color codes.
                    use_cterm = {
                        base00 = 0,
                        base01 = 18,
                        base02 = 19,
                        base03 = 8,
                        base04 = 20,
                        base05 = 7,
                        base06 = 21,
                        base07 = 15,
                        base08 = 1,
                        base09 = 16,
                        base0A = 3,
                        base0B = 2,
                        base0C = 6,
                        base0D = 4,
                        base0E = 5,
                        base0F = 17,
                    }
                },
```

The gui colors don't matter because I don't run in gui mode anyway so I set them all black, and the use_cterm being set to terminal color codes makes [this code here](https://github.com/echasnovski/mini.nvim/blob/main/lua/mini/base16.lua#L1170) properly assign the right terminal colors as I would have expected the defaults to be (see [this base16-shell script example](https://github.com/chriskempson/base16-shell/blob/master/scripts/base16-dracula.sh#L7) to see official the termcolor <-> base assigns).

With this I can change my terminal-wide theme and vim changes as expected (along with all other terminal apps). This seems like an intended way to use base16 in general so I wanted to see if this default base to termcolor map could be supported by the plugin itself, perhaps through `use_cterm = 'default'` or some other hint.

arcxio:
Make `config.palette` optional if `config.use_cterm` is provided as a table with cterm colors.

Resolve #456

### DIFF
--- a/doc/mini-base16.txt
+++ b/doc/mini-base16.txt
@@ -143,9 +143,9 @@ groups make an extended set from original
 good idea to have `config.palette` respect the original [styling
 principles](https://github.com/chriskempson/base16/blob/master/styling.md).
 
-By default only 'gui highlighting' (see |highlight-gui| and
-|termguicolors|) is supported. To support 'cterm highlighting' (see
-|highlight-cterm|) supply `config.use_cterm` argument in one of the formats:
+To support 'gui highlighting' (see |highlight-gui| and |termguicolors|), supply
+`config.palette`. To support 'cterm highlighting' (see |highlight-cterm|),
+supply `config.use_cterm` argument in one of the formats:
 - `true` to auto-generate from `palette` (as closest colors).
 - Table with similar structure to `palette` but having terminal colors
   (integers from 0 to 255) instead of hex strings.
@@ -155,7 +155,8 @@ Parameters ~
 
 Usage ~
 `require('mini.base16').setup({})` (replace `{}` with your `config`
-  table; `config.palette` should be a table with colors)
+  table; either or both of `config.palette` and `config.use_term` should be a
+  table with colors)
 
 ------------------------------------------------------------------------------
                                                              *MiniBase16.config*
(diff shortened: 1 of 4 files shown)


echasnovski:
Thanks for the PR!

> my solution is simpler and more obvious and would make gui and cterm highlighting symmetrical, which IMO makes more sense than arbitrarily treating cterm highlighting as fallback-only second-class citizen, considering `termguicolors` is disabled by default

I appreciate the effort, but this is exactly what 'mini.base16' does and will keep doing: "gui highlighting" is the main concern for color scheme while "cterm highlighting" is a fallback. The reason for this is the ever growing number of people who use terminal emulators with "gui highlighting" support. This is now even more reasonable choice on Neovim>=0.10 as there is an automated 'termguicolors' detection depending on terminal emulator capabilities.

Closing as not planned.

arcxio:
> The reason for this is the ever growing number of people who use terminal emulators with "gui highlighting" support

the goal is not to support terminal emulators without gui highlighting, but to avoid dublicating base colors for a terminal emulator and neovim and to make neovim's colors consistent with every other terminal application, which is one of the main selling points of base16.

echasnovski:
> this is not a good design choice to force users to hack around the plugin for no good reason.

Nobody forces anyone to hack anything. 'mini.base16' is a color scheme generator that applies highlight groups based on the supplied hex colors. Support for 0-15 colors is secondary. This is a design choice aimed for an easier long term maintainability.

arcxio:
> This is a design choice aimed for an easier long term maintainability

based on the docs I didn't get that supplying hex colors is a core to mini.base16's design and I didn't think that making palette optional would hurt long term maintainability, I wouldn't submit this if I realized that, but I don't feel the full scope of your work of course, and I respect that. thanks for your consideration.

PR #894, opened 2024-05-21
docs(mini.test): adding note about lua-language-server related config 

MagicDuck:
- [ x ] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [ x ] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

Spent like a couple of hours getting completion working for MiniTest and I figured I would add a note to the docs since it greatly helps with discoverability of the api:

here's a screenshot of it in action:
image

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -136,6 +136,7 @@ Overview of full file structure used in for testing 'hello_lines' plugin:
 │   └── hello_lines
 │       └── init.lua # Mandatory
 ├── Makefile # Recommended
+├── .luarc.json # Recommended
 ├── scripts
 │   ├── minimal_init.lua # Mandatory
 │   └── minitest.lua # Recommended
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

I didn't bother setting up completion because it mostly a handful of methods most of which mimic Neovim's API. But I am not against adding this to 'TESTING.md' indeed. I'll get back to it after testing it out myself.

MagicDuck:
I just realized that doing this will screw up the autocomplete for the rest of the nvim API (like for example, `vim.api.nvim_buf_set_lines()`). 

image


I wonder if there is a way to append to the workspace library settings 🤔

MagicDuck:
Yah, there is no way to make it work with both the vim.api.... types injected by neodev and add have a custom `.luarc.json` unfortunately. The workaround that I've found is to configure neodev like so:
```lua
      require('neodev').setup({
        override = function(root_dir, options)
          if vim.startswith(root_dir, '/path/to/your/repos/') then
            options.plugins = { 'mini.test' }
            return options
          end
        end,
      })
```
So am abandoning this PR, but feel free to include that snippet somewhere in TESTING.md if you feel like it. I personally think it's quite important to have completion for both beginners like me and for easy recall / saving on typing.

echasnovski:
Yeah, I am not opposed to having in 'TESTING.md' a suggestion on better completions, but also not really invested in investigating how to do that properly.

Closing as "would be good to have, but might require a deep dive into how Lua-LS does things".

PR #900, opened 2024-05-21
feat(deps): react to `vim.loop.cpu_info` returning `nil`

lrntgr:

## Problem
When running Neovim in a [Termux](https://termux.dev/en/) environment, the `vim.loop.cpu_info` function is returning `nil`.  
Hence, running `DepsUpdate` fails with a `attempt to get length of a nil value` message.

## Solution
Ensure that `vim.loop.cpu_info()` is returning a table before getting its length: if not a table, use `1` as a fallback value of number of processors.

### DIFF
--- a/lua/mini/deps.lua
+++ b/lua/mini/deps.lua
@@ -1459,7 +1459,12 @@ end
 -- CLI ------------------------------------------------------------------------
 H.cli_run = function(jobs)
   local config_job = H.get_config().job
-  local n_threads = config_job.n_threads or math.floor(0.8 * #vim.loop.cpu_info())
+  local n_threads = config_job.n_threads
+  if not n_threads then
+    local cpus = vim.loop.cpu_info()
+    local n_cpus = (type(cpus) == 'table') and #cpus or 1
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

This is interesting behavior. I'll resolve this probably tomorrow.

------

How is developing on Termux? Is is snappy enough with 'mini.nvim'?

lrntgr:
> Thanks for the PR!
> 
> This is interesting behavior. I'll resolve this probably tomorrow.
> 
> How is developing on Termux? Is is snappy enough with 'mini.nvim'?

I think 'mini.nvim' is a good fit for Termux: minimal but powerful, I love it (_mini.pick in particular_)!  
I've not been disturbed by latency issues so far, but at this point my config is pretty simple (no LSP for instance). The limitation is more about the small screen and the keyboard.

echasnovski:
This should now be resolved on latest `main`. Thanks for the PR and interesting use case!

lrntgr:
> This should now be resolved on latest `main`. Thanks for the PR and interesting use case!

Great, thanks @echasnovski for your awesome work.

PR #907, opened 2024-05-23
fix(statusline)!: make compute_attached_lsp() depend only number of attached client

xuanduc987:
`H.get_buf_lsp_clients` will return a map of `client_id:client`. So if we handle it as array, there could be hole in it, and `table.concat` will fail with "invalid value (nil) at index [i] in table for 'concat'".

### DIFF
--- a/lua/mini/statusline.lua
+++ b/lua/mini/statusline.lua
@@ -631,8 +631,8 @@ if vim.fn.has('nvim-0.8') == 0 then
 end
 
 H.compute_attached_lsp = function(buf_id)
-  local names = vim.tbl_map(function() return '+' end, H.get_buf_lsp_clients(buf_id))
-  return table.concat(names, '')
+  local count = vim.tbl_count(H.get_buf_lsp_clients(buf_id))
+  return string.rep('+', count)
 end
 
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Do you have any reproducible example of when you encounter the error? Documentation of any core functions used in `H.get_buf_lsp_clients()` says that it returns "list" (a.k.a. array without holes).

I am not objecting the change per se, as it is cleaner. But having reproducible example will mean the need to add a test.

xuanduc987:
I'm still on neovim 0.9.4, so `H.get_buf_lsp_clients()`  will call `vim.lsp.buf_get_clients` (returns a map)  instead of `vim.lsp.get_clients` (returns a list)

I've just tried to change from vim to neovim several days before, so I still don't understand my whole lsp server setting (seems to auto configured with mason?). But I think if we attach several different lsp server for the same file type, then the bug will show up almost surely on neovim version < 0.10

echasnovski:
Yeah, I missed that 'mini.statusline' uses `vim.lsp.buf_get_clients()` which indeed returns a map.

I've merged this in a temporary branch to polish a bit.

PR #937, opened 2024-06-03
feat(mini.files): Add customization for title position and new highlight for modified buffer window

augustocdias:

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -87,6 +87,7 @@ more details.
 
 * `MiniFilesBorder` - border of regular windows.
 * `MiniFilesBorderModified` - border of windows showing modified buffer.
+* `MiniFilesModified` - basic foreground/background for modified buffer
 * `MiniFilesCursorLine` - cursor line in explorer windows.
 * `MiniFilesDirectory` - text and icon representing directory.
 * `MiniFilesFile` - text representing file.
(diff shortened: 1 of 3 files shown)


echasnovski:
Thanks for the PR!

------

Modification of floating window title should be achieved the same way as [suggested window customization](https://github.com/echasnovski/mini.nvim/blob/e736fe131e6c18d39eecffb43d582c3e95550a90/doc/mini-files.txt#L372-L385) and not through special option.

Right now autocommand approach is not really straightforward, I'll look into it. **But** I don't think changing title position is a good UI, so it won't be top priority.

------

If I understand correctly, `MiniFilesModified` is intended to highlight the whole buffer inside window? If yes, then I think this is too much visual feedback. Changing only border highlight was a deliberate choice.

------

Closing as no suggested functionality aligns well with 'mini.files' design.

augustocdias:
Yeah. The auto command also adds overhead and complexity that I'd like to avoid in my config. Maybe exposing a callback would be a better option

echasnovski:
Using autocommands **is** and will be a suggested way to customize how windows look in 'mini.files'. This is mostly to avoid having to expose too much configuration in `setup()` while allowing much more flexibility.

echasnovski:
There was an issue with `title_pos` not being preserved after windows are refreshed, so setting it in `MiniFilesWindowOpen` was not possible (unlike `border` setting).

This should be fixed on latest `main`. The suggested approach to tweaking both `border` and `title_pos` is now [better documented](https://github.com/echasnovski/mini.nvim/blob/d330f2639462084d2ef6c699ccd6219b81c45bc7/doc/mini-files.txt#L372-L387).

PR #997, opened 2024-06-30
feat(extra): add flag 'current_dir' to `oldfiles()` picker

abeldekat:

Reference: This [discussion](https://github.com/echasnovski/mini.nvim/discussions/995)

In my workflow, each project is opened in a `tmux` session. As such, when querying `oldfiles`, I am mostly interested in files inside the current working directory.

`MiniExtra.pickers.oldfiles()` currently does not filter on `cwd`

This PR adds flag "current_dir" to `MiniExtra.pickers.oldfiles()`

### DIFF
--- a/doc/mini-extra.txt
+++ b/doc/mini-extra.txt
@@ -640,7 +640,9 @@ Pick from |v:oldfiles| entries representing readable files.
 
 Parameters ~
 {local_opts} `(table|nil)` Options defining behavior of this particular picker.
-  Not used at the moment.
+  Possible fields:
+  - <current_dir> `(boolean|nil)` - whether to return files only from
+    current working directory and its subdirectories. Default: `false`.
 {opts} `(table|nil)` Options forwarded to |MiniPick.start()|.
 
 Return ~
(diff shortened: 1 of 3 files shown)


abeldekat:
Remarks:

Name `current_dir`. Only the value of variable `current_dir` indicates that this is a boolean.

System call `vim.loop.os_uname()` is always executed, also when `current_dir` is not supplied. 
Would [this approach](https://github.com/abeldekat/nvim_pde/blob/6b892a934a0dec45e73c4eeef3eaf8f47f6d560c/lua/ak/config/ui/mini_starter.lua#L8) be appriopriate?

echasnovski:
Thanks for the PR!

I appreciate the exact copy of the approach from 'mini.starter' (I'd usually do that too), however I think the better approach here is to use plain `vim.startswith()` with `cwd .. '/'`. As far as I remember, built-in paths coming from Neovim always use forward slash. If not, we'll wait for a Windows user to report.

I've merged in the temporary branch to polish up a bit.

echasnovski:
Thanks again for following through!

This should now be part of latest `main`.

abeldekat:
Thank you!

PR #1004, opened 2024-07-01
fix(hipatterns): compute_hex_color_group: cache per style and set default style

folke:
Hi!

Two minor fixes for `compute_hex_color_group`:

- The docs mention that `style='bg'` is the default, but the code does not work without a default
- When using the same colors with multiple styles (different highlighters even), it doesn't use the correct style.

Having a lot of fun with hipatterns for developing tokyonight :)

image

### DIFF
--- a/lua/mini/hipatterns.lua
+++ b/lua/mini/hipatterns.lua
@@ -622,7 +622,7 @@ end
 ---   |:colorscheme| command to clear cache. Needs a call to |MiniHipatterns.setup()|.
 ---
 ---@param hex_color string Hex color string in format `#rrggbb`.
----@param style string One of:
+---@param style? string One of:
 ---   - `'bg'` - highlight background with `hex_color` and foreground with black or
 ---     white (whichever is more visible). Default.
 ---   - `'fg'` - highlight foreground with `hex_color`.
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR! Glad 'mini.hipatterns' is useful in so high stakes workflow :)

The highlight group change should be fine, but it needs test for when it is useful and adjusting existing ones. I am a bit busy today, so will take a look in a couple of days.

echasnovski:
Thanks again for suggesting this!

I've planned to release 'mini.icons' today, but polishing proved to be a bigger challenge (plus *someone* announced 'tokyonight.nvim' rewrite on Neovim and I don't want 'mini.icons' announcement to be in its shadow :cry:). So I've got some time on my hands.

This now should be resolved on latest `main` (both 'mini.nvim' and standalone 'mini.hipatterns'). Hope this will simplify the "Dev Environment" of 'tokyonight.nvim'.

echasnovski:
> Wait? Is **that** what those `MiniIcons` groups are for? I've been looking through your code everywhre can couldn't find where I had to put them.
> 
> They're currently under `mini.hue` 😅
> 
> https://github.com/folke/tokyonight.nvim/blob/main/lua/tokyonight/groups/mini_hue.lua#L7

Yeah, adding those groups was the primary motivator for early second round of contributing to all popular color schemes. Those highlight groups work OK-ish with default links but are sometimes not quite good for some color schemes.

And I [mentioned an upcoming module in PR](https://github.com/folke/tokyonight.nvim/pull/545#issue-2368779858) :)

folke:
I totally missed that! :)
Will move the groups to a new plugin then.

folke:
Fixed! https://github.com/folke/tokyonight.nvim/commit/5808dbed0682270f0c40ae7a1bc517ad1fbcaede

PR #1015, opened 2024-07-05
feat(icons): add more os icons

tranzystorekk:
Not sure if these icons can be accepted since they are found in the `nf-linux-*` class, but they're arguably popular linux distros so I decided to take my shot :)

### DIFF
--- a/lua/mini/icons.lua
+++ b/lua/mini/icons.lua
@@ -1618,6 +1618,7 @@ H.lsp_icons = {
 -- this feels representative of "popular" operating systems.
 --stylua: ignore
 H.os_icons = {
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Category "os" during development contained all operating system icons with `nf-linux-*` classes. This was a lot, so I decided to use only "popular" ones. The appropriate metric for this seems like "OS is popular if it has a dedicated `nf-md-*` icon". Although I agree that Alpine, PopOS, and Void do feel like a popular OSs, I'd like to stick to that rule. In the end, I've even not included OS *I* am using :)

If there are `nf-md-*` icons for OSs which are not currently present (due to I have missed them or they were just added), do feel free to PR them (both you and anyone else reading this).

Closing as not planned.

PR #1016, opened 2024-07-05
fix(icons): ensure highlights are setup when changing colorschemes

mehalter:
This adds an autocommand on the `ColorScheme` event to make sure highlights are created if necessary for `mini.icons`

### DIFF
--- a/lua/mini/icons.lua
+++ b/lua/mini/icons.lua
@@ -1674,6 +1674,14 @@ H.apply_config = function(config)
   H.init_cache(config)
 end
 
+H.create_autocommands = function()
+  local augroup = vim.api.nvim_create_augroup('MiniIcons', {})
+  vim.api.nvim_create_autocmd(
+    'ColorScheme',
+    { group = augroup, callback = H.create_default_hl, desc = 'Ensure proper colors' }
+  )
+end
+
 --stylua: ignore
 H.create_default_hl = function()
   local hi = function(name, opts)


echasnovski:
Thanks for the PR!

Highlight groups are created linked to some built-in highlight groups via default links (as in, `:h :hi-default`). Those links survive `:hi clear` if they were not overridden. *If* they are overridden, then managing them is laid on user's shoulders.

This is a common pattern for all modules which define highlight group via default link. So either they all should react to `ColorScheme` or none of them. I'll think about it.

In the meantime, closing as not planned (in this form; yet).

mehalter:
This is quite an interesting take. It seems to be that default highlight groups in other mini modules are updated on `ColorScheme`. Also nearly all colorschemes that come to mind run `:hi clear` on being set to make sure they do not have interaction with other colorschemes. Popular colorschemes such as TokyoNight by folke, Catppuccin, Nightfox, etc. all do this and do _not_ work given the current pattern in `mini.icons`, but do luckily work with other mini modules such as `mini.test` since it does set it's defaults on `ColorScheme` 🤔

mehalter:
Is there a reason to maintain different stances on the `ColorScheme` autocommand and default highlight situation in the same ecosystem? This block of code was copied exactly from your other modules

echasnovski:
> This is quite an interesting take. It seems to be that default highlight groups in other mini modules are updated on `ColorScheme`. 

> Is there a reason to maintain different stances on the `ColorScheme` autocommand and default highlight situation in the same ecosystem? This block of code was copied exactly from your other modules

This *is* is consistent. Here is once again what is the behavior: "This is a common pattern for all modules which define highlight group **via default link**.". All cases in 'mini.nvim' of adding autocommand for `ColorScheme` event contain at least one group which is *not* defined via default link. As those won't indeed survive `:hi clear`, `ColorScheme` event autocommand is needed.

> Also nearly all colorschemes that come to mind run `:hi clear` on being set to make sure they do not have interaction with other colorschemes. Popular colorschemes such as TokyoNight by folke, Catppuccin, Nightfox, etc. all do this and do _not_ work given the current pattern in `mini.icons`, but do luckily work with other mini modules such as `mini.test` since it does set it's defaults on `ColorScheme` 🤔

Indeed calling `:hi clear` at the start of color scheme is a good approach. And again, all those color schemes *do* work as described in the previous comment: "Those links survive :hi clear if they were not overridden. *If* they are overridden, then managing them is laid on user's shoulders.". Some popular color schemes now indeed override 'mini.icons' colors. Hence if the user wants to change several completely different color schemes in one session (which, let's be real, I don't think is a common use case worth optimizing for), they should take care of the overridden highlight groups.

mehalter:
Thanks for the explanation ❤️

echasnovski:
@mehalter, after thinking about this a bit longer, I think ensuring that *all* highlight groups are defined after any color scheme takes effect is indeed worth doing in 'mini.nvim' itself. The two observations that got this decision over the edge were:
- Actually realizing that not many code lines should be added to ensure this. There were only 5 modules which needed many added lines (the ones which defined highlight groups but no autocommands), while others either already have the 'ColorScheme' autocommand or a single call away from it. After some refactoring, the change resulted in on +39 lines added, which is less than 1 per module.
- Also realizing that this should greatly simplify adding support for 'mini.nvim' for color scheme authors. Since there are now 111 highlight groups, adding all of them is a bit of a hassle. And most of the default links should look fine (as they are semantically close).

mehalter:
Thanks so much for the update! Yeah totally agree with the reasoning you mentioned :)

Issue #666, opened 2024-01-23
Update mini.doc to LuaCATS


S1M0N38:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.doc

### Description

[LuaLS/lua-language-server](https://luals.github.io/) is the most used implementation Language server for Lua. One of its features are annotations implemented as `LuaCATS` (Lua Comment And Type System). As stated in their docs:

> LuaCATS annotations are [no longer cross-compatible](https://github.com/LuaLS/lua-language-server/issues/980) with [EmmyLua annotations](https://emmylua.github.io/annotation.html) as of v3

The current implementation of `mini.doc` try to parse EmmyLua-like annotations. Compare to EmmyLua annoation, `LuaCATS` define a precise specification for annotation.

For instance, the `mini.doc` :help section mentions:

> Hooks for sections which supposed to have "type-like" data ('@field',
  '@param', '@return', '@type') automatically enclose *first found*
  "type-like" word and its neighbor characters in '`(<type>)`' (expect
  false positives). Algorithm is far from being 100% correct ...

Conversely, the `@param` annotation supported by LuaLS/lua-language-server is explicitly defined as:
```
---@param <name[?]> <type[|type...]> [description]
```
In this format, the `type` directly follows the parameter `name`, eliminating the need for heuristic searches for the type. Additionally, this format simplifies the integration of user-defined types into help.txt.

----

While I could develop my own hooks to adhere to the `LuaCATS` specification (skill permitting), I believe that aligning with a dominant and well-documented standard would be more advantageous for the community.

echasnovski:
Thanks for the suggestion!

On first glance, this makes sense. Having some strict rules from 'lua-language-server' is good. I'll look into it (not very soon, I am afraid).

S1M0N38:
I'll give a shot as well. Instead of manually parse annotations line (the one starting with `---`) a viable approach is to use the json produce by the command for [docs generation](https://luals.github.io/wiki/export-docs/)

```sh
./lua-language-server --doc=path/to/lua/workspace
```

[...]

PR #1027, opened 2024-07-07
docs(icons): update `get` documentation to document all return values

mehalter:

The documentation strings should follow the LuaCATS standard for the most part which documents multiple returns with multiple `@return` comments.

Relevant links:
- Documentation for `@return`, only one that shows multiple return values is the example for variable arguemnt returns: https://luals.github.io/wiki/annotations/#return
- Example code in the LuaCATS organization in the `luv` repo of a function with multiple returns: https://github.com/LuaCATS/luv/blob/main/library/uv_signal_t.lua#L54-L61

Currently this results in documentation generated with multiple `Return` headers. I believe this is intended since this is a case explicitly tested in the `mini.doc` test suite. If we want to move this to behave more like `Parameters` that could also be nice and more clear to the user. I have this change working in a separate branch. I could add that to this PR if desired or follow up with a separate PR. There would also need to be some amount of discussion or thought of whether or not to stick with the word "Return" for the section header here or move to "Returns" since similarly to "Parameters" and "Fields", there can be many.

I know that these documentation strings are more meant to align with `mini.doc` rather than the standard LuaCATS right now and it may be a goal to move to supporting the standard in the future (#666). So the strings here still aren't "true" LuaCATS since they don't have names for the returns or the `#` indicating it's a labelled return. So let me know if this is just out of scope for the time being and will be implemented correctly later when the open issue about LuaCATS is resolved.

Let me know what you think and I can update my PR as necessary to align with the project's goals. Glad to help in any way I can 😄

### DIFF
--- a/doc/mini-icons.txt
+++ b/doc/mini-icons.txt
@@ -377,9 +377,12 @@ Parameters ~
   names which are explicitly supported for specific category.
 
 Return ~
-`([string,string,boolean])` Tuple of icon string, highlight group name it is
-  suggested to be highlighted with, and boolean indicating whether this icon
-  was returned as a result of fallback to default. Example: >lua
+`(string)` icon string
+Return ~
+`(string)` highlight group name it is
+Return ~
+`(boolean)` a boolean indicating whether this icon was returned as a
+  result of fallback to default. Example: >lua
 
   -- Results into `icon='󰢱'`, `hl='MiniIconsAzure'`, `is_default=false`
   local icon, hl, is_default = MiniIcons.get('file', 'file.lua')
--- a/lua/mini/icons.lua
+++ b/lua/mini/icons.lua
@@ -390,9 +390,10 @@ MiniIcons.config = {
 ---@param name string Icon name within category. Use |MiniIcons.list()| to get icon
 ---   names which are explicitly supported for specific category.
 ---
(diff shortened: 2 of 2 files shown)


echasnovski:
Thanks for the PR!

> I know that these documentation strings are more meant to align with `mini.doc` rather than the standard LuaCATS right now and it may be a goal to move to supporting the standard in the future (#666). So the strings here still aren't "true" LuaCATS since they don't have names for the returns or the `#` indicating it's a labelled return. So let me know if this is just out of scope for the time being and will be implemented correctly later when the open issue about LuaCATS is resolved.

Yes, I am afraid this is out of scope for the time being. I hope to eventually get back to this during #666. Nevertheless, information you linked in first comment will be useful. Thanks!

Closing as "planned but not now".

mehalter:
Ok cool thanks! It's just a shame this is now the one function in mini.nvim that is not compatible with the Lua language server at all. I'll just disable diagnostics for this module in my code

echasnovski:
After updating LuaLS and reading closely about what it means by "tuple", I realized that I misunderstood its meaning. I thought it was about separate return values as `MiniIcons.get()` returns, but in reality it is about tables with fixed number of number fields.

So the return value is now `...` (as in other modules for functions which return several separate values). This is (hopefully) a temporary solution until the 'mini.doc' update to better support LuaCATS.

PR #1031, opened 2024-07-08
Use correct mapping for elixir filetypes

calops:

Elixir files come under the extensions `.ex` and `.exs` (among others, but these are working properly).

In `mini.icons`, there is an explicit mapping for `.exs` files, even though:
- It's not the most common extension you'll come across, by far
- It already works well with `vim.filetype.match()`

However, `.ex` has no explicit mapping, even though:
- It is the most common elixir file extension you'll see
- It currently doesn't work with `vim.filetype.match()` for some reason

This PR adjusts the mapping so that both extensions are represented in their own specific ways (there is a nerd symbol for elixir scripts that's distinct from normal elixir files so I made use of that). If the mapping for `.exs` sounds too specific for the scope of this module, I guess it would also be fine to remove it entirely.


calops:
Thanks for the quick reply!

> The 'ex' extensions was not added because [it is not recognized unconditionally](https://github.com/neovim/neovim/blob/b3d94b10870ce494941279f78387e1d98a7e48e7/runtime/lua/vim/filetype.lua#L419). In particular, it [can be matched as Euphoria language](https://github.com/neovim/neovim/blob/b3d94b10870ce494941279f78387e1d98a7e48e7/runtime/lua/vim/filetype/detect.lua#L556-L557) (which indeed [also uses '.ex' extension](https://en.wikipedia.org/wiki/Euphoria_(programming_language))).

Thanks for the insight, I had no idea about euphoria, and yeah I guess it makes sense.

> I am not sure why, because for me it does. Executing :=MiniIcons.get('file', 'hello.ex') returns icon data for 'elixir' filetype.

Huh. Maybe there's something going on on my side then, I'll look into that.

> I am happy to add 'elixirscript' as separate filetype if there is evidence it is a relatively used one.

Well, it *is* commonly used if only because a typical elixir package will always come with a `mix.exs` script as the main setup utility. It's also the format used to write tests and migrations, so you'll see a lot of those in a big project.

However, it's not an actual `filetype` as far as neovim is concerned (it's still only recognized as `elixir`, which makes sense since it's the same thing within). But there *is* a dedicated icon for those so maybe we still want to make use of it? I guess it would be similar to `.h` files that are still `C` code, but come with their own mappings.

echasnovski:
> Huh. Maybe there's something going on on my side then, I'll look into that.

Would you mind adding a follow up here if/when you figure it out?

> However, it's not an actual `filetype` as far as neovim is concerned (it's still only recognized as `elixir`, which makes sense since it's the same thing within). But there _is_ a dedicated icon for those so maybe we still want to make use of it? I guess it would be similar to `.h` files that are still `C` code, but come with their own mappings.

If it is not an actual filetype (i.e. a value of 'filetype' option when you open such files), then it will not have an entry in "filetype" category.

I am open to changing "exs" extension have "" icon (i.e. "nf-seti-elixir_script" class), but I would like to ask to for some reputable sources that this is the icon that to be used for "exs" extensions.

echasnovski:
I've added special support to 'exs' extension as it seems that built-in `vim.filetype.match()` indeed doesn't differentiate between 'ex' and 'exs' (because there is no dedicated 'elixirscript' filetype, it seems). So this is similar to a precedent of 'h' extension.

Thanks for bringing this up!

calops:
Thanks! That sounds completely reasonable.

As for my previous issue, I don't really know what I did, but a minimal setup doesn't show the bug so you can disregard my comments about it.

PR #1061, opened 2024-07-15
docs: add links to modules

unrealapex:

Add links to the modules to make it convenient for people to access them without having to go inside the module's README.


echasnovski:
Thanks for the PR!

The current formatting is intentional. Each module is called that way because it is indeed a Lua module of 'mini.nvim' library. Using them this way is recommended.

The links in this PR go to standalone repositories for each module, which is a secondary form of their distribution. Having their names mentioned in module READMEs should be enough.

Closing as not planned.

PR #1067, opened 2024-07-17
feat(icons): add gleam filetype icon

DJAruun:

[Gleam](https://gleam.run/) is a new functional language that runs on Erlang and Javascript targets and I feel like it deserves an icon here for the `gleam` filetype. In the commit i used `nf-fa-star` because I think it looks more authentic to the actual [Gleam logo](https://avatars0.githubusercontent.com/u/36161205?s=400&v=4). If it really has to be from the `nf-md` I'll gladly change it, but I feel like the Font Awesome icon is more like the real logo.

**Example of it in Oil**
example in oil

### DIFF
--- a/lua/mini/icons.lua
+++ b/lua/mini/icons.lua
@@ -1087,7 +1087,7 @@ H.filetype_icons = {
   gitrebase          = { glyph = '󰊢', hl = 'MiniIconsAzure'  },
   gitsendemail       = { glyph = '󰊢', hl = 'MiniIconsBlue'   },
   gkrellmrc          = { glyph = '󰒓', hl = 'MiniIconsRed'    },
-  gleam              = { glyph = '', hl = 'MiniIconsPurple' },
+  gleam              = { glyph = '󰦥', hl = 'MiniIconsPurple' },
   gnash              = { glyph = '󰒓', hl = 'MiniIconsGreen'  },
   gnuplot            = { glyph = '󰺒', hl = 'MiniIconsPurple' },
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

> If it really has to be from the `nf-md` I'll gladly change it, but I feel like the Font Awesome icon is more like the real logo.

If the icon is not **specifically** designed for its target (like, for example, `nf-custom-elixir` for Elixir), then it is **has to be** from `nf-md-*` class. I think `nf-md-star_face` seems to be even a better fit for Gleam (based on the logo).

DJAruun:
Okay I get that, and I would say you're right with the `nf-md-star_face`. Made a new commit with that icon instead!

example

echasnovski:
Thanks!

I'll merge it once I am at my "main" machine (hopefully tomorrow).

------

Yours and a couple of several other similar PRs (for filetype which is one way or another can be a return value of `vim.filetype.match()`) made me think that initial way of getting all common filetypes (via `vim.fn.getcompletion('', 'filetype')`) might have been not full enough.
A more labor-intensive digging uncovered 112 more filetypes :scream: (which also includes 'gleam', by the way). So I guess I have slightly more work to do :(

echasnovski:
Thanks again for the PR! This should be part of `main` branch now.

lpil:
Wow! Thank you both!!

lpil:
Hello! I'm the creator of Gleam.

I've just tried this out and the icon used isn't the Gleam logo, it's some other star icon. Is it possible to get the correct logo? If not, could we get this reverted please? We don't want to use some other icon in place of our logo.

PR #1122, opened 2024-08-07
fix(move): correctly restore registers, avoid error when buffer is empty

Diomendius:

Currently, calling `move_selection()` or `move_line()` on an empty buffer causes this error:
```
E5108: Error executing lua vim/_editor.lua:0: nvim_exec2(): Vim(normal):E353: Nothing in register z
stack traceback:
        [C]: in function 'nvim_exec2'
        vim/_editor.lua: in function 'cmd'
        .../mini.nvim/lua/mini/move.lua:465: in function 'cmd'
        .../mini.nvim/lua/mini/move.lua:270: in function 'move_selection'
```
If the `z` register is set beforehand, `move_selection()` and `move_line()` insert the contents of `z` into the buffer instead of triggering the error.

This is because `mini.move` assumes that `d` and `x` always set their register, but they don't if they don't change the buffer. I think an empty buffer is the only case where this can happen in visual mode; even with `set selection=exclusive` at least one character is always selected if possible, as confusing as that is. For `move_line()` an empty buffer is still the only case that triggers the bug, but if for instance `"zdi"` were used somewhere, that would also fail to set the `z` register if used on `""`.

`mini.move` also doesn't properly restore the `"` and `z` registers:
- The `0` register is overwritten when restoring the `"` register.
- All register restorations convert the register to either charwise or linewise. Any `NUL` bytes stored in the register are converted to newlines when restoring, which isn't great for all those people who like to put `NUL`s in their code.

These are due to two oversights:
- The `"` register does not actually hold any text, it's only a pointer to another register. `setreg('"', 'foo')` actually sets the `0` register to `foo` and points the `"` register at it.
- `getreg()` only returns the contents of a register, not its type. Setting a register without specifying its type sets it to linewise if the text ends in a newline, and charwise if not. `getreg()` returns a string by default, with `NUL` bytes replaced with newlines due to a quirk of Vim's internal representation of text.

This PR avoids the "nothing in register" error by setting the `z` register to an empty string as a safe default before attempting to cut into it, and replaces `getreg()` with `getreginfo()`, which returns a Dictionary (which behaves as a table in Lua) which holds the entire state of a register and can be passed back to `setreg()` to restore that state.

For saving and restoring the `"` register, only the `points_to` entry is saved; calling `setreg('"')` with a dict/table containing only `points_to` sets the pointee of the `"` register without changing its contents. Calling `setreg('"')` with the saved output of `getreginfo()` without omitting `regcontents` will actually overwrite the contents of the register named by `points_to`, which is a bit of a footgun.

I also added some regression tests which catch the issues I mentioned, except for the clobbering of the `0` register, which could be added to the "has no side effects" tests, I guess.

### DIFF
--- a/lua/mini/move.lua
+++ b/lua/mini/move.lua
@@ -221,8 +221,10 @@ MiniMove.move_selection = function(direction, opts)
   local cache_virtualedit = vim.o.virtualedit
   if not cache_virtualedit:find('all') then vim.o.virtualedit = 'onemore' end
 
-  -- Cut selection while saving caching register
-  local cache_z_reg = vim.fn.getreg('z')
+  -- Save original state of z register
+  local cache_z_reg = vim.fn.getreginfo('z')
+
+  -- Cut into z register
   cmd('"zx')
 
   -- Detect edge selection: last line(s) for vertical and last character(s)
@@ -340,8 +342,10 @@ MiniMove.move_line = function(direction, opts)
     return
   end
 
-  -- Cut curre lint while saving caching register
-  local cache_z_reg = vim.fn.getreg('z')
+  -- Save original state of z register
+  local cache_z_reg = vim.fn.getreginfo('z')
+
+  -- Cut current line into z register
   cmd('"zdd')
 
   -- Move cursor
@@ -457,7 +461,7 @@ H.make_cmd_normal = function(include_undojoin)
 
     -- Disable 'mini.bracketed' to avoid unwanted entries to its yank history
     local cache_minibracketed_disable = vim.b.minibracketed_disable
-    local cache_unnamed_register = vim.fn.getreg('"')
+    local cache_unnamed_register = { points_to = vim.fn.getreginfo('"').points_to }
 
     -- Don't track possible put commands into yank history
     vim.b.minibracketed_disable = true
--- a/lua/mini/move.lua
+++ b/lua/mini/move.lua
@@ -224,6 +224,10 @@ MiniMove.move_selection = function(direction, opts)
   -- Save original state of z register
   local cache_z_reg = vim.fn.getreginfo('z')
 
+  -- If buffer has no lines, deletion is a no-op and won't set the register, so
+  -- set it to a safe default.
+  vim.fn.setreg('z', '')
+
   -- Cut into z register
   cmd('"zx')
 
@@ -344,6 +348,10 @@ MiniMove.move_line = function(direction, opts)
 
   -- Save original state of z register
   local cache_z_reg = vim.fn.getreginfo('z')
+
+  -- If buffer has no lines, deletion is a no-op and won't set the register, so
+  -- set it to a safe default.
+  vim.fn.setreg('z', '')
 
   -- Cut current line into z register
   cmd('"zdd')
--- a/tests/test_move.lua
+++ b/tests/test_move.lua
@@ -109,6 +109,26 @@ T['move_selection()'] = new_set()
 
 local move = function(direction, opts) child.lua('MiniMove.move_selection(...)', { direction, opts }) end
 
+T['move_selection()']['is no-op on empty buffer'] = function()
+  for _ = 1, 2 do
+    -- First, try with registers unset
+    for _, visual_mode in ipairs({ 'v', 'V' }) do
+      type_keys(visual_mode)
+      for _, direction in ipairs({ 'down', 'up', 'left', 'right' }) do
+        move(direction)
+        validate_state({ '' }, { { 1, 1 }, { 1, 1 } })
+      end
+    end
+
+    -- Try again with " and z registers set
+    child.fn.setreg('"', 'a')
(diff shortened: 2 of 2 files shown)


echasnovski:
Thanks for the PR and extensive description!

For the empty buffer case I *think* I'd rather add explicit checks:
- Do nothing for linewise movement if there is only single line.
- Do nothing for charwise and blockwise movement if there is only single empty line.

Using `getreginfo()` instead of `getreg()` sounds like a good idea, but it might have been intentional. I need to investigate further.

> For saving and restoring the `"` register, only the `points_to` entry is saved; calling `setreg('"')` with a dict/table containing only `points_to` sets the pointee of the `"` register without changing its contents. Calling `setreg('"')` with the saved output of `getreginfo()` without omitting `regcontents` will actually overwrite the contents of the register named by `points_to`, which is a bit of a footgun.

I'd also need to take a closer look at this.

echasnovski:
Thanks again for the PR!

I ended up incorporating a universal `getreginfo()` approach (setting only `points_to` for unnamed register because it broke updated tests, while `getreginfo()` worked) and having an explicit early return instead of setting `z` register. Plus added tests (although I could not write ones that failed for "z" register with the previous approach).

This should now be part of the `main` branch.

Diomendius:
I assume you mean this failed test:
```
FAIL in tests/test_move.lua | move_line() | has no side effects:
  Failed expectation for equality.
  Left:  {
    points_to = "1",
    regcontents = { "aaa" },
    regtype = "V"
  }
  Right: {
    points_to = "1",
    regcontents = { "wor" },
    regtype = "V"
  }
  Traceback:
    tests/test_move.lua:1139
```
This is actually catching a different bug not caused by setting only `points_to`.

https://github.com/echasnovski/mini.nvim/blob/cf81e70e8f1fdc36a3c684efc45c4ee561285137/lua/mini/move.lua#L352 Here, `"zdd` actually sets multiple registers:
- `z` is set to the text of the deleted line as intended
- The numbered registers are shifted up and `1` is also set to the line's text, as if `dd` was executed with no specified register
- `"` is pointed to the `1` register

I believe this is actually a bug in Nvim, as `:help quote1` clearly states:
> Numbered register 1 contains the text deleted by the most recent delete or change command, **unless the command specified another register** or the text is less than one line (the small delete register is used then).

This also happens with `"zx`: https://github.com/echasnovski/mini.nvim/blob/cf81e70e8f1fdc36a3c684efc45c4ee561285137/lua/mini/move.lua#L232

A workaround is to yank into `z`, then delete into the black hole register:
```patch
--- a/lua/mini/move.lua
+++ b/lua/mini/move.lua
@@ -232 +232 @@ MiniMove.move_selection = function(direction, opts)
-  cmd('"zx')
+  cmd('"zygv"_x')
@@ -352 +352 @@ MiniMove.move_line = function(direction, opts)
-  cmd('"zdd')
+  cmd('"zyy"_dd')
```

The footgun I mentioned about restoring `"` with a table including more than just `points_to` still applies. This causes problems when `"` points to `z`, because `z` is overwritten with its old value before `cmd()` returns:
```vim
execute "norm ifoo\<cr>bar"
1yank z
lua require("mini.move").move_line("up")
```
Expected buffer contents:
```
bar
foo
```
Actual buffer contents:
```
foo
foo
```

Diomendius:
More specifically it seems to be how Vi originally behaved, but Vim's documentation is self-contradictory. See vim/vim#2346 and vim/vim@9d7fdd403a3a9ee0d008b6dcbcd2ecc9ec0f57b7. The workaround is likely permanently necessary, though Vim's and Neovim's docs should be updated for clarity.

echasnovski:
> ```viml
> execute "norm ifoo\<cr>bar"
> 1yank z
> lua require("mini.move").move_line("up")
> ```

Yeah, that's not ideal. Thanks for the actual (and small) examples! This is very helpful.
I'll look into this.

echasnovski:
Both of these should now be fixed on latest `main`. And indeed there is now test for using only `points_to` for unnamed register (had to point it to "z" register). Thanks for the example!

> I believe this is actually a bug in Nvim, as `:help quote1` clearly states:
> 
> > Numbered register 1 contains the text deleted by the most recent delete or change command, **unless the command specified another register** or the text is less than one line (the small delete register is used then).

The help [was updated](https://github.com/neovim/neovim/blob/40bc48479dca0d4a02c6445ceccab700847dcbe9/runtime/doc/change.txt#L1230) on Nightly [several months ago](https://github.com/neovim/neovim/commit/1922f7e32b1c61e892bf4bedaf938e8b7aa44c43) following the change in Vim. The reasoning is apparently is [to be compatible with POSIX `vi`](https://github.com/vim/vim/issues/15077#issuecomment-2185013629).

Diomendius:
Ah, I didn't see that that had already been committed. I only checked the help in the `v0.10.1` release.

Looks like everything works well now. Thanks!

PR #1148, opened 2024-08-16
fix(mini.files): correct optional type hint for go_in

aarnphm:
Since we do check for opts below, the hint here should be optional

as `go_in()` or `go_in {close_on_file = true}` will work here


echasnovski:
Thanks again for spotting this! It should be resolved on latest `main`.

aarnphm:
nice thanks a bunch

Issue #1111, opened 2024-08-03
(misc): setup_termbg_sync reset can spuriously fail


wroyca:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

misc

### Description

*This is a tracking issue for https://github.com/echasnovski/mini.nvim/issues/999. Currently, we don’t have a reproducible case for this issue, and it is highly timing-sensitive, so its behavior may vary depending on various factors on user machine.*

> [!NOTE]
> From the information we've received thus far, it doesn't seem to affect macOS users.

From time to time, resetting the terminal background might fail:

https://github.com/user-attachments/assets/f87fe333-3422-44a5-8c51-05da3b96116d

I confirmed (in my case*) that I could reproduce with:

* Fish
* Zsh
* Kitty
* St
* Bash

#### Testimony

From `@echasnovski`:

> It still works in all terminal emulators I've tried. The difference is that I have zsh

From `@gpanders`:

> I did ask @gpanders earlier to reproduce with fish. Everything works as expected in Ghostty and iTerm2 terminal emulators (on MacOS, I believe).

From `@pkazmier`: https://github.com/echasnovski/mini.nvim/issues/1111#issuecomment-2267556583

#### Workaround?

There are currently two workarounds that I know of which work in my case:

1. **Replacing `{ "UILeave" }` with `{ "VimLeavePre", "VimSuspend" }`**

   - This is a solution I've been using personally for months before the introduction of `setup_termbg_sync`.

2. **Wrapping `io.write` in a `defer_fn`**

   - Although I haven’t tested this extensively, it appears to work with the following code:
     ```lua
     vim.defer_fn(function() 
       io.write('\027]11;' .. H.termbg_init .. '\007')
     end, 10)
     ```

### Neovim version

0.10.0

### Steps to reproduce

1. TBD

echasnovski:
Thanks for the screencast!

Here is what I see from it:
- It requires about 15 **extremely** quick transitions from background to foreground to reproduce this at least once on your setup.
- Even after that, one more `fg` and `<C-z>` can restore to the correct state.

[...]

PR #1202, opened 2024-09-07
fix(misc): target stdout stream explicitly

wroyca:
Use explicit targeting to sync the terminal background. 

Nvim changes standard streams to non-blocking mode, so we should use io.stdout:write to ensure that writes are handled properly (that is, directly addressing the term's standard output).

~~This should address the last remaining bits from #1111:~~

> ~~That's both good (it fixes something) and confusing (it introduces very similar issue for me) news. I'll just focus on the first interpretation :)~~

Let me know how it goes :)

### DIFF
--- a/lua/mini/misc.lua
+++ b/lua/mini/misc.lua
@@ -302,7 +302,7 @@ H.root_cache = {}
 --- Works only on Neovim>=0.10.
 MiniMisc.setup_termbg_sync = function()
   if vim.fn.has('nvim-0.10') == 0 then
-    -- Handling `io.write('\027]11;?\007')` response was added in Neovim 0.10
+    -- Handling `io.stdout:write('\027]11;?\007')` response was added in Neovim 0.10
     H.notify('`setup_termbg_sync()` requires Neovim>=0.10', 'WARN')
     return
   end
@@ -326,13 +326,13 @@ MiniMisc.setup_termbg_sync = function()
     local sync = function()
       local normal = vim.api.nvim_get_hl_by_name('Normal', true)
       if normal.background == nil then return end
-      io.write(string.format('\027]11;#%06x\007', normal.background))
+      io.stdout:write(string.format('\027]11;#%06x\007', normal.background))
     end
     vim.api.nvim_create_autocmd({ 'VimResume', 'ColorScheme' }, { group = augroup, callback = sync })
 
     -- Set up reset to the color returned from the very first call
     H.termbg_init = H.termbg_init or bg_init
-    local reset = function() io.write('\027]11;' .. H.termbg_init .. '\007') end
+    local reset = function() io.stdout:write('\027]11;' .. H.termbg_init .. '\007') end
     vim.api.nvim_create_autocmd({ 'VimLeavePre', 'VimSuspend' }, { group = augroup, callback = reset })
 
     -- Sync immediately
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

> Nvim changes standard streams to non-blocking mode, so we should use io.stdout:write to ensure that writes are handled properly (that is, directly addressing the term's standard output).

Could you, please, share a link to where this information coming from?

> This should address the last remaining bits from #1111:
> 
> > That's both good (it fixes something) and confusing (it introduces very similar issue for me) news. I'll just focus on the first interpretation :)
> 
> Let me know how it goes :)

Did you manage to reproduce similar issue as I described there? And this change should fix it, right?

echasnovski:
I've just tried reproducing "<C-z> -> `fg` -> `<C-z> -> ... -> <C-z> (background is the same as in Neovim)"  with and without the change in the PR. Both times I was able to reproduce. So I don't this fixes the issue.

wroyca:
> Thanks for the PR!
> 
> > Nvim changes standard streams to non-blocking mode, so we should use io.stdout:write to ensure that writes are handled properly (that is, directly addressing the term's standard output).
> 
> Could you, please, share a link to where this information coming from?
> 
> > This should address the last remaining bits from #1111:
> > > That's both good (it fixes something) and confusing (it introduces very similar issue for me) news. I'll just focus on the first interpretation :)
> > 
> > 
> > Let me know how it goes :)
> 
> Did you manage to reproduce similar issue as I described there? And this change should fix it, right?

@echasnovski Sorry, didn't see that comment yesterday - oops.

> Could you, please, share a link to where this information coming from?

This issue stems from a problem I encountered with build2, which only occurred when Neovim was started in the background. The build2 team investigated and discovered that non-blocking streams were the cause. For background, see https://github.com/build2/build2/issues/417

> Did you manage to reproduce similar issue as I described there? And this change should fix it, right?

No — I’m unable to reproduce the sync issue at all, so it’s surprising that you can reproduce it on your machine. Why is that? What’s the difference? Who knows, haha.

That said, there is a subtle difference between io.write and io.stdout:write when output redirection is involved. io.write writes to the current default output stream, which can be redirected. If the default output stream changes, io.write will respect that redirection. io.stdout:write, on the other hand, always writes to stdout, regardless of any redirection. It specifically targets the stdout stream, which always goes to the terminal (or the original standard output).

I think this change is worthwhile if only for that, even though I had hoped it would fix the sync issue on your side, but alas. Let me know if you’d like to reopen this, in any case.

echasnovski:
> Adding to this, here’s an example to illustrate the difference:
> 
> ```lua
> local file = io.open("output.txt", "w")
> io.output(file)
> file:close()
> ```
> 
> In this case, file:close will cause mini.misc (io.write) to error out with:
> 
> ```
> Error executing lua callback: .../wroy/.local/share/nvim/lazy/mini.nvim/lua/mini/misc.lua:329: standard file is closed  
> ```
> 
> But io.stdout:write will still work as expected, since it directly writes to stdout, "bypassing" the redirection and file closure.

Oh, my... You should have started with that. I can reproduce. Will schedule to finish and merge this PR tomorrow.

echasnovski:
This now should be part of `main` branch. Thanks for finding this!

PR #1205, opened 2024-09-08
docs(files): add example for searching inside a directory

niknbr:

Thank you for this plugin, `mini.files` is the most ergonomic file explorer I have used. If you are interested, I'd like to contribute this example which I use a lot. I tried to do this using `mini.pick` but couldn't find a way to set search on selected directory using `MiniPick.builtin.grep_live()`.

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -535,6 +535,31 @@ Use |MiniFiles.set_bookmark()| inside `MiniFilesExplorerOpen` event: >lua
     end,
   })
 <
+# Search inside a directory ~
+
+Search inside selected directory using Telescope: >lua
+
+  vim.api.nvim_create_autocmd('User', {
+    pattern = 'MiniFilesWindowUpdate',
+    callback = function(args)
+      local buf_id = args.data.buf_id
+
+      local search_in_directory = function()
+        local fs_entry = MiniFiles.get_fs_entry()
+        if fs_entry.fs_type == 'directory' then
+          MiniFiles.close()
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

I think this type of example config is a bit too much to be part of `:help`. Especially since it mentions something outside of 'mini.nvim' :) Maybe for a future wiki page.

Closing as not planned in this form.

------

For reference, here is a little bit cleaner and more robust version:

```lua
local search_in_dir = function()
  local fs_entry = MiniFiles.get_fs_entry()
  if fs_entry.fs_type ~= 'directory' then return vim.notify('Not a directory', 'WARN') end
  MiniFiles.close()
  vim.schedule(function()
    local command = 'Telescope live_grep search_dirs=' .. vim.fn.fnameescape(fs_entry.path)
    vim.cmd(command)
  end)
end

local make_buf_keymaps = function(args)
  local opts = { buffer = args.data.buf_id, desc = 'Search in directory under cursor' }
  vim.keymap.set('n', '<C-f>', search_in_dir, opts)
end

vim.api.nvim_create_autocmd('User', { pattern = 'MiniFilesBufferCreate', callback = make_buf_keymaps })
```

Note, that creating buffer-local mappings is much better to do [inside `MiniFilesBufferCreate`](https://github.com/echasnovski/mini.nvim/blob/8413efde2546be4dec9ea424bc6af346a14dffbe/doc/mini-files.txt#L336) event and not `MiniFilesWindowUpdate` (which gets triggered *way* more frequently than `MiniFilesBufferCreate`).

PR #1214, opened 2024-09-12
fix(align): blockwise align with wide and combining characters

Diomendius:
Blockwise align is currently broken when lines contain multi-cell characters such as:
- Tab
- CJK characters
- Control codes (Vim represents these as `^*`)
- Some invisible Unicode code points (Vim represents these as `<xxxx>`)

Single display cells containing multiple Unicode code points also cause problems, such as when `é` is decomposed to `e<U+0301 COMBINING ACUTE ACCENT>`.

This is mainly caused by passing virtual column positions to `strcharpart()` and `byteidx()` as if they were character indices.

When the leftmost selection corner sits directly on a wide character, only the last virtual column the character occupies is included in the selection, as `virtcol()` is called without the `list` argument; without this argument, `virtcol()` only returns the position of the rightmost column containing the character.

Exclusive block selections also aren't handled correctly, as the right vcol is always shifted left. An exclusive block selection extends from and including the corner with the lowest byte offset, to the left edge of the corner with the highest. This means that the selected region is the same as for an inclusive selection unless the corner with the highest byte offset is at least one virtual column to the right of the other corner.

This PR fixes all these issues and adds regression tests. I've added `T['Align']['all modes']` and `T['Align']['blockwise mode']` sets, which many of the existing tests could be adapted to use. This should be applicable to any tests which operate over the entire text, as all modes should behave identically in this case.

One issue I encountered which this PR does not fix is that tabs not at the start of the line are aligned incorrectly. I don't think this is worth fixing as using tabs in the middle of lines is a terrible idea in general, and it's not obvious what the correct behaviour should even be.


echasnovski:
Thanks for the PR!

Code change at glance looks a bit complicated, maybe even to the point of opting for not supporting these cases. And definitely uses functionality present only in Neovim>=0.10.

Currently there is a "code freeze" period before the next 0.14.0 release, so this has to wait a bit. After that I'll have a closer look.

Diomendius:
It looks like the `list` arg to `virtcol()` requires Nvim 0.10. It should be possible to implement a workaround for older versions, probably by grabbing text from the buffer and using `strdisplaywidth()`. I'll implement a fix for older versions soon.

To my eye, the new implementation looks about as complex as the old implementation. The current handling of virtual columns causes text to be deleted or duplicated, and causes a `'start' is higher than 'end'` error in `nvim_buf_set_text()` when the selected portion of a line starts with a multi-cell character and ends on or beyond the last character in the line, so keeping this behaviour is not desirable.

You can run the tests at c2078cb946a81b284ab7a9630a86670a13c2c852 to see the problems with the current implementation.

<details>
<summary>Notable test failures @ c2078cb946a81b284ab7a9630a86670a13c2c852</summary>

These also fail in the same way for `Normal-block` mode.
```
FAIL in tests/test_align.lua | Align | all modes | works with wide characters + args { "Visual-block" }:
Failed expectation for equality.
Left:  { "ыыф  фццф", "ы фのфц фыфのфцф" }
Right: { "ыыф  фццф", "ы фのфц ф" }
Traceback:
    tests/test_align.lua:1163
    tests/test_align.lua:1314
```
```
FAIL in tests/test_align.lua | Align | blockwise mode | correctly handles virtual column bounds + args { "Visual-block" }: tests/test_align.lua:1162: E5108: Error executing lua: vim/_editor.lua:0: nvim_exec2(): Vim(lua):E5108: Error executing lua [string ":lua"]:1: 'start' is higher than 'end'
stack traceback:
    [C]: in function 'nvim_buf_set_text'
    [C]: in function 'nvim_exec2'
    vim/_editor.lua: in function 'cmd'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1902: in function 'set_text'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1765: in function 'region_set_text'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1676: in function 'process_current_region'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:743: in function 'align_user'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1360: in function <.../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1353>
stack traceback:
    [C]: in function 'nvim_exec2'
    vim/_editor.lua: in function 'cmd'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1902: in function 'set_text'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1765: in function 'region_set_text'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1676: in function 'process_current_region'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:743: in function 'align_user'
    .../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1360: in function <.../.local/share/nvim/lazy/mini.nvim/lua/mini/align.lua:1353>
Traceback:
    tests/test_align.lua:1162
    tests/test_align.lua:1345
```
</details>

echasnovski:
@Diomendius, thanks for the work! I'll take it from here after the next release of 'mini.nvim'.

echasnovski:
@Diomendius, thanks again for this huge amount of meticulous work in the PR!

After reading into this more closely, here is the final result of what I think is reasonable to have:
- The main usage of `virtcol2col()` with adjustment to make 'mini.align' work in presence of wide and composing characters is reasonable after some refactor. The main goal was to have it fixed with as simple code as possible. I refactored it to use the same code snippet for both `region_get_text` and `region_set_text` (in blockwise mode) plus small adjustments depending on Neovim version.

    Tests for these are a bit over the top in this PR, so I decided to keep only necessary minimum.
- Supporting 'ambiwidth=double' will probably be out of scope for the whole 'mini.nvim'. Because having it enabled kind of breaks most of modules altogether (like usage of icons, listchars/fillchars, etc.). Plus it seems to be mostly about visual representation of the text, which I don't think is a good thing to depend on when aligning text.
- Supporting "to EOL" blockwise selection although good, actually doing so seems too hacky and not clean. Plus this type of selection also is not supported in 'mini.surround' and 'mini.operators' (maybe others also) so is better for consistency.

    The suggested way to approach "to EOL" blockwise selection is to have 'virtualedit=block' and selecting manually the exact rectangle. Yes, this is not ideal, but will have to be enough. At least for while until there is no pressing issues with modules other than supporting that across the whole 'mini.nvim'.
- Supporting the case of "all edges with selection=exclusive" also seems to not worth the extra code complexity. The suggested approach is to select across topleft-bottomright diagonal which should work.

In the future I highly suggest first opening an issue to discuss the problem before making the PR. This will most certainly save time for both parties.

Diomendius:
I skimmed your comment and commit when you made them, but didn't reply at the time. Thanks for what you've done here. I don't blame you for omitting support for `ambiwidth` or the edge cases of blockwise exclusive selection, and not just because I don't use them myself.

I happened to revisit your comment today and noticed your comments on workarounds for blockwise to-EOL selection. Using `virtualedit=block` to explicitly select a region past EOL seems to work fine with `mini.align`, but with `mini.surround` it actually doesn't work correctly, while blockwise to-EOL selection *does*, even if you didn't intend it to.

`mini.surround` seems to effectively ignore `virtualedit` by clamping both the start and end columns to the EOL column (one past the last character) of their respective lines. There are also some edge cases around wide characters, but nothing major and nothing as bad as the problems in this issue. I don't know if or when I'll make a detailed issue about this, but I might as well mention this in case someone finds it useful.

I think what this really highlights is the deficiencies in the Nvim API regarding selections. Having a single representation of a visual selection that unambiguously indicates which cells are included, regardless of selection order or `selection=exclusive` or anything else, would be a huge benefit to almost every plugin that needs to handle selections in any way.

Issue #839, opened 2024-04-24
mini.pairs: Input lag for {<CR>} expansion


cthoma:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

mini.pairs

### Description

I very much appreciate the work and effort that is being put into the mini plugin collection, but
I have something to complain, and maybe a solution for it.

pairs.lua:390
```
-- Precomputed keys to increase speed
-- stylua: ignore start
local function escape(s) return vim.api.nvim_replace_termcodes(s, true, true, true) end
H.keys = {
  above     = escape('<C-o>O'), -- <<<< This is the offending line.
  bs        = escape('<bs>'),
  cr        = escape('<cr>'),
  del       = escape('<del>'),
  keep_undo = escape('<C-g>U'),
  -- NOTE: use `get_arrow_key()` instead of `H.keys.left` or `H.keys.right`
  left      = escape('<left>'),
  right     = escape('<right>')
}
-- stylua: ignore end
```

The issue arises from the fact the `<C-o>O` is used to add a blank line between opening and closing curly braces.
This mapping brings us briefly into normal mode, which triggers events we don't need -- that's what causes the lag.
A blank line above is all we want.

I rewrote the offending line to `above = escape('<cr><up><C-f>')`.
This sequence keeps us in insert mode and there is no lag.

`<C-f>` ... reindent current line `:h i_CTRL-F`

### Neovim version

-

### Steps to reproduce

-

### Actual behavior

-

echasnovski:
Thanks for the issue!

This was already brought up and discussed in #626 with a detailed resolution in [this comment](https://github.com/echasnovski/mini.nvim/issues/626#issuecomment-1868256077).

However, I think I like the idea of using `<C-f>` for re-indent, but not with `<Up>` as it might be disabled by some users and seems to break some later remapping (like to work well with 'mini.completion'). Maybe `<C-g>k` could work, though.

That said, direct keys replace breaks tests and I did not manage to quickly figure out why exactly. Will take a look later.

echasnovski:
This should now be fixed on latest `main`.

[...]

Issue #196, opened 2023-01-13
'mini.pairs' lag on large files


barrett-ruth:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

mini.pairs

### Description

Slightly laggy pairing on large files

### Neovim version

NVIM v0.9.0-dev-626+g42afa0369

### Steps to reproduce

minimal.lua:
```lua
require('mini.pairs').setup {}
```

1. `nvim -nu minimal.lua`
2. Open up a large file, like [this one](https://git.suckless.org/dwm/file/dwm.c.html)
3. Go to a random line
4. Begin typing pairs

### Expected behavior

Instantaneous pairing

### Actual behavior

First off, I love your plugins. Your work is great.

Now I hate to make comparisons, but in my experience [nvim-autopairs](https://github.com/windwp/nvim-autopairs) just has mini.pairs beat in this regard. While I really want to use mini.pairs as I adore the minimal, non-bloated design of your plugins, there is a *slight* lag on large files that I just can't ignore. 

On small files mini.pairs works instantaneously, but what I need from a pairing plugin is for it to do that all the time, no matter what.

Then again, it is definitely possible I'm missing something. Let me know if you can replicate the issue.

echasnovski:
I *think* I can't reproduce. This is how I have it on your large file (and even larger):

https://user-images.githubusercontent.com/24854248/212401246-ea6f0661-1c22-4ac3-95be-294591606388.mp4

Do you experience lag on all or only some pairs? The usual reason for any kind of lag when dealing with mappings is that there is some other one which "extends" the trouble one. In that case (Neo)vim should wait for some time (`:h 'timeoutlen'`) because it is unsure if user is going to add another keys.

Try executing `:imap (` (and other troubled keys) and seeing if there is anything else.

barrett-ruth:
It is not the timeoutlen issue. I will try to debug literally every plugin I have in my config (a fun process).

Thanks for the quick response.

echasnovski:
> It is not the timeoutlen issue. I will try to debug literally every plugin I have in my config (a fun process).
> 
> Thanks for the quick response.

[...]

PR #1229, opened 2024-09-18
fix(pairs): consider using `lazyredraw` to prevent cursor flickering

39555:
Temporarily toggle `vim.o.lazyredraw = true` while inserting pairs to prevent weird cursor flickering

### Before this pr

I reduced the video speed to 1/4. The cursor jumping forth and back drives me nuts))

https://github.com/user-attachments/assets/2ef07b04-cea1-4c51-8535-8502a65ba859

### After this pr

The screen will be redrawn only on the next tick. Similar to your using of  `vim.o.eventignore`

https://github.com/user-attachments/assets/8a902893-6886-42f3-9dd8-345c66eb2a21

### DIFF
--- a/lua/mini/pairs.lua
+++ b/lua/mini/pairs.lua
@@ -281,6 +281,7 @@ end
 MiniPairs.open = function(pair, neigh_pattern)
   if H.is_disabled() or not H.neigh_match(neigh_pattern) then return pair:sub(1, 1) end
 
+  H.with_lazyredraw()
   return ('%s%s'):format(pair, H.get_arrow_key('left'))
 end
 
@@ -303,6 +304,7 @@ MiniPairs.close = function(pair, neigh_pattern)
 
   local close = pair:sub(2, 2)
   if H.get_cursor_neigh(1, 1) == close then
+    H.with_lazyredraw()
     return H.get_arrow_key('right')
   else
     return close
@@ -326,6 +328,7 @@ MiniPairs.closeopen = function(pair, neigh_pattern)
   if H.is_disabled() or H.get_cursor_neigh(1, 1) ~= pair:sub(2, 2) then
     return MiniPairs.open(pair, neigh_pattern)
   else
+    H.with_lazyredraw()
     return H.get_arrow_key('right')
   end
 end
@@ -362,6 +365,7 @@ MiniPairs.bs = function(key)
     res = ('%s%s'):format(res, H.keys.del)
   end
 
+  H.with_lazyredraw()
   return res
 end
 
@@ -391,6 +395,8 @@ MiniPairs.cr = function(key)
     local cache_eventignore = vim.o.eventignore
     vim.o.eventignore = 'InsertLeave,InsertLeavePre,InsertEnter,ModeChanged'
     H.restore_eventignore(cache_eventignore)
+
+    H.with_lazyredraw()
 
     res = ('%s%s'):format(res, H.keys.above)
   end
@@ -641,5 +647,12 @@ H.map = function(mode, lhs, rhs, opts)
 end
 
 H.restore_eventignore = vim.schedule_wrap(function(val) vim.o.eventignore = val end)
+
+-- Redraw the screen only after the action is complete to prevent cursor flickering
+H.with_lazyredraw = function()
+  local cache_lazyredraw = vim.o.lazyredraw
+  vim.o.lazyredraw = true
+  vim.schedule_wrap(function() vim.o.lazyredraw = cache_lazyredraw end)
+end
 
 return MiniPairs
--- a/lua/mini/pairs.lua
+++ b/lua/mini/pairs.lua
@@ -281,7 +281,6 @@ end
 MiniPairs.open = function(pair, neigh_pattern)
   if H.is_disabled() or not H.neigh_match(neigh_pattern) then return pair:sub(1, 1) end
 
-  H.with_lazyredraw()
   return ('%s%s'):format(pair, H.get_arrow_key('left'))
 end
 
@@ -304,7 +303,6 @@ MiniPairs.close = function(pair, neigh_pattern)
 
   local close = pair:sub(2, 2)
   if H.get_cursor_neigh(1, 1) == close then
-    H.with_lazyredraw()
     return H.get_arrow_key('right')
   else
     return close
@@ -328,7 +326,6 @@ MiniPairs.closeopen = function(pair, neigh_pattern)
   if H.is_disabled() or H.get_cursor_neigh(1, 1) ~= pair:sub(2, 2) then
     return MiniPairs.open(pair, neigh_pattern)
   else
-    H.with_lazyredraw()
     return H.get_arrow_key('right')
   end
 end
@@ -365,7 +362,6 @@ MiniPairs.bs = function(key)
     res = ('%s%s'):format(res, H.keys.del)
   end
 
-  H.with_lazyredraw()
   return res
 end
 
@@ -395,8 +391,6 @@ MiniPairs.cr = function(key)
     local cache_eventignore = vim.o.eventignore
     vim.o.eventignore = 'InsertLeave,InsertLeavePre,InsertEnter,ModeChanged'
     H.restore_eventignore(cache_eventignore)
-
-    H.with_lazyredraw()
 
     res = ('%s%s'):format(res, H.keys.above)
   end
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Having the `lazyredraw` temporarily set on almost every 'mini.pairs' related occasions is a bit too much. The `eventignore` workaround was necessary to not trigger `InsertLeave` event which can result in big lags (see #839) and it is only triggered when pressing `<CR>` inside registered pair, which is not that much.

And I also can not reproduce this cursor jumping. Can you reproduce this with only 'mini.pairs' enabled? Because I suspect the real reason for the lag is the rainbow coloring of brackets. For example, see #196.

39555:
I can reproduce with tree-sitter enabled `:=vim.treesitter.start()` with just mini.pairs, with your dotfiles and with LazyVim
1. `git clone --filter=blob:none https://github.com/echasnovski/nvim.git ~/.config/nvim-mini`
2. `NVIM_APPNAME=nvim-mini nvim init.lua`
3. `:=vim.treesitter.start()`

```
NVIM v0.11.0-nightly+bcae8be
Build type: Release
LuaJIT 2.1.1713773202
```


https://github.com/user-attachments/assets/609f7b78-5797-45bc-a1e2-e820ad866b74

Lets wait maybe someone also has such an issue. 

> is a bit too much

Is switching options or `schedule_wrap` expensive here? TIL

echasnovski:
> I can reproduce with tree-sitter enabled `:=vim.treesitter.start()` with just mini.pairs

Yes, I can *kind of* reproduce this with tree-sitter enabled, but it depends on the size of a file. At the end of large file (like 'lua/mini/pick.lua') it is indeed fairly consistent, but in smaller file - not so much.

> Is switching options or `schedule_wrap` expensive here? TIL

Well, for starters, `schedule_wrap()` doesn't do what you assume it would do :) It creates a wrapper function that executes its input inside `vim.schedule()`. See `vim.o.eventignore` approach.

But the option switch is kind of costly. I'd need to benchmark to decided based on numbers.

For the 'eventignore' case it was around 0.002 ms when triggered inside registered pair, which is ok. Here it will be invoked fairly more frequently, so need to think if removing this (not entirely consistent) flicker is worth it.

------

Are **all** calls of `H.with_lazyredraw()` needed? For example, I don't think they are needed before moving cursor to the right. The same goes for `<BS>`.

I can not reproduce the issue with the following diff:

```diff
diff --git a/lua/mini/pairs.lua b/lua/mini/pairs.lua
index 5bea3c9..45c196a 100644
--- a/lua/mini/pairs.lua
+++ b/lua/mini/pairs.lua
@@ -281,6 +281,10 @@ end
 MiniPairs.open = function(pair, neigh_pattern)
   if H.is_disabled() or not H.neigh_match(neigh_pattern) then return pair:sub(1, 1) end
 
+  local cache_lazyredraw = vim.go.lazyredraw
+  vim.go.lazyredraw = true
+  H.restore_lazyredraw(cache_lazyredraw)
+
   return ('%s%s'):format(pair, H.get_arrow_key('left'))
 end
 
@@ -641,5 +645,6 @@ H.map = function(mode, lhs, rhs, opts)
 end
 
 H.restore_eventignore = vim.schedule_wrap(function(val) vim.o.eventignore = val end)
+H.restore_lazyredraw = vim.schedule_wrap(function(val) vim.go.lazyredraw = val end)
 
 return MiniPairs
```

Could you, please, check if there are any visible lags left after applying this change?

39555:
Yeah, using it only inside `MiniPairs.open` is sufficient

echasnovski:
Thanks for changes!

I'll ponder a bit on this at least until the next 0.14.0 release.

39555:
Evgeni, can you tell me what the benchmark you used looked like? I came across different variations of this but failed to get reliable results.

```bash
#!/usr/bin/env bash

function run() {
    nvim --clean -c 'lua _G.package.path = [[./lua/?.lua;]] .. _G.package.path' -c 'lua require("mini.pairs").setup({})'  -u './benchmark.lua' -c 'e ./lua/mini/pick.lua' -c 'lua vim.treesitter.start()'
}

git checkout pairs-lazyredraw
run
git checkout main
run


```

```lua
-- ./benchmark.lua
local key = vim.api.nvim_replace_termcodes('(', true, true, false)
local runs = 1000000

local function benchmark(callback)
  local start_time = vim.uv.hrtime()
  for _ = 0, runs, 1 do
    callback()
  end
  return (vim.uv.hrtime() - start_time) / 1e6 -- microseconds
end

run = function()
    vim.cmd.startinsert()

    local r = benchmark((function() vim.api.nvim_feedkeys(key, 'i', false) end)
    vim.notify(('%.9f'):format(r / runs), vim.log.levels.INFO)
    vim.cmd([[qa!]])
end

vim.schedule(run, 100)

```

echasnovski:
> Evgeni, can you tell me what the benchmark you used looked like? I came across different variations of this but failed to get reliable results.

Not so much elaborated ones :)

Basically:
- Add `_G.log = {}` somewhere inside 'lua/mini/pairs.lua'.
- Add `local start_time = vim.loop.hrtime()` before `local cache_lazyredraw = vim.go.lazyredraw`.
- Add `table.insert(_G.log, 0.000001 * (vim.loop.hrtime() - start_time))` after `H.restore_lazyredraw(cache_lazyredraw)`.
- Restart Neovim and try inserting opening characters. Then look at `_G.log` times (they are in milliseconds).

This approach is more indicative of which delays are present during interactive session.

flukeeey:
> > I can reproduce with tree-sitter enabled `:=vim.treesitter.start()` with just mini.pairs
> 
> Yes, I can _kind of_ reproduce this with tree-sitter enabled, but it depends on the size of a file. At the end of large file (like 'lua/mini/pick.lua') it is indeed fairly consistent, but in smaller file - not so much.
> 
> > Is switching options or `schedule_wrap` expensive here? TIL
> 
> Well, for starters, `schedule_wrap()` doesn't do what you assume it would do :) It creates a wrapper function that executes its input inside `vim.schedule()`. See `vim.o.eventignore` approach.
> 
> But the option switch is kind of costly. I'd need to benchmark to decided based on numbers.
> 
> For the 'eventignore' case it was around 0.002 ms when triggered inside registered pair, which is ok. Here it will be invoked fairly more frequently, so need to think if removing this (not entirely consistent) flicker is worth it.
> 
> Are **all** calls of `H.with_lazyredraw()` needed? For example, I don't think they are needed before moving cursor to the right. The same goes for `<BS>`.
> 
> I can not reproduce the issue with the following diff:
> 
> ```diff
> diff --git a/lua/mini/pairs.lua b/lua/mini/pairs.lua
> index 5bea3c9..45c196a 100644
> --- a/lua/mini/pairs.lua
> +++ b/lua/mini/pairs.lua
> @@ -281,6 +281,10 @@ end
>  MiniPairs.open = function(pair, neigh_pattern)
>    if H.is_disabled() or not H.neigh_match(neigh_pattern) then return pair:sub(1, 1) end
>  
> +  local cache_lazyredraw = vim.go.lazyredraw
> +  vim.go.lazyredraw = true
> +  H.restore_lazyredraw(cache_lazyredraw)
> +
>    return ('%s%s'):format(pair, H.get_arrow_key('left'))
>  end
>  
> @@ -641,5 +645,6 @@ H.map = function(mode, lhs, rhs, opts)
>  end
>  
>  H.restore_eventignore = vim.schedule_wrap(function(val) vim.o.eventignore = val end)
> +H.restore_lazyredraw = vim.schedule_wrap(function(val) vim.go.lazyredraw = val end)
>  
>  return MiniPairs
> ```
> 
> Could you, please, check if there are any visible lags left after applying this change?

Applying this diff fixed some very noticeable and jarring flickering when opening new pairs. :+1:

echasnovski:
I've been daily driving these changes (with also a similar `lazyredraw` switch in `cr()`, as there are flickering there also) for about a week now and didn't notice any obvious performance or usability issues.

The actual benchmarking shows the same level of delay as for 'eventignore', but it is not on 0.002 ms scale as I mentioned earlier: more like 0.04 ms for me. Still relatively fast, although `open()` mappings are used more often than `cr()`.

Let's hope that there will be no performance issues or issues with macros / any other programmatic usage of these keys.

I've merged it into separate branch to polish a bit. It should appear in `main` soon.

Thanks again for finding this interesting workaround for the flickering problem!

PR #1241, opened 2024-09-24
fix(align): prevent undo if no change was made

afonsocarlos:
Prevent undoing previous actions if no change was made in the current alignment iteration. This is very important, for example, when running alignment under a `:normal` command so it can play well with other actions.

### DIFF
--- a/lua/mini/align.lua
+++ b/lua/mini/align.lua
@@ -1679,7 +1679,7 @@ H.process_current_region = function(lines_were_set, mode, opts, steps)
   vim.cmd('redraw')
 
   -- Confirm that lines were actually set
-  return true
+  return table.concat(strings) ~= table.concat(strings_aligned)
 end
 
 H.get_current_region = function()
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

I am not sure that returning "whether lines were set" should be conditioned on whether the lines were the same. Even if lines are the same, the separate undo state is still created. It can be verified by executing `vim.api.nvim_buf_set_lines(0, 0, 1, false, vim.api.nvim_buf_get_lines(0, 0, 1, false))` - this makes buffer change `modified` status which can be reverted by `u`. It does indeed seem to fix the new test while not breaking others, but something is not right here.

The issue with not working on every line with `:%normal` also doesn't seem to be a huge problem. But I'll take a look later.

echasnovski:
Thanks again for the PR!

Sorry for the late processing. Waited to finish 'mini.snippets' to then try and clean backlog from non-pressing issues.

Upon revisiting, this PR seems good. Needs adjusting the `lines_were_set` naming (as it is somewhat confusing), which I'll do myself.

Issue #1207, opened 2024-09-09
icons  get_icons function only return a small set of icons


linsui:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

icons

### Description

https://github.com/echasnovski/mini.nvim/blob/8413efde2546be4dec9ea424bc6af346a14dffbe/lua/mini/icons.lua#L595 only returns a small set of icons (148). Can you also add the filetype category so that it will return a larger set of icons (1050)?

### Neovim version

0.10.0

### Steps to reproduce

```lua
devicons_ok, devicons = pcall(require, "nvim-web-devicons")
icon = devicons.get_icons()["lua"]
```

This doesn't work because lua icon is not in the set.

### Actual behavior

https://github.com/utilyre/barbecue.nvim/blob/main/lua/barbecue/theme.lua#L159 barbecue.nvim can't get the icons.

echasnovski:
This is mocking [how 'nvim-tree/nvim-web-devicons' does this](https://github.com/nvim-tree/nvim-web-devicons/blob/9793801f974bba70e4ac5d7eae6c4f5659993d8e/lua/nvim-web-devicons.lua#L15) (i.e. it [does not include filetypes](https://github.com/nvim-tree/nvim-web-devicons/blob/9793801f974bba70e4ac5d7eae6c4f5659993d8e/lua/nvim-web-devicons.lua#L62-L66)).

If anything, 'utilyre/barbecue.nvim' needs to use `get_icon` here, which is the intended use in this kind of situations. It was changed [here](https://github.com/utilyre/barbecue.nvim/commit/4c43a3cf31ac10c564f5df30385f57e9c81c0465), yet I am still not sure why. Or add direct support for 'mini.icons'.

Closing as working as expected.

PR #1265, opened 2024-10-08
fix(icons): including filetype icons when mocking `nvim-web-devicons`'s `get_icons()` function

xulongwu4:
`nvim-web-devicons`'s `get_icons()` function returns a table of icons which includes the filetype icons, as can be seen here: https://github.com/nvim-tree/nvim-web-devicons/blob/56f17def81478e406e3a8ec4aa727558e79786f3/lua/nvim-web-devicons/icons-default.lua#L994

In the mocking `get_icons()` function, we need to include the `filetype_icons` as well. This fixes an issue when using [barbecue.nvim](https://github.com/utilyre/barbecue.nvim).

Before fix:
Before_fix

After fix:
After_fix

### DIFF
--- a/lua/mini/icons.lua
+++ b/lua/mini/icons.lua
@@ -598,6 +598,7 @@ MiniIcons.mock_nvim_web_devicons = function()
       { [1] = M.get_default_icon() },
       make_category_tbl('os'),
       make_category_tbl('file'),
+      make_category_tbl('filetype'),
       make_category_tbl('extension')
     )
   end


echasnovski:
I am afraid this is not entirely accurate.

'nvim-tree/nvim-web-devicons' does not use filetype icons but only [these sets of icons](https://github.com/nvim-tree/nvim-web-devicons/blob/56f17def81478e406e3a8ec4aa727558e79786f3/lua/nvim-web-devicons.lua#L61-L75). Those sets intersect with 'mini.icons' only via 'filename' and 'extension' categories, which are used in 'mini.icons' mock.

The link you provide clearly states that it is a table based on extension and not filetype (you can read about it in `:h filetype` and `:h 'filetype'`).

Regarding an issue with 'barbecue.nvim', see #1207.

xulongwu4:
> The link you provide clearly states that it is a table based on extension and not filetype (you can read about it in `:h filetype` and `:h 'filetype'`).

Yes, but this table has a key of `lua` [there](https://github.com/nvim-tree/nvim-web-devicons/blob/56f17def81478e406e3a8ec4aa727558e79786f3/lua/nvim-web-devicons/icons-default.lua#L2363), while `mini.icons`'s extension-based table does not have a key for `lua`. Do you mind helping me understand why `mini.icons` does not have such common file extensions included in its `extension_icons` table? @echasnovski

echasnovski:
> Yes, but this table has a key of `lua` [there](https://github.com/nvim-tree/nvim-web-devicons/blob/56f17def81478e406e3a8ec4aa727558e79786f3/lua/nvim-web-devicons/icons-default.lua#L2363), while `mini.icons`'s extension-based table does not have a key for `lua`. Do you mind helping me understand why `mini.icons` does not have such common file extensions included in its `extension_icons` table?

Because it [gets resolved via `vim.filetype.match()`](https://github.com/echasnovski/mini.nvim/blob/a535342b6105a684f16114f598d44318e6a1d20f/doc/mini-icons.txt#L356-L357) removing the need to actually track all possible extensions in 'mini.icons' itself, only all valid filetypes.

To be clear, 'mini.icons' very much does have support for '.lua' extension (this is even explicitly [mentioned in help](https://github.com/echasnovski/mini.nvim/blob/a535342b6105a684f16114f598d44318e6a1d20f/doc/mini-icons.txt#L297-L298)). The problem here is that it is not listed in `MiniIcons.list('extension')`, which is mostly a small price to pay for not having to track all possible extensions.

There are at least two ways to have 'mini.icons' work with 'barbecue.nvim':
- Make PR to 'barbecue.nvim' to stop using `require('nvim-web-devicons').get_icons()` in favor of a single `require('nvim-web-devicons').get_icon()` (which is already optimized).
- Make PR to 'barbecue.nvim' to have direct support for 'mini.icons': check if `_G.MiniIcons` is present and use `_G.MiniIcons.get('file', ...)` to get icon data.
- Add specific extensions in [`config.extensions`](https://github.com/echasnovski/mini.nvim/blob/a535342b6105a684f16114f598d44318e6a1d20f/doc/mini-icons.txt#L195) inside `require('mini.icons').setup()` . Those will be recognized inside `MiniIcons.list()` (and `require('nvim-web-devicons').get_icons()` respectively).

xulongwu4:
@echasnovski Thank you so much for the detailed answer!

PR #1276, opened 2024-10-13
feat(notify) add lsp progress formatter

KaspervdHeijden:
Fixes https://github.com/echasnovski/mini.nvim/issues/1275

### DIFF
--- a/lua/mini/notify.lua
+++ b/lua/mini/notify.lua
@@ -221,6 +221,9 @@ MiniNotify.config = {
     -- Whether to enable showing
     enable = true,
 
+    -- Function to format the progress message
+    format = nil,
+
     -- Duration (in ms) of how long last message should be shown
     duration_last = 1000,
   },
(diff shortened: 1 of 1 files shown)


KaspervdHeijden:
The formatted string will still be post processed by `config.content.format`. An argument could be made that this added layer could instead be added/merged into `config.content.format`. Chose to not do that since it is more work:

- Add `type` to the `MiniNotify-specification` object
- Make `MiniNotify.default_format()` aware
- Update documentation
- Make sure it does not break existing users configurations

echasnovski:
Thanks for the PR!

Adding new option specifically for formatting LSP progress definitely does not fit in 'mini.nvim' design. There is a `config.content.format` and it should be able to be used for this. If no good way to do this, then I am afraid the only way to do this will be manually overriding `vim.lsp.handlers['$/progress']`.

Now, the only missing part for a more straightforward adoption is indeed some type of `source` field for notification. Maybe adding to specification is the right thing to do.

In any case, I'll close this PR and will think a bit more about this. Thanks again for your time!

PR #1288, opened 2024-10-18
feat(deps): support `git@github.com:user/repo` style sources

jshumway:
Exclude `@` characters when detecting `user/repo` style sources while deciding to prepend `https://github.com/`.


echasnovski:
Thanks for finding this and making a PR!

I think it is reasonable to have even more restrictions for which 'user/repo' should be used to infer 'https://github.com/user/repo' source. In particular, only characters which GitHub itself allows to be included in 'user' (alphanumeric and "-") and 'repo' (alphanumeric, "-", "_", ".").

This now should fixed on latest `main` branch.

PR #1327, opened 2024-11-07
feat(mini.files): add a mapping for `go_in_only` without opening file buffer

mau-mauricelim:

@echasnovski I would prefer the default explorer navigation to `go_in` without opening the file buffer unless a different key map is pressed (example `L` or `CR`). Similar to most terminal file manager (example `yazi`).

I am not very good at lua so I am not very sure how to write the tests but I tried to write it to my best ability. I did notice that `make test_files` on `main` branch was already failing. If someone could help update the tests it would be great!


echasnovski:
Thanks for the suggestion!

This is out of scope for 'mini.files' as too specialized and which I don't really support as a workflow. Also this can be done with custom buffer-local mappings (following [this example](https://github.com/echasnovski/mini.nvim/blob/c8bc67a35e627908da644c5f8cd26b3f987f2627/doc/mini-files.txt#L445-L472)):

```lua
local go_in_if_dir = function()
  if (MiniFiles.get_fs_entry() or {}).fs_type ~= 'directory' then return end
  MiniFiles.go_in()
end

  pattern = 'MiniFilesBufferCreate',
  callback = function(args) vim.keymap.set('n', 'l', go_in_if_dir, { buffer = args.data.buf_id }) end,
})

require('mini.files').setup({ mappings = { go_in = '' } })
```

Closing as not planned.

mau-mauricelim:
Amazing! This is exactly what I needed. Thank you for this amazing plugin.

PR #1341, opened 2024-11-16
fix(notify): restrict `schedule_wrap` to `refresh()`

przepompownia:
It allows adding notification and returning its id in fast event.

### DIFF
--- a/lua/mini/notify.lua
+++ b/lua/mini/notify.lua
@@ -298,8 +298,7 @@ MiniNotify.make_notify = function(opts)
     if type(val.hl_group) ~= 'string' then H.error('`hl_group` in level data should be string.') end
   end
 
-  -- Use `vim.schedule_wrap` for output to be usable inside `vim.uv` callbacks
-  local notify = function(msg, level)
+  return function(msg, level)
     level = level or vim.log.levels.INFO
     local level_name = level_names[level]
     if level_name == nil then H.error('Only valid values of `vim.log.levels` are supported.') end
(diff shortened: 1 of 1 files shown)


lua/mini/notify.lua
```diff
@@ -309,10 +308,8 @@ MiniNotify.make_notify = function(opts)
 
     local id = MiniNotify.add(msg, level_name, level_data.hl_group)
     vim.defer_fn(function() MiniNotify.remove(id) end, level_data.duration)
-  end
-  return function(msg, level)
-    if not vim.in_fast_event() then return notify(msg, level) end
-    vim.schedule(function() notify(msg, level) end)
+
+    return id
```
przepompownia, line 312: If compliance with `vim.notify` is important I can remove it - it's not essential and the user can write custom replacement of `make_notify`.

echasnovski:
Thanks for the PR!

This indeed caused problems once or twice before (but allowed to find issues with other modules, but not really important).

I think I'd prefer to have `refresh()` itself schedule its effects depending on whether currently inside fast event. Not sure about the best approach yet, though.

przepompownia:
> Thanks for the PR!
> 
> This indeed caused problems once or twice before (but allowed to find issues with other modules, but not really important).

Probably it doesn't matter much assuming using `make_notify` wrapper. I found the problem when I tried to call `update()` inside `vim.ui_attach` callback (probably especially if `return_prompt` event causes waiting for `<CR>` input). Example (with modified mini.notify): https://github.com/przepompownia/mini-ui-messages-example

> I think I'd prefer to have `refresh()` itself schedule its effects depending on whether currently inside fast event. Not sure about the best approach yet, though.

Do you mean the `if not vim.in_fast_event()` logic could be placed inside `M.refresh()`? I considered it at the start but didn't found any simpler approach than the current.

echasnovski:
> > I think I'd prefer to have `refresh()` itself schedule its effects depending on whether currently inside fast event. Not sure about the best approach yet, though.
> 
> Do you mean the `if not vim.in_fast_event()` logic could be placed inside `M.refresh()`? I considered it at the start but didn't found any simpler approach than the current.

Yeah, so that all current direct `refresh()` usages stay as is. I'll take a look later.

echasnovski:
Calling `MiniNotify.refresh()` directly inside fast events now should be possible on latest `main`.

PR #1392, opened 2024-12-11
docs(sessions): add readme note init global session

mosqueteiro:
Struggled to figure this out so I thought this might help someone else figure it out sooner. If no global session exists and there is no local session, MiniSessions will not automatically create one.

### DIFF
--- a/readmes/mini-sessions.md
+++ b/readmes/mini-sessions.md
@@ -132,6 +132,8 @@ Here are code snippets for some common installation methods (use only one):
 - Enable corresponding git global config value: `git config --system core.longpaths true`. Then try to reinstall.
 - Install plugin in other place with shorter path.
 
+**Note**: a global session needs to exist before it will be able to read or write, unless a local session is detected. Initialize a global session using `:lua MiniSessions.write('session_name')`.
+
 ## Default config
 
 ```lua


echasnovski:
Thanks for the PR!

There was already an attempt to clarify this in 9618cf56e76cc0caab54ad60116a8d72f99d7b3b. In particular, [this lines](https://github.com/echasnovski/mini.nvim/blob/2435d0de00174a45d6b352fefeaa6008ebe9f23b/doc/mini-sessions.txt#L30-L33):

> - Autoread default session (local if detected, latest written otherwise) if Neovim was called without intention to show something else.

This should imply that in order to autoread a session must be written first.

> - Autowrite currently read session before quitting Neovim.

This should imply that no session will be autowritten if it wasn't read (i.e. loaded) before.

Does this make sense?

mosqueteiro:
I had read that when I was trying to figure this out and was not making the connection that I needed to initialize a session first. I'm also pretty new to neovim so it could just be that I'm a newb.

echasnovski:
I went ahead and made [this adjustment](https://github.com/echasnovski/mini.nvim/blob/d084df4d6bf58abcc768b9bb326241ff0c502793/doc/mini-sessions.txt#L24). It is present both in help and README. Hope this makes it more clear to use `write()` in order to "initialize" a session.

mosqueteiro:
That makes sense. Thanks for all your work on these plugins!

PR #1408, opened 2024-12-19
fix(pick): support 'ignorecase' in `rg`'s picker

td0m:

This fixes the `live_grep` behavior when `ignorecase` is enabled. Previously, searching for `todo` in this file would **only** match the lowercase line:
```
// TODO: fix this
// todo: fix this
```

PS: had a couple of tests in `test_pick` fail on me, but this was also the case prior to my PR. Not sure what to do about that.

PS: love your plugins man. <3

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -3290,6 +3290,7 @@ H.grep_get_command = function(tool, pattern, globs)
     local res = {
       'rg', '--column', '--line-number', '--no-heading', '--field-match-separator', '\\x00', '--no-follow', '--color=never'
     }
+    if vim.o.ignorecase then table.insert(res, '--ignore-case') end
     for _, g in ipairs(globs) do
       table.insert(res, '--glob')
       -- NOTE: no `*` as default is important to not "override" ignoring files


echasnovski:
Thanks for the PR!

The current `rg` command line arguments are picked as smallest set possible to have a working grep. Making `live_grep` respect 'ingorecase'/'smartcase' (as other "regular" pickers do) was considered before, but it introduces issues.

The problem with adding more arguments is that they will override user's config for `rg` itself. I can imagine that there might be the case of having different preferences for `live_grep` and Neovim's searching when it comes to ignoring case.

So my suggestion here would be to [create `rg` config](https://github.com/echasnovski/mini.nvim/blob/7ebfab26d77a4b9b05aaae565907e7fa4b2ee154/doc/mini-pick.txt#L80-L85) and set this argument there. This way it will be respected on "per config" basis instead of forcing arguments based on Neovim's options. I personally [use this approach](https://github.com/echasnovski/dotfiles/blob/master/ripgrep/.config/ripgrep/.ripgreprc) (don't forget to `export RIPGREP_CONFIG_PATH="$HOME/.config/ripgrep/.ripgreprc"` with config path of your choice).

PR #1443, opened 2024-12-30
refactor(statusline): switch to feature detection for statusline

aniaan:
Improve mini.statusline's compatibility by switching from version checks to feature detection.

- Replace Neovim version checks with native feature detection
- Simplify LSP client detection logic
- Update diagnostic count and status implementations
- Keep backward compatibility

This makes the code more maintainable and resilient to Neovim API changes.

### DIFF
--- a/lua/mini/statusline.lua
+++ b/lua/mini/statusline.lua
@@ -639,24 +639,28 @@ H.get_attached_lsp = function() return H.attached_lsp[vim.api.nvim_get_current_b
 
 H.compute_attached_lsp = function(buf_id) return string.rep('+', vim.tbl_count(H.get_buf_lsp_clients(buf_id))) end
 
-H.get_buf_lsp_clients = function(buf_id) return vim.lsp.get_clients({ bufnr = buf_id }) end
-if vim.fn.has('nvim-0.10') == 0 then
+if vim.lsp.get_clients then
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

The current approach of checking Neovim version instead of actual functions is intentional. This way `require('mini.statusline')` call doesn't have to load `vim.lsp` and `vim.diagnostic` modules, which visibly improves startup time. As statusline is something that should be not lazy loaded, this is important.

So although version checking is less robust in theory, it is a necessary evil here and in practice shouldn't give much trouble.

aniaan:
I understand, you are right. However, I am currently using the nightly version nvim-0.11. According to the current conditional judgment, it will go to the branch processing for less than 0.10, so I need to wait for the official release of 0.11 before making it compatible, right?

echasnovski:
> However, I am currently using the nightly version nvim-0.11. According to the current conditional judgment, it will go to the branch processing for less than 0.10, so I need to wait for the official release of 0.11 before making it compatible, right?

It won't, it will use the `vim.fn.has('nvim-0.10') == 1` equivalent branch because it is true for all Neovim versions after 0.10.0. You can check that in `:h has()` and manually by executing `:=vim.fn.has('nvim-0.10')` when inside Neovim>0.10.

PR #1454, opened 2025-01-03
[mini.snippets]: Add new function and documentation

krovuxdev:

The function `add_variable_evaluators` is added, allowing users to define custom variable evaluators for snippets. This feature enables greater flexibility in using dynamic values within snippet placeholders.

While exploring, I found [vim-vsnip](https://github.com/hrsh7th/vim-vsnip), which offers a similar feature. This addition felt valuable due to the resemblance and utility, making it a worthwhile feature to include.

### DIFF
--- a/doc/mini-snippets.txt
+++ b/doc/mini-snippets.txt
@@ -1179,5 +1179,20 @@ Return ~
     - <placeholder> `(table|nil)` - array of nodes to be used as placeholder.
     - <transform> `(table|nil)` - array of transformation string parts.
 
+------------------------------------------------------------------------------
+                                        *MiniSnippets.add_variable_evaluators()*
+                  `MiniSnippets.add_variable_evaluators`({opts})
+Add custom variable evaluators
+
+Parameters ~
+{opts} `(table)` Table of custom evaluators. Possible fields:
+  - <VARIABLE> `(function)` - A function that returns the value of the variable.
+    Example:
+      - `VARIABLE` returns the file name.
+      - Custom variables can be defined, such as current date or user-defined values.
+    Default: `{}`.
+
+Return ~
+`(nil)` This function does not return a value. It modifies the global `H.var_evaluators` table.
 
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

This kind of functionality is already possible via [`lookup`](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L1069-L1070) option of `default_insert()`. So users are expected to write custom `config.expand.insert` which pre-evaluates custom variables (as values can only be strings; function evaluators are a bit too much in my opinion but can be added in theory).

What is missing, though, is a clear documentation about that (probably as a part of `MiniSnippets-examples`). Would you mind sharing which variable(s) you'd like to adjust?

krovuxdev:
Thank you very much. I apologize for not knowing that this functionality can be achieved using the `lookup` option in `default_insert()`. I appreciate the clarification.

While I'm not very skilled at writing documentation, I can suggest adding a simple example or a brief description to clarify how the `lookup` option can be used to define variables or functions dynamically. For instance, including a code snippet like `VARIABLE = function() end` could be helpful.

(Additionally, I am still learning English, so I apologize for any mistakes in my writing.)

echasnovski:
> For instance, including a code snippet like `VARIABLE = function() end` could be helpful.

Not quite like that, as lookup values are expected to be strings. Function evaluators are a bit too much in my opinion but can be added in theory.

There is now [an example of how to customize variable evaluation](https://github.com/echasnovski/mini.nvim/blob/546b9c8bd8ebba594bd18811e54bd3e6f9a662ec/doc/mini-snippets.txt#L448-L465).

> (Additionally, I am still learning English, so I apologize for any mistakes in my writing.)

No worries. All non-native English speakers know that feeling. And there are no noticeable issues with you writing, so you're good :)

PR #1483, opened 2025-01-15
fix(starter): empty buffer issue when editing recent file with swapfile

wenjinnn:
When `swapfile` is set, a file has an existing swapfile, and the file is listed in the recent files section, editing this file and selecting `e` or `o` in the popup swapfile menu results in an error and ultimately leaves you with an empty buffer, maybe this problem also appears in other similar case. Fix this by add a `silent! ` at the front of `vim.cmd` parameter in `eval_fun_or_string` function.

Reproduce it in gif:
mini starter-problem

### DIFF
--- a/lua/mini/starter.lua
+++ b/lua/mini/starter.lua
@@ -1526,7 +1526,7 @@ H.eval_fun_or_string = function(x, string_as_cmd)
   if type(x) == 'function' then return x() end
   if type(x) == 'string' then
     if string_as_cmd then
-      vim.cmd(x)
+      vim.cmd(("silent! %s"):format(x))
     else
       return x
     end


echasnovski:
Thanks for the PR!

I don't think using `silent!` for any "string as command" is a good idea as it would hide possibly valuable errors. The same is true, but on a smaller scale for `recent_files` section (as it would probably hide the swap popup entirely, which is not great).

Right now I am having trouble reproducing this, as I can not find a way to reliably reproduce existing swap file (probably a slow morning).
Does presence of 'mini.misc' setup and `setup_restore_cursor()` matter here?

wenjinnn:
Thanks for the quickly response!


> I don't think using `silent!` for any "string as command" is a good idea as it would hide possibly valuable errors. The same is true, but on a smaller scale for `recent_files` section (as it would probably hide the swap popup entirely, which is not great).

Ture, but in my opinion, I find it difficult to retain valuable errors while ensuring the process runs normally. The empty buffer we ended up with was caused by the error throwing, which halted the normal flow, right? (It's not 100% confirmed, just a reasonable guess), So I think we are facing a dilemma: if we suppress the error messages, the process can proceed normally; if we throw the error messages, the process will be interrupted. 
Or maybe just keep the `silent!` command within the code related to `recent_files`.

> Right now I am having trouble reproducing this, as I can not find a way to reliably reproduce existing swap file (probably a slow morning).
> Does presence of 'mini.misc' setup and `setup_restore_cursor()` matter here?

You can reproduce it like this: open a test file in one terminal, insert some characters but don’t write them, and don't exit Neovim. Instead, close the terminal directly. Then open another terminal and open the file again through the starter.
The mini.misc part is not important. it was just test code I wrote when I initially thought the issue was related to mini.misc. Sorry for the confusion caused.

echasnovski:
> You can reproduce it like this: open a test file in one terminal, insert some characters but don’t write them, and don't exit Neovim. Instead, close the terminal directly (maybe wait a moment before you close the terminal, because the `updatetime` is default to 4 sec.) Then open another terminal and open the file again through the starter.

Aha, finally! Not writing the buffer was the key step I was missing. Thanks!

Unfortunately, I can reproduce both the swap message and the following error, but after pressing `e` or `o` the file opens as expected (i.e. no empty buffer).

wenjinnn:
Oops. This shocked me a bit.  in the end, it still has something to do with `mini.misc`. To be precise, it's related to the `BufReadPre` autocmd set in the `setup_restore_cursor` of mini.misc, because that's where I initially discovered the issue. At that time, I changed the event to `BufReadPost`, and it worked fine. For convenience, I'll still share my test code here.
```lua
for name, url in pairs {
  ["mini.nvim"] = "https://github.com/echasnovski/mini.nvim",
} do
  local install_path = vim.fn.fnamemodify("nvim_issue/" .. name, ":p")
  if vim.fn.isdirectory(install_path) == 0 then
    vim.fn.system { "git", "clone", "--depth=1", url, install_path }
  end
  vim.opt.runtimepath:append(install_path)
end

require("mini.starter").setup()
require("mini.misc").setup()
MiniMisc.setup_restore_cursor()

```

echasnovski:
I tried it both with `nvim -u minimal.lua` and `NVIM_APPNAME=nvim-repro nvim` with the following config file:

```lua
vim.fn.system(clone_cmd)

vim.cmd('packadd mini.nvim')

require('mini.starter').setup()
```

The only difference I see is that in the gif there is also a "NEWER than swap file" message, which I don't get with the way swap file is created. Maybe it matters?

Showing extra error is okay, but showing empty buffer is not. So I'd need something more to reproduce this before making any decision about what to do next.

echasnovski:
> Oops. This shocked me a bit. in the end, it still has something to do with `mini.misc`. To be precise, it's related to the `BufReadPre` autocmd set in the `setup_restore_cursor` of mini.misc, because that's where I initially discovered the issue.

Aha! Yes, now I can reproduce. I'll take a closer look, as this might be better handled in 'mini.misc' then.

wenjinnn:
Sure! feel free to close this PR if you find a better way to fix this.

echasnovski:
As far as I can tell, this is already fixed on Neovim Nightly. The same repro steps show empty buffer with Neovim 0.10.3 but don't show it with Neovim Nightly (NVIM v0.11.0-dev-1545+ga4f575abd8 in my case).

The extra error which requires pressing extra `<CR>` is an unfortunate byproduct of `vim.cmd('edit ...')` which is currently the most robust way of opening a file.

So closing this as already fixed on newer versions of Neovim and fixing for earlier ones (even for current stable) would change either 'mini.starter' or 'mini.misc' usability (changing 'BufReadPre' to 'BufReadPost' is probably not a good idea as 'FileType' event is usually triggered in between of them).

echasnovski:
There now shouldn't be an extra error shown, only a single message about swap. I found out that using `vim.fn.bufadd()` together with `pcall(vim.api.nvim_win_set_buf, ...)` works well here.

wenjinnn:
> There now shouldn't be an extra error shown, only a single message about swap. I found out that using `vim.fn.bufadd()` together with `pcall(vim.api.nvim_win_set_buf, ...)` works well here.

That's nice to hear! Thanks for the notice

PR #1484, opened 2025-01-15
fix(math/convert): [mini.base16] `xyz2rgb` range miss accurate.

AgatZan:

@echasnovski Sorry if it's just me, but I have found a problem with implementation of the paper on concertation. After `luv2xyz` XYZ is in the range [0,1] and `xyz2rgb` [XYZ is also in the nominal range of [0, 1].](http://www.brucelindbloom.com/Eqn_XYZ_to_RGB.html). But all axios `c/100` without any reason.

Also, I don't like the algorithm change, because it is really confusing when I compare it to the original. I understand that it has been made in accordance with the provided link.

And also, I would be very grateful if you provided where you got the conditions from, [like](https://github.com/echasnovski/mini.nvim/blob/9e603a31d3fbc7ee61c9b556cd84e97fa2dcff53/lua/mini/base16.lua#L1487).

### DIFF
--- a/lua/mini/base16.lua
+++ b/lua/mini/base16.lua
@@ -1463,7 +1463,6 @@ H.xyz2rgb = function(xyz)
   -- stylua: ignore end
 
   return vim.tbl_map(function(c)
-    c = c / 100
     if c > 0.0031308 then
       c = 1.055 * (c ^ (1 / 2.4)) - 0.055
     else


echasnovski:
> After `luv2xyz` XYZ is in the range [0,1] and `xyz2rgb` [XYZ is also in the nominal range of [0, 1].](http://www.brucelindbloom.com/Eqn_XYZ_to_RGB.html). But all axios `c/100` without any reason.

It is not. You can check it with something like this:
- Apply the following changes:
    ```diff
    diff --git a/lua/mini/base16.lua b/lua/mini/base16.lua
    index 989212e..0b79978 100644
    --- a/lua/mini/base16.lua
    +++ b/lua/mini/base16.lua
    @@ -1501,6 +1501,7 @@ H.xyz2luv = function(xyz)
       return { l = l, u = u, v = v }
     end
     
    +_G.log = {}
     H.luv2xyz = function(luv)
       if luv.l == 0 then return { x = 0, y = 0, z = 0 } end
     
    @@ -1517,6 +1518,7 @@ H.luv2xyz = function(luv)
       local y = var_y * 100
       local x = -(9 * y * var_u) / ((var_u - 4) * var_v - var_u * var_v)
       local z = (9 * y - 15 * var_v * y - var_v * x) / (3 * var_v)
    +  table.insert(_G.log, { x = x, y = y, z = z })
       return { x = x, y = y, z = z }
     end
     
    ```
- Run the following command `:lua require('mini.base16').mini_palette('#112233', '#ddeeff')`.
- Check the log with `:=log`. Values are in the [0; 100] range.

Main source of color conversion algorithm is [documented here](https://github.com/echasnovski/mini.nvim/blob/9e603a31d3fbc7ee61c9b556cd84e97fa2dcff53/lua/mini/base16.lua#L1397) which indeed contains those `100 * ` and ` / 100` steps.

> And also, I would be very grateful if you provided where you got the conditions from, [like](https://github.com/echasnovski/mini.nvim/blob/9e603a31d3fbc7ee61c9b556cd84e97fa2dcff53/lua/mini/base16.lua#L1487).

There is later a linear combination of `x`, `y`, `z` in the denominator, so they can't be zero all at once. As they are non-negative, `x + y + z == 0` is a more compact way of writing `x == 0 and y == 0 and z == 0` for which the Luv output is known (as it is a `#000000` color). There might have also been some floating point numeric considerations here, but I am not really sure.

PR #1492, opened 2025-01-18
feat(mini.extra): implement pastedtext text object

rainerborene:

Inspired by [saaguero/vim-textobj-pastedtext](https://github.com/saaguero/vim-textobj-pastedtext).

### DIFF
--- a/doc/mini-extra.txt
+++ b/doc/mini-extra.txt
@@ -174,6 +174,18 @@ Return ~
 `(function)` Function implementing |MiniAi-textobject-specification|.
 
 ------------------------------------------------------------------------------
+                                            *MiniExtra.gen_ai_spec.pastedtext()*
+                        `MiniExtra.gen_ai_spec.pastedtext`()
+A convenient text object that selects the last pasted text.
+
+Notes:
+- Both `a` and `i` text objects will return the same range, defined
+  by the markers `'[` and `']`.
+
+Return ~
+`(function)` Function implementing |MiniAi-textobject-specification|.
+
+------------------------------------------------------------------------------
                                                      *MiniExtra.gen_highlighter*
                           `MiniExtra.gen_highlighter`
 'mini.hipatterns' highlighter generators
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

I don't think this is a good fit for 'mini.extra' at least for the following reasons:
- There is already very similar functionality in ['mini.basics'](https://github.com/echasnovski/mini.nvim/blob/44a3ab2bc6b40edeb7a76359826c9779d5c70cb5/doc/mini-basics.txt#L239) under `gV` mapping. Yes, it is only for Visual mode, but it feels usually enough for this particular case. Maybe, should have it in Operator-pending mode there also :thinking: 
- It is fairly compact and straightforward to have in one's config verbatim.

Closing as "not the best fit for 'mini.extra'".

rainerborene:
@echasnovski No problem! Do you think it would be worth including a `fold` text object on `mini.extra`? Here’s what I currently have in my configuration:
```lua
z = function(ai_type)
  local foldlevel = vim.fn.foldlevel(vim.fn.line ".")
  if foldlevel <= 0 then
    return
  end
  vim.cmd.normal { "zxzc", bang = true }
  local start_line = vim.fn.foldclosed(vim.fn.line ".")
  local end_line = vim.fn.foldclosedend(vim.fn.line ".")
  vim.cmd.normal { "zo", bang = true }
  if ai_type == "i" then
    start_line = start_line + 1
    end_line = end_line - 1
  end
  return {
    from = { line = start_line, col = 1 },
    to = { line = end_line, col = 1 },
    vis_mode = "V",
  }
end,
```

echasnovski:
> Do you think it would be worth including a `fold` text object on `mini.extra`? Here’s what I currently have in my configuration:

In theory - maybe, but there is no good/"proper" way of dealing with folds in Neovim (yet). Having to close and open fold is not something I'd be comfortable having in 'mini.extra'. Maybe custom line traversing up and down while inspecting their `vim.fn.foldlevel()` would be good, but... Having to have *extensive* testing for many fold methods for this textobject doesn't bring me joy.

I feel bad for denying second time in less than 30 minutes, but let's not have it in 'mini.extra'. Maybe in the future when there is a 'mini.folds' or similar. Huge thanks for the enthusiasm, though!

PR #1516, opened 2025-01-23
Basics: Map gP to paste system clipboard before cursor in normal mode

W4RH4WK:
While `gp` is already mapped to paste from the system's clipboard, `gP` has not been adjusted accordingly. For symmetry, this should be added as well in normal mode.

This should not apply to visual mode as the selection is replaced, so we can keep `gP` in visual mode as is.


echasnovski:
Thanks for the PR!

Although reasonable, I think this kind of customization is better left to the user. There might be better uses for `gP` mapping in the future 'mini.nvim' modules.

PR #1526, opened 2025-01-25
feat(clue): allow mapping desc fallback to config clue desc

jameshurst:
This PR would allow keymaps without descriptions to fallback to the clue description as configured by the user. 
I had also noticed the comment indicating clue precedence was incorrect and corrected it. 

This would allow the user to create clues with dynamic descriptions by providing a function as the clue `desc` and omitting the keymap `desc`. For example, with this change it would be possible to create "toggle" keymaps with clues prefixed by an icon to indicate the toggle state.

However, there would be the drawback that should the user configure a clue description and then create a keymap without any description, then the clue may be incorrect.

I'm wondering what your thoughts are on this change. Would there be an alternative way to display clues with a dynamic description for something like toggle keymaps?

### DIFF
--- a/lua/mini/clue.lua
+++ b/lua/mini/clue.lua
@@ -1696,7 +1696,7 @@ end
 H.clues_get_all = function(mode)
   local res = {}
 
-  -- Order of clue precedence: config clues < buffer mappings < global mappings
+  -- Order of clue precedence: config clues < global mappings < buffer mappings
   local config_clues = H.clues_normalize(H.get_config().clues) or {}
   local mode_clues = vim.tbl_filter(function(x) return x.mode == mode end, config_clues)
   for _, clue in ipairs(mode_clues) do
@@ -1709,6 +1709,7 @@ H.clues_get_all = function(mode)
     -- - Fall back to possibly already present fields to allow partial
     --   overwrite in later clues. Like to add `postkeys` and inherit `desc`.
     res_data.desc = desc or res_data.desc
+    res_data.desc_fallback = res_data.desc
     res_data.postkeys = H.replace_termcodes(clue.postkeys) or res_data.postkeys
 
     res[lhsraw] = res_data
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

> However, there would be the drawback that should the user configure a clue description and then create a keymap without any description, then the clue may be incorrect.

Yes, and this is the main reason why description from the mapping is taken as is if there is a mapping present. It is a strong indicator that there is a mapping present that user might not know about and [it is documented](https://github.com/echasnovski/mini.nvim/blob/7a75a970d0bae09ef579e41dbb20551bdf596f31/doc/mini-clue.txt#L79-L82).

I am afraid I am going to close this as working as intended and not planned.

------

> Would there be an alternative way to display clues with a dynamic description for something like toggle keymaps?

Description of the mapping itself can be updated to show its up to date action. Either manually or via [`set_mapping_desc()`](https://github.com/echasnovski/mini.nvim/blob/7a75a970d0bae09ef579e41dbb20551bdf596f31/doc/mini-clue.txt#L632).

jameshurst:
> Description of the mapping itself can be updated to show its up to date action. Either manually or via [set_mapping_desc()](https://github.com/echasnovski/mini.nvim/blob/7a75a970d0bae09ef579e41dbb20551bdf596f31/doc/mini-clue.txt#L632).

Unfortunately I had tried that and the descriptions could easily get out of sync, such as when creating a `ts_highlight` toggle mapping before treesitter had been configured. Thank you anyways for taking a look!

echasnovski:
> I had tried that and the descriptions could easily get out of sync, such as when creating a `ts_highlight` toggle mapping before treesitter had been configured.

This does indeed look like an issue with creating a toggling mapping. And if there is a way to keep 'mini.clue' clues up to date (as this PR suggests it as a dynamic fallback), then calling `set_mapping_desc()` can be done at that same place. If there is mapping yet, then it can be created with the initial description.

jameshurst:
> And if there is a way to keep 'mini.clue' clues up to date (as this PR suggests it as a dynamic fallback), then calling set_mapping_desc() can be done at that same place.

It had just occured to me that this could work without any changes by providing the clue `desc` as a function that simply calls `set_mapping_desc`, as the `desc` function will be evaluated before the keymap `desc` overwrites it. Thank you for the *clues*!

PR #1533, opened 2025-01-28
fix: use non-deprecated versions of vim.validate

dundargoc:
This will remove stack traces from `:checkhealth vim.deprecated`.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -1268,22 +1268,32 @@ H.opened_buffers = {}
 -- File system information
 H.is_windows = vim.loop.os_uname().sysname == 'Windows_NT'
 
+local function validate(t)
+  if vim.fn.has('nvim-0.11') == 1 then
+    for k, v in pairs(t) do
+      vim.validate(k, v[1], v[2], v[3])
+    end
+  else
+    vim.validate(t)
+  end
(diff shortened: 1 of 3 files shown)


echasnovski:
Thanks for the PR!

This is an interesting coincidence, as I am *right now* fixing this with a bit more scope in mind (fully switching to custom helper and unifying error handling across modules). So I am afraid I'm going to close this.

echasnovski:
@dundargoc, this should now be fixed on latest `main`. 0de7ac99824a315b9ef37aba0161f70bdbb1e66b is what I had in mind, if you are curious.

dundargoc:
Looks good. Mostly curious, are you planning to re-use vim.validate once the minimum supported version is large enough or is this for long-term usage?

echasnovski:
With the current Neovim release schedule, this will happen in about two years (after possible 0.13.0 release). So I'll refrain from making such long term prediction :)

The advantages of current approach are that type checking errors neatly have `(mini.xxx)` prefix and it is even faster than the current Nightly fast `vim.validate()`. The disadvantage is that it requires 5 extra line per module (i.e. 200+ lines per 'mini.nvim'). Probably keeping current solution is fine, but we'll see.

PR #1535, opened 2025-01-29
feat(snippets): better integration with completion engines

abeldekat:

The PR's for both [nvim-cmp](https://github.com/hrsh7th/nvim-cmp/pull/2126) and [blink.cmp](https://github.com/Saghen/blink.cmp/pull/1035) have been merged, adding a `resubscribe` to those engines in order to avoid outdated completion items.

That solution, although fine, has the following disadvantages:

1. The `resubscribe` is an extra line in the user config, needed to connect `mini.snippets` and the completion engine.
2. It does involve some overhead.
3. As @Saghen said [here](https://github.com/Saghen/blink.cmp/pull/1035#issuecomment-2619796846), it is a bit "hacky".

This PR fixes the issue in `mini.snippets` itself.

The `autocommands` in the `MiniSnippetsTracking` group are created in the setup and never deleted. Thus, `mini.snippets` is always "first". Inside the callbacks, the first line is:

```lua
    if not H.is_tracking then return end
```

With these changes, `mini.snippets` works with `nvim-cmp` and `blink.cmp`. Personally, I think this is the best solution. Do feel free of course to close this PR, as I might have missed something.

I also think the solution is slightly more efficient if the user works with `callsnippets` from the `lsp`. In that case, most of the completion will involve `snippets`...

### DIFF
--- a/lua/mini/snippets.lua
+++ b/lua/mini/snippets.lua
@@ -1374,7 +1374,7 @@ MiniSnippets.session.stop = function()
   H.session_deinit(cur_session, true)
   H.sessions[#H.sessions] = nil
   if #H.sessions == 0 then
-    vim.api.nvim_del_augroup_by_name('MiniSnippetsTrack')
+    H.is_tracking = false
     H.unmap_in_sessions()
   end
   H.session_init(H.get_active_session(), false)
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

I afraid the choice of creating those autocommands only for the duration of snippet session(s) is very much deliberate for performance reasons.

They involve two of the most frequently triggered events (all of `TextChanged` family and `ModeChanged`). And although there is an early return via `H.is_tracking`, having those autocommands just be there introduces overhead. It is not huge by any means, but after taking into account the shear frequency of those events I am not comfortable doing that. I'd be okay if those autocommands themselves would do something useful frequently, but they would not (as the time spent inside active snippet session is far less compared to regular Neovim usage).

So I am going to close this as "works as intended". At least for now, maybe I'd revisit after 'mini.completion' snippet integration.

abeldekat:
> At least for now, maybe I'd revisit after 'mini.completion' snippet integration.

That's one reason I wrote the PR. It's likely you'll encounter the same problem.

abeldekat:
> This change doesn't really ensure that autocommand from 'mini.snippets' is created before autocommands from 'blink.cmp' and/or 'nvim-cmp'.

True. I do think it's reasonable to add the following constraint to the help: Please load `mini.snippets` before the completion engine.

Saghen:
> I got the equivalent of 0.013ms per 25 events (after 132 events of random typing). I imagine the numbers are even higher for older machines.

It's worth emphasizing that we likely can't accurately measure the duration since it's so short. We're likely measuring the overhead of grabbing and storing the time. I don't have an easy way to measure the overhead in neovim but I'd imagine it's negligible. 

> This change doesn't really ensure that autocommand from 'mini.snippets' is created before autocommands from 'blink.cmp' and/or 'nvim-cmp'. As the order in which plugins are loaded depends on the user who can lazy load beyond reasonable doubt. Not everyone will use 'lazy.nvim' and 'dependencies' field with standalone 'mini.snippets'.

Totally agree, perhaps we should bring up the idea of a priority system for autocmds in the neovim repo?

echasnovski:
> Totally agree, perhaps we should bring up the idea of a priority system for autocmds in the neovim repo?

[It was already rejected](https://github.com/neovim/neovim/issues/18279).

abeldekat:
Using mini.completion with snippet support the issue does not occur.

I wonder if there is a lesson to be learned here, a better approach then `resubscribe`.

echasnovski:
> Using mini.completion with snippet support the issue does not occur.

Yeah, I specifically tested against all use cases from the similar issues.

> I wonder if there is a lesson to be learned here, a better approach than `resubscribe`.

My *guess* as to why this happened remains the same. Both 'nvim-cmp' and 'blink.cmp' probably do the following sequence after completion item is accepted:
- Send `completionItem/resolve` request to LSP server(s) to resolve the item for possibly missed earlier `additionalTextEdits`.
- Remove currently placed word and start snippet session.
- After some (small) time, callback from LSP requests act on the text. But the problem is that they act on the version of text that was *before* starting snippet session, which can cause issue.

'mini.completion' does it a bit differently:
- Apply already present `additionalTextEdits` with [the following caveat](https://github.com/echasnovski/mini.nvim/blob/0a9259e95dbfcd07d5d506c56afd615e018acc4e/doc/mini-completion.txt#L117-L119). Do so while tracking possible changes to currently inserted word range (for the next step).
- Remove currently inserted word and insert snippet.

The solution here might be to check for `vim.b.changedtick` in the `completionItem/resolve` LSP callback: if it increased, do nothing. Not sure how well this goes with respect to usability, though.

abeldekat:
Thanks!

PR #1549, opened 2025-02-04
fix(doc): ensure section is not empty when marking optional

iamgideonidoko:

I encountered the error below that kept the doc generation failing for me, so I added a safe check.

Screenshot 2025-02-03 at 2 09 08 PM

### DIFF
--- a/lua/mini/doc.lua
+++ b/lua/mini/doc.lua
@@ -1019,6 +1019,7 @@ H.add_section_heading = function(s, heading)
 end
 
 H.mark_optional = function(s)
+  if #s == 0 or s.type ~= 'section' then return end
   -- Treat question mark at end of first word as "optional" indicator. See:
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Could you please provide an example annotation which triggers this error? This will be useful for tests.

echasnovski:
Thanks for the follow up!

The issue here is with `@private` itself, as it just removes all lines in the block. This makes *all* later sections act on sections without lines. It was reasonable as `@private` is assumed to be used last.

I'll take a closer look for a possibly more robust solution here.

echasnovski:
The underlying issue should now be resolved on latest `main`. I took a slightly more general approach of skipping processing sections which don't have lines in them. This even allowed to reduce total line count :)
Thanks again for spotting this and providing a reproducible example!

PR #1552, opened 2025-02-04
fix(indentscope): use of global and local variable is consistent with nvim

epheien:
After fixing, it can be enabled by whitelisting, and provide more flexibility.

### DIFF
--- a/lua/mini/indentscope.lua
+++ b/lua/mini/indentscope.lua
@@ -693,7 +693,10 @@ H.create_default_hl = function()
   vim.api.nvim_set_hl(0, 'MiniIndentscopeSymbolOff', { default = true, link = 'MiniIndentscopeSymbol' })
 end
 
-H.is_disabled = function() return vim.g.miniindentscope_disable == true or vim.b.miniindentscope_disable == true end
+H.is_disabled = function()
+  if vim.b.miniindentscope_disable ~= nil then return vim.b.miniindentscope_disable end
+  return vim.g.miniindentscope_disable == true
+end
 
 H.get_config = function(config)
   return vim.tbl_deep_extend('force', MiniIndentscope.config, vim.b.miniindentscope_config or {}, config or {})


echasnovski:
Thanks for the PR!

Could you, please, elaborate on why you think this is needed?

epheien:
> Thanks for the PR!
> 
> Could you, please, elaborate on why you think this is needed?

For example, help window of nvim-tree is a scratch buffer, and ignore all autocmd to open_win, so I cannot disable indentscope through autocmd.

Xnip2025-02-05_12-46-14

According to the original implementation, to disable indentscope in a specific buffer, only the blacklist method can be used (by setting b:miniindentscope_disable=true). It is difficult to achieve under the above circumstances.

So I want to enable indentscope by using the whitelist. To achieve this function, it is only necessary to modify the global and local usage logic of the `miniindentscope_disable` variable to be consistent with the behavior of neovim.

echasnovski:
> According to the original implementation, to disable indentscope in a specific buffer, only the blacklist method can be used (by setting b:miniindentscope_disable=true). It is difficult to achieve under the above circumstances.
> 
> So I want to enable indentscope by using the whitelist. To achieve this function, it is only necessary to modify the global and local usage logic of the `miniindentscope_disable` variable to be consistent with the behavior of neovim.

The present approach can work both for [selective automated disabling](https://github.com/echasnovski/mini.nvim/blob/87fc38a8a2b8d498cecb021d057a38ea2e9f5a79/doc/mini.txt#L345-L347) (i.e. blocklist) and [selective automated enabling](https://github.com/echasnovski/mini.nvim/blob/87fc38a8a2b8d498cecb021d057a38ea2e9f5a79/doc/mini.txt#L349-L355) (i.e. allowlist). The latter can be done by setting `vim.b.miniindentscope_disable = true` for *all* buffers while excluding only those which user wants to enable.

As far as I understand, having no way to automatically disable inside help window of 'nvim-tree/nvim-tree' is fully equivalent to having now way to automatically enable it. Both of them rely on *some* approach to detect when the buffer for the help window is created.

The current approach of treating `g:minixxx_disable` and `b:minixxx_disable` (i.e. [disable if at least one is `v:true`](https://github.com/echasnovski/mini.nvim/blob/87fc38a8a2b8d498cecb021d057a38ea2e9f5a79/doc/mini.txt#L318)) is more targeted towards the use case of "set `vim.g.minixxx_disable` to unconditionally disable certain buffer *everywhere*". I.e. even if some buffer-local disable variable is set to `false`.

I can see the merit of respecting actually set buffer-local variable over global one, but that would essentially mean that buffer-local variable has three possible states (`nil`, `false`, `true`) which slightly conflicts with how Lua treats both `nil` and `false` as falsy. If implementing this approach today, I *might* consider doing it the "three state way", but changing it right now feels more like an unnecessary chore (especially as I don't see a use case which makes it possible while impossible under current approach).

So I am going to close this as not planned.

------

The proper approach in this particular 'nvim-tree/nvim-tree' help window use case I *think* is to create custom `g?` mapping that would toggle help window and then set `vim.b.miniindentscope_disable = false`. For example, by following [this example](https://github.com/nvim-tree/nvim-tree.lua/blob/70825f23db61ecd900c4cfea169bffe931926a9d/doc/nvim-tree-lua.txt#L217-L243).

epheien:
> The present approach can work both for [selective automated disabling](https://github.com/echasnovski/mini.nvim/blob/87fc38a8a2b8d498cecb021d057a38ea2e9f5a79/doc/mini.txt#L345-L347) (i.e. blocklist) and [selective automated enabling](https://github.com/echasnovski/mini.nvim/blob/87fc38a8a2b8d498cecb021d057a38ea2e9f5a79/doc/mini.txt#L349-L355) (i.e. allowlist). The latter can be done by setting vim.b.miniindentscope_disable = true for all buffers while excluding only those which user wants to enable.

The issue is that many plugins may create scratch buffers and not set filetype in many cases (not limited to nvim-tree), which makes it quite difficult to automatically set variables locally within buffers.

The essence is that the `mini.indentscope` plugin generally works in normal buffers, which will set basic information such as filetype, and for those scratch buffers (which often have many strange behaviors, such as ignoring all autocmds), `b:miniindentscope_disable` must be set to be disabled, which is quite difficult to achieve in practice.

Therefore, I think it is necessary to modify the working method of this variable to be consistent with the local option behavior of neovim.

echasnovski:
> The essence is that the `mini.indentscope` plugin generally works in normal buffers, which will set basic information such as filetype, and for those scratch buffers (which often have many strange behaviors, such as no filetype, ignoring all autocmds), `b:miniindentscope_disable` must be set to be disabled, which is quite difficult to achieve in practice.

Scratch buffers can be identified by checking 'buftype' option. And if they block autocommands from executing, then changing the behavior won't make it easier for automated disabling.

> Therefore, I think it is necessary to modify the working method of this variable to be consistent with the local option behavior of neovim.

As was said in previous comment, any solution that can be done with this PR's suggested approach can be rewritten in the current approach. Although being consistent with how local buffer options work is good, it was (and is) not the goal here.

PR #1560, opened 2025-02-09
fix(all): alias replacement is now more accurate

ColinKennedy:

While working with mini.doc, I noticed that a class type was rendering as `table<string, integer>Bar`. While looking into it, I'd defined an alias called `Foo` and the class's name was originally `FooBar`. The alias replacement accidentally matched the `Foo` part of `FooBar` which resulted in `table<string, integer>Bar`.

This PR is a fix in order to prevent those issues from happening. It matches whole words.

<details>
<summary>See all example replacements</summary>

```lua
local function substitute_current(section_text, alias_name, alias_desc)
  local name_escaped = vim.pesc(alias_name)
  local desc_escaped = vim.pesc(alias_desc)

  return (section_text:gsub(name_escaped, desc_escaped))
end

local function substitute_suggested(section_text, alias_name, alias_desc)
  local name_escaped = vim.pesc(alias_name)
  local desc_escaped = vim.pesc(alias_desc)

  if string.find(desc_escaped, "|", 1, true) then
    desc_escaped = "(" .. desc_escaped .. ")"
  end

  local placeholder = "qqqq"
  section_text = section_text:gsub("([_%.]+)%f[%a]" .. name_escaped .. "%f[%A]", "%1" .. placeholder .. name_escaped)
  section_text = section_text:gsub("([^_%w]*)%f[%a]" .. name_escaped .. "%f[%A]", "%1" .. desc_escaped)
  section_text = section_text:gsub("([_%.]+)%f[%a]" .. placeholder .. name_escaped .. "%f[%A]", "%1" .. name_escaped)

  return section_text
end

local base_entries = {
  -- Exact match whole line
  {"Foo", "Bar"},
  {"Foot", "Foot"},
  {"FooBar", "FooBar"},

  -- Prefix
  {"value Foo", "value Bar"},
  {"name_here Foot", "name_here Foot"},
  {"thing FooBar", "thing FooBar"},

  -- Suffix
  {"Foo value", "Bar value"},
  {"Foot name_here description and words", "Foot name_here description and words"},
  {"FooBar thing", "FooBar thing"},

  -- Infix
  {"something_before Foo value", "something_before Bar value"},
  {"more_things Foot name_here description and words", "more_things Foot name_here description and words"},
  {"blah FooBar thing", "blah FooBar thing"},

  -- Optional
  {"Foo?", "Bar?"},
  {"Foot?", "Foot?"},
  {"FooBar?", "FooBar?"},

  -- Privates
  {"_Foo", "_Foo"},
  {"_Foo?", "_Foo?"},
  {"_Some.Foo.thing?", "_Some.Foo.thing?"},
}

print("Current, broken:")
for _, entry in ipairs(base_entries) do
  local input = entry[1]
  local expected = entry[2]

  -- Replacing Foo with Bar
  local found = substitute_current(input, "Foo", "Bar")

  print(string.format('    Old: "%s"', input))
  print(string.format('    Expected: "%s"', expected))
  print(string.format('    Found: "%s"', found))
  print("\n")
end

print("\nNew, suggested:")
for _, entry in ipairs(base_entries) do
  local input = entry[1]
  local expected = entry[2]

  -- Replacing Foo with Bar
  local found = substitute_suggested(input, "Foo", "Bar")

  print(string.format('    Old: "%s"', input))
  print(string.format('    Expected: "%s"', expected))
  print(string.format('    Found: "%s"', found))
  print("\n")
  assert(expected == found, string.format('Expected: "%s", got "%s".', expected, found))
end


local similar_name_entries = {
  -- Complex
  {"some_var_name fun(variable_name: var, another_var_here: var, last_var: var): (var | Bar)", "some_var_name fun(variable_name: mar, another_var_here: mar, last_var: mar): (mar | Bar)"}
}

print("Current, functions, broken:")

for _, entry in ipairs(similar_name_entries) do
  local input = entry[1]
  local expected = entry[2]

  -- Replacing Foo with Bar
  local found = substitute_current(input, "var", "mar")

  print(string.format('    Old: "%s"', input))
  print(string.format('    Expected: "%s"', expected))
  print(string.format('    Found: "%s"', found))
  print("\n")
end

print("\nNew, functions:")

for _, entry in ipairs(similar_name_entries) do
  local input = entry[1]
  local expected = entry[2]

  -- Replacing Foo with Bar
  local found = substitute_suggested(input, "var", "mar")

  print(string.format('    Old: "%s"', input))
  print(string.format('    Expected: "%s"', expected))
  print(string.format('    Found: "%s"', found))
  print("\n")
  assert(expected == found, string.format('Expected: "%s", got "%s".', expected, found))
end
```
</details>

In the example code above, `substitute_current` represents the current mini.doc alias replacement. And `substitute_suggested` is this PR. As you can see, `substitute_current` has many false positives and `substitute_suggested` is working in every defined case.

## How It Works
`%f[%a] ... %f[%A]` matches whole words, which is great for what we need. Just one problem - `_` and `.` are not considered part of a word. So `var` or `Foo` in `fun(some_var_name: string, another: some.Foo.blah): nil` might accidentally get matched even if we don't mean to.

To prevent that from happening, we find all of those matches, prefix it with a unique string (in this case, `qqqq`), then do the replacement (which otherwise is perfect except for that one bug) and then remove the `qqqq` placeholder again.

The result is an all-around better alias replacement outcome.

## Caveats
There's likely a simpler regex that can do the job. Just that considering `fun(foo: bar, fizz: buzz): (extra | thing)` syntax and the possibility of nested types, this was the best I could come up with, quickly.

### DIFF
--- a/lua/mini/doc.lua
+++ b/lua/mini/doc.lua
@@ -971,7 +971,17 @@ H.alias_replace = function(s)
       -- two values which might conflict if outputs are used as arguments.
       local name_escaped = vim.pesc(alias_name)
       local desc_escaped = vim.pesc(alias_desc)
-      s[i] = s[i]:gsub(name_escaped, desc_escaped)
+
+      if string.find(desc_escaped, "|", 1, true) then
+        desc_escaped = "(" .. desc_escaped .. ")"
+      end
+
+      local placeholder = "qqqq" .. name_escaped
+      local section_text = s[i]
+      section_text = section_text:gsub("([_%.0-9]+)%f[%a]" .. name_escaped .. "%f[%A]", "%1" .. placeholder)
+      section_text = section_text:gsub("([^_%w]*)%f[%a]" .. name_escaped .. "%f[%A]", "%1" .. desc_escaped)
+      section_text = section_text:gsub("([_%.0-9]+)%f[%a]" .. placeholder .. "%f[%A]", "%1" .. name_escaped)
+      s[i] = section_text
     end
   end
 end


lua/mini/doc.lua
```diff
@@ -971,7 +971,17 @@ H.alias_replace = function(s)
       -- two values which might conflict if outputs are used as arguments.
       local name_escaped = vim.pesc(alias_name)
       local desc_escaped = vim.pesc(alias_desc)
-      s[i] = s[i]:gsub(name_escaped, desc_escaped)
+
+      if string.find(desc_escaped, "|", 1, true) then
+        desc_escaped = "(" .. desc_escaped .. ")"
+      end
```
ColinKennedy, line 977: A small explanation of this code - When an alias contains multiple types, it can have some unintended effects on the final output.

For example

```
---@alias Foo Bar | Fizz | Buzz

---@class Thing
---@field item Foo?
```

If left alone it expands to

```
---@class Thing
---@field item Bar | Fizz | Buzz?
```

But that's not actually the correct meaning. The correct thing would be do to

```
---@class Thing
---@field item (Bar | Fizz | Buzz)?
```

This if statement checks for that possibility and, if found, preemptively adds ()s so there won't be any issues after substitution.

A downside to this approach is that if you have a field like `---@field item Foo` then it would expand to `---@field item (Bar | Fizz | Buzz)`. The ()s were unnecessary here but also harmless so I figure this is fine. If it's absolutely necessary that ()s are only added when needed a different approach would need to be taken.

echasnovski:
Thanks for the PR!

I agree that there needs to be a better handling in these edge cases, but not sure if this PR's approach for Lua patterns and adding brackets is optimal. I'll add comments in the review.

lua/mini/doc.lua
echasnovski: Yes, I don't think adding `()` where they not needed is good. I think it can be an extra step in the later chain of `gsub()` calls. So something like `gsub(name_escaped .. '%?', '(' .. desc_escaped .. ')?')`. This will trip if alias is not used as a type, but it might be a worthwhile compromise.

lua/mini/doc.lua
```diff
@@ -971,7 +971,17 @@ H.alias_replace = function(s)
...
+
+      local placeholder = "qqqq" .. name_escaped
```
echasnovski, line 979: I don't think I like using this "placeholder" approach. It usually indicates that some logic is not quite right. My first instinct tells me that similar effect can be achieved at least by an extra `gsub` step or slightly more complicated Lua patterns.

lua/mini/doc.lua
```diff
@@ -971,7 +971,17 @@ H.alias_replace = function(s)
...
+      local section_text = s[i]
+      section_text = section_text:gsub("([_%.]+)%f[%a]" .. name_escaped .. "%f[%A]", "%1" .. placeholder)
```
echasnovski, line 981: Can't alias names contain digits or underscore? They might not start with digits, but I didn't see any restriction on starting with underscore or ending with underscore/digit. So at least frontier pattern needs to adjust for that. Or even better is to rework the whole replacement to not use them at all. Which might be possible since alias name is used as an explicit Lua pattern and there is no `*` / `-` / `+` quantifiers for which frontier patterns are needed.

echasnovski (changes requested):

lua/mini/doc.lua
ColinKennedy: I think what you've said is a good suggestion but I don't have a lot time to translate your suggestion into working code.

echasnovski: No worries. I'll take a look later.

ColinKennedy:
Hi @echasnovski thank you for your review. On the whole I think your suggestions and points are valid however I'm really short on time so unfortunately I won't be able to take suggestions that are not explicit (e.g. code suggestions). However I can say definitively while this PR is 100% perfect it does generally produce better outcomes and at least this raises awareness of the problem. Maybe you or another can pick up where we left off and get something in the future that fits perfectly.

echasnovski:
> Maybe you or another can pick up where we left off and get something in the future that fits perfectly.

Yeah, I'll take a look later. And even then it won't be perfect, but hopefully more robust.

echasnovski:
Thanks again for bringing up this issue and providing useful examples!

I've went a bit different route which required a little refactoring and upgraded formatting of types, but I think 'mini.doc' should work reasonably well with aliases. It makes the replacement [based on the character before and after the alias's name](https://github.com/echasnovski/mini.nvim/blob/4dd209e4f03bacd46d6d87408b2529a8823112fb/lua/mini/doc.lua#L981-L990): if both of them are absent, whitespace, or special (taken form LuaCATS docs for allowed types), then replacement is performed. The smart enclosing into parenthesis should also be pretty robust.

ColinKennedy:
This is a great improvement! However it looks like there's a bug where `@field` and `@param`, which both expect the first word to be a variable / identifier, are able to be replaced by aliases. Idk if that was always an issue or a new bug but just wanted to mention. I'll make a separate issue + repro for that

PR #1586, opened 2025-02-19
feat(files): add autocommand events before performing file actions

mehalter:
The exposed autocommands make it very easy to add support for things such as the file operations defined in the LSP specification. Currently there are some events missing immediately before an operation is taken to support the rest of the events (`workspace/willCreateFiles`/`workspace/willDeleteFiles`/`workspace/willRenameFiles`). I'm sure there are other use cases as well where the user may want to hook into events right before an action is taken place.

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -354,13 +354,23 @@ with the following information:
 
 File action events ~
 
+- `MiniFilesActionCreatePre` - before entry is created.
+
 - `MiniFilesActionCreate` - after entry is successfully created.
+
+- `MiniFilesActionDeletePre` - before entry is deleted.
 
 - `MiniFilesActionDelete` - after entry is successfully deleted.
 
+- `MiniFilesActionRenamePre` - before entry is renamed.
+
 - `MiniFilesActionRename` - after entry is successfully renamed.
 
+- `MiniFilesActionCopyPre` - before entry is copied.
+
 - `MiniFilesActionCopy` - after entry is successfully copied.
+
+- `MiniFilesActionMovePre` - before entry is moved.
 
 - `MiniFilesActionMove` - after entry is successfully moved.
 
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -350,13 +350,23 @@
 ---
 --- File action events ~
 ---
+--- - `MiniFilesActionCreatePre` - before entry is created.
+---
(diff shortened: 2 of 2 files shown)


echasnovski:
Thanks for the PR!

I plan to add support for LSP file create/rename/delete directly into 'mini.files', as it seems to be possible to do so relatively concisely. If that would be the case, I'd like to not expose another set of events. My main concern is that there is no guarantee that a file system operation will actually happen.

> I'm sure there are other use cases as well where the user may want to hook into events right before an action is taken place.

Are you aware of any other such use cases? If not, I think I'd close this PR in favor of a future work I'd like to do in 'mini.files'.

mehalter:
I'm not sure off the top of my head of any other use cases of these events so all good if you want to just close this.

I'm also not sure how the specification guarantees validity in a rigorous manner for operations being unsuccessful. Especially because based on the semantics of the spec, the workspace edits also must be applied before the action takes place. So if you get the edit and apply it, I'm not sure of the means of reverting said edit in the case of the operation being unsuccessful (at least with the current neovim lua API).

Very exciting to hear you are adding the LSP support directly! I do agree it can be implemented relatively concise. One thing I would note with the implementation is to make sure to support the regex filtering on the files when deciding if you are sending the LSP requests to the client. I have seen several implementations of the LSP file operations and that is typically missed. If you are curious about some other implementations, AstroNvim's LSP plugin recently got an implementation in place that is pretty complete (and hopefully correct as well) that also implements some caching of the filtering resolution for performance: https://github.com/AstroNvim/astrolsp/blob/main/lua/astrolsp/file_operations.lua

mehalter:
`lua_ls` uses `didRename` with filters provided, granted the filter is just the root directory `<root_dir>/**` with `ignoreCase = true`

`vtsls` impelemnts `didRename` with filters

`rust-analyzer` implements `willRename` with filters

`basedpyright` implements `willRename` with filters

there are a lot of language servers it seems that support `didRename`/`willRename` but I have searched and can't find any that actually use the `{will,did}{Create,Delete}Files`

I also think all of the instances I have checked that implement `willRename` or `didRename` only implement one of them rather than both of them

echasnovski:
@mehalter, just a follow up. LSP integration is now [part of the `main` branch](https://github.com/nvim-mini/mini.nvim/blob/4f6f84a96b076747b736d100388a8d7844771d7d/doc/mini-files.txt#L320-L332).

PR #1631, opened 2025-03-02
feat(files): add an option to persist bookmarks

filiptrplan:

Added a feature to persist bookmarks to disk so they are saved when you open from the same directory again. Currently they are saved per current working directory and restored upon opening the explorer. 

If it's possible I would like some suggestions on how to better integrate it with the existing structure of the mini.files plugin (like the history and tabs). If the current approach with the cwd is good enough I would like just the go ahead before I start writing tests.

Tests aren't yet written because currently the way the bookmarks are persisted is subject to change.

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -606,6 +606,8 @@ Default values:
       permanent_delete = true,
       -- Whether to use for editing directories
       use_as_default_explorer = true,
+      -- Whether to persist bookmarks (per root directory)
+      persist_bookmarks = false
     },
 
     -- Customization of explorer windows
@@ -683,6 +685,13 @@ Note: to work with directory in |arglist|, do not lazy load this module.
 permanent delete or move into special trash directory.
 This is a module-specific variant of "remove to trash".
 Target directory is 'mini.files/trash' inside standard path of Neovim data
+directory (execute `:echo stdpath('data')` to see its path in your case).
+
+`options.persist_bookmarks` is a boolean indicating whether to persist bookmarks
+made while using this plugin. The bookmarks are persisted per current working directory, 
+that means that they get restored if you open neovim from the same directory you made
+them in.
+Target directory is 'mini.files/bookmarks' inside standard path of Neovim data
 directory (execute `:echo stdpath('data')` to see its path in your case).
 
 # Windows ~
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

I am afraid, this is not quite aligned with the 'mini.files' design (and 'mini.nvim' in general). I'd like to do persistent features only if absolutely necessary for the module's functionality. This results in both more secure and less complex code overall.

There is an example of doing a ["persistent" set of custom bookmarks](https://github.com/echasnovski/mini.nvim/blob/f40da50a3dfbbc53e6173beef8dc28626e1b5ebc/doc/mini-files.txt#L537-L552) in a sense that they will be present in any explorer. Extending the example to depend on some kind of path (be it explorer's anchor or cwd) should be straightforward with different code paths inside autocommand's callback.

Closing as not planned.

---

Some tips for the future:
- Please, create a feature request first before doing a PR ( as is suggested in 'CONTRIBUTING.md'). This will save both sides some time.
- Implementing a feature with +163 lines diff while +98 lines are responsible for the custom table serialization helper code is bound to be met with skepticism. Luckily, there is no need for that, as Neovim comes with `:h vim.inspect()`.

filiptrplan:
Thank you for the tips. I am quite new to the Lua/Nvim ecosystem so I appreciate the help.

All the best!

filiptrplan:
I then implemented the functionality of this PR with the `auto-session` plugin. I'm leaving my config here in the case someone would want to use this functionality.

```lua
{
  'rmagatti/auto-session',
  lazy = false,
  opts = {
    save_extra_cmds = {
      function()
        if MiniFilesBookmarks == nil then
          return
        end
        local t = vim.inspect(MiniFilesBookmarks):gsub('[\n\r]', ' ')
        return 'lua MiniFilesBookmarks=' .. t
      end,
    },
  },
}
```

```lua
require('mini.files').setup()

-- Bookmarks autosave
local set_mark = function(id, path)
  MiniFiles.set_bookmark(id, path)
end
  pattern = 'MiniFilesExplorerOpen',
  callback = function()
    if MiniFilesBookmarks ~= nil then
      for id, mark in pairs(MiniFilesBookmarks) do
        set_mark(id, mark.path)
      end
    end
  end,
})
local save_marks = function()
  local e = MiniFiles.get_explorer_state()
  E = e
  if e ~= nil then
    MiniFilesBookmarks = e.bookmarks
  end
end
  pattern = 'MiniFilesExplorerClose',
  callback = save_marks,
})
```

PR #1648, opened 2025-03-07
fix(completion): validation of process_items

Lattay:
This patch addresses a recent regression where an error would be raised if lsp_completion.process_items was set.
The validation now behaves as expected: it raises if lsp_completion.process_items is set to something that is not a function.


echasnovski:
There is always something obvious missing :( And this time it is a big one. Thanks for noticing and the PR! Should be resolved on latest `main`.

Issue #1657, opened 2025-03-09
snippets: Comment string parts not following "*:*" pattern break snippet completion


shuber2:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [x] I have updated 'mini.nvim' to latest version of the `main` branch

### Module(s)

mini.snippets

### Neovim version

0.10.x

### Description

I wrote a `mail.json` snippet file that caused the following error upon any completion:

```
E5108: Error executing lua vim/shared.lua:0: s: expected string, got nil
stack traceback:
        [C]: in function 'error'
        vim/shared.lua: in function 'validate'
        vim/shared.lua: in function 'trim'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:2533: in function 'get_comment_leaders'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:2514: in function 'get_indent'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:2041: in function 'session_init'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:1308: in function 'call_in_insert_mode'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:1306: in function 'insert'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:1690: in function 'call_in_insert_mode'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:1682: in function 'insert'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:1153: in function 'select'
        /home/shuber/.vim/plugged/mini.nvim/lua/mini/snippets.lua:798: in function 'expand'
        [string ":lua"]:1: in main chunk
```

The underlying reason is that in `H.get_comment_leaders()` in `lua/mini/snippets.lua` fails if `comment_part` happens to be an empty string (or would not contain a ":").

### Reproduction

[...]

PR #1658, opened 2025-03-09
fix: Sanitize H.get_comment_leaders() in mini.snippets

shuber2:
When comment_part in H.get_comment_leaders() is empty we get a nil value for suffix, which then breaks vim.trim(suffix). Sanitize this function for this case.

Fixes bug #1657.


echasnovski:
Thanks for the PR!

As described in [this comment](https://github.com/echasnovski/mini.nvim/issues/1657#issuecomment-2708969108), I don't think it is a good idea to have this check.

Issue #886, opened 2024-05-17
snippet support for mini.completion


alekspickle:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

completion

### Description

I already created the issue #882 about having $0 after completion.
I also then created an issue on LSP side where they [explained](https://github.com/mrcjkb/rustaceanvim/issues/393) that this is what is coming to Neovim 0.10 natively.

But mini.completion isn't supporting snippets, correct? if so, would you be open to support it or accept PR? currently I have no idea about volume of work to be done for it, but sound interesting and might help me get familiar with all this Lua madness.

echasnovski:
Thanks for the suggestion!

Yes, [this is planned](https://github.com/echasnovski/mini.nvim/issues/574#issuecomment-1819531525) after the release of 'mini.snippets'. As 0.10 is released, this will now be possible to start working on. Yet, no concrete plans.

I appreciated the suggestion, but 'mini.completion' is one of the messiest module in 'mini.nvim' and I'll tackle it myself.

Let's have this issue open so as an extra measure to not forget about it.

flexagoon:
@echasnovski can the 0.11 LSP completion side effects api be used for this?

echasnovski:
> @echasnovski can the 0.11 LSP completion side effects api be used for this?

I am afraid I am not really sure what you're referring to. Could you link to it, please?

But if it is only 0.11, then it probably won't be used as I'd like 'mini.nvim' (and snippet support in 'mini.completion') to support as many versions as possible. Right now it should be possible to support 0.10+ but maybe after 'mini.snippets' this can get better.

echasnovski:
> I was referring to the `vim.lsp.util.apply_text_edits` function. It seems like unless you need to have custom snippet support then using it together with the new snippets API should provide a very simple way to expand most existing LSP completions?

I *think* just using `vim.snippet` should be enough. It is just that resolving this issue is intentionally postponed after 'mini.snippets' is a thing. So that I'd have more knowledge about snippets and can be sure that I've tried to support as many previous Neovim versions as possible.

flexagoon:
@echasnovski another thing that I thought:

[...]

PR #1672, opened 2025-03-12
fix(completion): handle edit range properly in `apply_item_defaults`

wenjinnn:
Hi echasnovski, thanks for another great work. I found out that `editRange` in `itemDefaults` didn't handle properly, according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem, the `texitEdit` in `completionItem` should be [TextEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit) or [InsertReplaceEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertReplaceEdit), but now if it's a `TextEdit`, the code here handle it as a [Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range), which it's just a part of `TextEdit` structure. this could cause some problem, for me it make `jdtls` didn't response with the completion item's documentation.
related to #886.

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -993,10 +993,12 @@ H.apply_item_defaults = function(items, defaults)
       -- Infer new text from `item.textEditText` designed for default edit case
       item.textEdit.newText = item.textEdit.newText or item.textEditText or item.label
       -- Default `editRange` is range (start+end) or insert+replace ranges
-      item.textEdit.start = item.textEdit.start or edit_range.start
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Yes, indeed, my bad. I'll handle this at least tomorrow (I'd like to avoid `if ... else ...` due to extra nesting and tests need updating).

echasnovski:
Should now be fixed on latest `main`. Thanks again for noticing this!

PR #1683, opened 2025-03-16
fix(completion): check lsp result for both err and error key

abeldekat:

This PR fixes the error mentioned in [this](https://github.com/echasnovski/mini.nvim/discussions/1681) discussion: Setting up completion + snippets fails.

I added `or item.error` to three locations, and tested on both v0.10.4 and nightly according to the setup used in the discussion.

I tried to copy and modify testcase `prefer snippet from resolved item`. Unfortunately I could not reliably test the premise.

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -981,7 +981,9 @@ H.process_lsp_response = function(request_result, processor)
 
   local res = {}
   for client_id, item in pairs(request_result) do
-    if not item.err and item.result then vim.list_extend(res, processor(item.result, client_id) or {}) end
+    if not (item.err or item.error) and item.result then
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

> I tried to copy and modify testcase `prefer snippet from resolved item`. Unfortunately I could not reliably test the premise.

No worries. I'll merge into backlog branch and add proper cover for error fields in `buf_request_all()` callback input.

echasnovski:
I've slightly adjusted commit wording and code changes. Here is the [separate commit with test updates](https://github.com/echasnovski/mini.nvim/commit/88f18a7d018d8bb4d32f6be4c712134a10e826d3), if you are curious.

abeldekat:
That's great. Thanks!

PR #1689, opened 2025-03-20
feat(misc): trigger nested autocommands in setup_auto_root

ojroques:

Hello,

I'm currently developping a small plugin that performs some actions when the current directory is changed. It works by defining autocommands triggered on [`DirChanged`](https://neovim.io/doc/user/autocmd.html#DirChanged) events.

However these autocommands are not triggered by `setup_auto_root`. This is because by default, autocommands are not triggered inside another autocommand ([source](https://neovim.io/doc/user/autocmd.html#autocmd-nested)).

This PR changes that by setting the `nested` option to `true` when creating the autocommand in `setup_auto_root`.

It should not trigger any other events than `DirChanged` and `DirChangedPre` since the callback of `setup_auto_root` basically only calls `vim.fn.chdir()`, and it seems logical to allow those two events to be triggered.

Thanks for this great suite of plugins by the way :)

### DIFF
--- a/lua/mini/misc.lua
+++ b/lua/mini/misc.lua
@@ -212,7 +212,7 @@ MiniMisc.setup_auto_root = function(names, fallback)
   local augroup = vim.api.nvim_create_augroup('MiniMiscAutoRoot', {})
   vim.api.nvim_create_autocmd(
     'BufEnter',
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

Yeah, this change makes sense. I'll merge it in a temporary branch to tweak commit a bit and later push to `main`.

PR #1716, opened 2025-03-30
feat(test): add timeout for `reference_screenshot`

phanen:

With this PR, we can wait for screenshot by:
```lua
    opts.timeout = opts.timeout or 1000
    opts.interval = opts.interval or 100
    MiniTest.expect.reference_screenshot(
      function() return child.get_screen_lines(screenshot_opts) end, path, opts)
```

### DIFF
--- a/lua/mini/test.lua
+++ b/lua/mini/test.lua
@@ -721,7 +721,7 @@ end
 
 --- Expect equality to reference screenshot
 ---
----@param screenshot table|nil Array with screenshot information. Usually an output
+---@param screenshot table|function|nil Array with screenshot information. Usually an output
 ---   of `child.get_screenshot()` (see |MiniTest-child-neovim.get_screenshot()|).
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

As I've said in the ibhagwan/fzf-lua#1936, I don't really the pattern of "wait unknown amount of time until the screen is like this". Having an explicit wait time and known in advance checks seems to be more robust testing long term.

Closing as not planned.

echasnovski:
Another reason for not willing to support this is that it breaks the whole workflow around screenshot testing. Right now adding an `expect_screenshot()` call (in the way that it will be later used for testing) will create the reference screenshot file on the first run and compare to it on later ones. While "wait for the right screenshot" workflow requires an extra manual step for creating a reference screenshot (by writing essentially a different test code) and *then* replacing it with "wait for the right one" code.

Issue #1339, opened 2024-11-15
Mark and go to next position


phrmendes:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.pick

### Description

Is there a way to use the `telescope.nvim` or `fzf` behavior so that when you select an entry, the selection goes to the next item below?

echasnovski:
Yes, there is, albeit not very convenient one. It requires making a custom mapping which emulates pressing "mark" + "go down" keys (thus should be mapped to some other key and not override default 'mark' keys, `<C-x>`). See [this comment](https://github.com/echasnovski/mini.nvim/issues/513#issuecomment-1764784504).

phrmendes:
Thanks!

PR #1729, opened 2025-04-05
Pass `actions` to custom mapping `<func>`

feelamee:
- **feat(pick): pass actions to custom mapping function**
- **docs(pick): document custom mapping function**

Hey, I suggest to pass table with actions to custom mapping function. This
allow more customization. E.g. I want to configure keybindings similar to fzf.
In fzf `<Tab>` mark current item and move to next. To reproduce this in
Mini.Pick I need access to actions in my custom mapping function.

How I used this you can see in my [config](https://github.com/feelamee/dotfiles/blob/5ce29b6d769ab51d33b12f0304fd5433e76a5a74/nvim/init.lua#L188).

This feature don't introduce breaking changes. That's why it can be easily merged.

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -2171,9 +2171,14 @@ H.picker_advance = function(picker)
     do_match = cur_action.name == nil or vim.startswith(cur_action.name, 'delete') or cur_action.name == 'paste'
     is_aborted = cur_action.name == 'stop'
 
+	local actions = {}
+	for name, fn in pairs(H.actions) do
+		actions[name] = function() return fn(picker, char) end
+	end
+
     local should_stop
     if cur_action.is_custom then
-      should_stop = cur_action.func()
+      should_stop = cur_action.func(actions)
     else
       should_stop = (cur_action.func or H.picker_query_add)(picker, char)
     end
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Although implementing "mark and move down" functionality does indeed look straightforward, I think this (constructing the all action functions) is a bit too much to do on every picker advance.

The recommended approach of using built-in actions is to get picker's `mapping` data and use `vim.api.nvim_input()` to simulate pressing those keys. This was discussed in #1339. Here is a more general solution for Neovim>=0.10:

```lua
local input_actions = function(actions)
  return function()
    local mappings = MiniPick.get_picker_opts().mappings
    local keys = vim.tbl_map(function(a) return mappings[a] end, actions)
    vim.api.nvim_input(vim.keycode(table.concat(keys)))
  end
end

require('mini.pick').setup({
  mappings = {
    -- Don't use built-in <Tab> and <S-Tab>
    toggle_preview = '<C-l>',
    toggle_info = '<C-h>',
    scroll_left = '',
    scroll_right = '',

    -- Use <Tab> and <S-Tab> to mark+move
    mark_and_down = { char = '<Tab>', func = input_actions({ 'mark', 'move_down' }) },
    mark_and_up = { char = '<S-Tab>', func = input_actions({ 'mark', 'move_up' }) },
  },
})
```

So I am going to close this as "a bit too much because it is already possible".

feelamee:
Thanks!
> Although implementing "mark and move down" functionality does indeed look straightforward

ofc simulating key pressings to call needed functions is less straightforward than just calling this functions directly, but thanks for solution.
Probably this is better than adding changes .-.

PR #1745, opened 2025-04-17
feat(trailspace): continue highlighting whitespace in insert mode

dtor:
Continue highlighting trailing whitespace in insert mode so that trailing whitespace is clearly visible.

To avoid highlighting spaces when typing normally, replace the standard trailing whitespace regex with one that does negative lookbehind to check the cursor position when entering insert mode. The regex ensures that highlighting does not happen if cursor is at the very end of the current line.

### DIFF
--- a/lua/mini/trailspace.lua
+++ b/lua/mini/trailspace.lua
@@ -97,20 +97,29 @@ MiniTrailspace.config = {
 
 -- Module functionality =======================================================
 --- Highlight trailing whitespace in current window
-MiniTrailspace.highlight = function()
-  -- Highlight only in normal mode
-  if H.is_disabled() or vim.fn.mode() ~= 'n' then
-    MiniTrailspace.unhighlight()
-    return
-  end
+MiniTrailspace.highlight = function(ev)
+  MiniTrailspace.unhighlight()
+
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

The main idea of 'mini.trailspace' is to highlight trailing whitespce [only in Normal mode](https://github.com/echasnovski/mini.nvim/blob/a8b1d74f45dadf3554c0eed784c6ccd8ade67883/doc/mini-trailspace.txt#L9-L10). This is because I find it distracting to have them visible during typing. While this PR's approach of "highlight all but don't highlight if on the end of lines in Insert mode" is interesting, I find the end result too close to already built-in Neovim capability of `trail:c` flag in 'listchars' option. Yes, it will still show trailing whitespace even if on the end of the current line, but still quite similar.

If anything, having in Insert mode pattern something like `\%#\s\+$` (highlight only to the right of cursor) might be interesting, but my first instinct is that it is of small-ish use while being a breaking change (I'd like to avoid adding new config options).

So all in all, thank you for the PR with an interesting idea, but I think I am going to close this as not planned (at least in the current version; I might change my mind about `\%#\s\+$` pattern).

PR #1784, opened 2025-05-05
feat(icons): update css icon

skoch13:

Nerd Fonts now [includes](https://github.com/ryanoasis/nerd-fonts/releases/tag/v3.4.0) updated official CSS icon, so this PR updates mini.icons accordingly.

### DIFF
--- a/lua/mini/icons.lua
+++ b/lua/mini/icons.lua
@@ -1100,7 +1100,7 @@ H.filetype_icons = {
   csdl                   = { glyph = '󰫰', hl = 'MiniIconsCyan'   },
   csh                    = { glyph = '', hl = 'MiniIconsGrey'   },
   csp                    = { glyph = '󰫰', hl = 'MiniIconsAzure'  },
-  css                    = { glyph = '󰌜', hl = 'MiniIconsAzure'  },
+  css                    = { glyph = '', hl = 'MiniIconsAzure'  },
   csv                    = { glyph = '', hl = 'MiniIconsGreen'  },
   csv_pipe               = { glyph = '', hl = 'MiniIconsAzure'  },
   csv_semicolon          = { glyph = '', hl = 'MiniIconsRed'    },


krovuxdev:
Regarding the `nerdfonts`, there were no changes to the CSS icon. Instead, a new icon was added, as mentioned in this [commit](https://github.com/ryanoasis/nerd-fonts/commit/fa7b859994228a9c8759f99c55a8d31ee92a1b5e#diff-d953ede3105e41a2f4a93de933a2fee961492da7a445483978895ee5f37a7cd0R1370-R1372). Therefore, I don't believe the current icon (`f031c`) is deprecated.

Based on the image, the first code `e6b8` is the new icon, while `f031c` is the current one in use. This means that `f031c` is not deprecated.
imagen

krovuxdev:
Although the icon can be changed, this is mentioned [here](https://github.com/echasnovski/mini.nvim/blob/473cdbff498ed120f1768c75ef65fa42859aa14f/doc/mini-icons.txt#L231-L249). In the meantime, the current icon will remain in use until it becomes obsolete (possibly in the future).

skoch13:
New official icon was added in the scope of https://github.com/ryanoasis/nerd-fonts/pull/1762, so I guess those who want to preserve the non-official can do so with the settings modifications you mentioned.

krovuxdev:
> New official icon was added in the scope of https://github.com/ryanoasis/nerd-fonts/pull/1762, so I guess those who want to preserve the non-official can do so with the settings modifications you mentioned.


I understand, thank you for creating the new NerdFonts icon, it is very nice. However, I'm sorry, I am against this PR as it could affect users who are not up to date, forcing them to update on Linux or Windows, which they might not like. As I mentioned, the preferred icon can be changed, as I indicated earlier

echasnovski:
Thanks for the PR!

For a wider support, 'mini.icons' [prefers 'nf-md'*' glyphs](https://github.com/echasnovski/mini.nvim/blob/473cdbff498ed120f1768c75ef65fa42859aa14f/doc/mini-icons.txt#L305-L306) if they are present. This is the case with ['css' filetype](https://github.com/echasnovski/mini.nvim/blob/473cdbff498ed120f1768c75ef65fa42859aa14f/lua/mini/icons.lua#L1103). Exceptions can be made, of course, but I don't think this is the case with CSS (at least yet).

Closing as not planned (maybe can be revisited after significant amount of time has passed for the new-ish icon to be widely available; maybe after 4.0.0 Nerd font release).

Issue #651, opened 2024-01-11
Add clues for `[` mappings


nkakouros:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.clue

### Description

Typing `help [` and pressing Tab shows a long list of `[` movements. It would be nice to have clues for them similar to `g`, etc.

echasnovski:
Thanks for the suggestion!

This sounds reasonable. I'll see if it will add too many lines to the code. Otherwise I might leave this to users.

doomgutt:
I'll piggyback and ask here so as to not open a new issue - is there or will there be any support for default vim mappings such as `d`, `y` etc? If no, do you think there's a convenient way to mooch off of which-key? Thanks.

PR #1789, opened 2025-05-07
feat(extra): add colorscheme picker

pkazmier:

### DIFF
--- a/lua/mini/align.lua
+++ b/lua/mini/align.lua
@@ -1617,13 +1617,11 @@ H.default_action_merge = function(parts, opts)
   end
 
   -- Do not change indentation
-  if #delimiter_arr > 0 then
-    local first_parts_are_indent = true
-    for i = 1, dims.row do
-      first_parts_are_indent = first_parts_are_indent and H.is_whitespace(parts[i][1])
-    end
-    delimiter_arr[1] = first_parts_are_indent and delimiter_arr[1]:gsub('^%s*', '') or delimiter_arr[1]
+  local first_parts_are_indent = true
+  for i = 1, dims.row do
+    first_parts_are_indent = first_parts_are_indent and H.is_whitespace(parts[i][1])
   end
+  delimiter_arr[1] = first_parts_are_indent and delimiter_arr[1]:gsub('^%s*', '') or delimiter_arr[1]
 
   return vim.tbl_map(function(row) return H.concat_array(row, delimiter_arr) end, parts)
 end
--- a/tests/test_align.lua
+++ b/tests/test_align.lua
@@ -953,13 +953,6 @@ T['gen_step']['default_merge()']['works'] = function()
   )
 end
 
-T['gen_step']['default_merge()']['indentity for empty list and zero-length strings'] = function()
-  validate_align_strings({}, {}, {})
-  validate_align_strings({ '' }, {}, { '' })
-  validate_align_strings({ '', '' }, {}, { '', '' })
-  validate_align_strings({ '', ' ' }, {}, { '', ' ' })
-end
-
 T['gen_step']['default_merge()']['verifies relevant options'] = function()
   expect.error(
     function() child.lua([[MiniAlign.align_strings({ 'a' }, { merge_delimiter = 1 }, {})]]) end,


pkazmier:
First draft. Next steps:

1. Use 'mini.colors' (if available) to get and restore colorscheme. This will restore any changes a user has made in between setting their original colorscheme and running this picker. Also, if a colorscheme does not set vim.g.colors_name.
2. Add tests.

pkazmier:
With regards to the placement of the one-time autocommand (didn't know these were a thing, very cool). I placed it in the `items` callback as that ensures we are already in a colorscheme picker, so presumably we are guaranteed that the `MiniPickStop` will be executed. Just trying to make sure we don't leave any unexecuted one-time auto commands. For example, if the user somehow interrupted between the setting of the auto command and `pick_start`. I don't really know if that is a possibility or not, but that was my logic.

And, I was unsure of how defensive you like to get, but I suppose to be really cautious, I could create a auto command group that clears any old unexecuted auto commands when the picker is created. That prevents stacking of more than one auto commands that would load a theme. And, also, to be extra safe, I could check the source name in the callback to make sure it was from this colorscheme picker. Again, I'm just not sure how defensive you want to be and if it is really necessary.

I'll defer to you as you are the expert.

pkazmier:
> I think the whole autocommand approach is unnecessary. Any picker waits for the user to finish action and resumes code execution afterwards. So the `load_colorscheme` equivalent can be done after `H.pick_start` call (while still the whole picker returning the output of `H.pick_start`).

Even better. I removed the autocommand and now preserve the `background` setting of the user. Tomorrow will look at the `mini.color` integration as lunch time is over 😄

pkazmier:
Added 'mini.color' support. Two versions of the picker are included: `colorschemes_1` and `colorschemes_2`.

The first version is more concise and does not try to avoid setting the colorscheme more than once. For example, in that version, if you open the picker and don't preview or select anything, when you quit, your original colorscheme is set. Likewise, if you preview a colorscheme (results in setting the colorscheme) and then select it, it will set the colorscheme again.

The second option is more verbose, but it avoids setting the colorscheme unnecessarily. So, my question to you is what is more important in this specific case: conciseness or avoiding the extra calls to set the colorscheme? As a novice, I guess I don't see the big deal if the colorscheme is multiple times and I think I favor the conciseness of version 1 given the spirt of mini being as small as possible.

Just looking for your feedback. Thanks.

lua/mini/extra.lua
```diff
@@ -1725,6 +1786,24 @@ H.choose_with_buflisted = function(item)
...
+  vim.o.background = background
+  vim.schedule(function()
+    if type(colorscheme) == 'string' then
+      vim.cmd('colorscheme ' .. colorscheme)
+    else
+      colorscheme:apply()
+      vim.cmd('doautocmd ColorScheme') -- trigger for things like mini.misc termbg_sync
+    end
+  end)
+end
```
echasnovski, line 1805:
```suggestion
H.set_colorscheme = vim.schedule_wrap(function(colorscheme, background)
  vim.o.background = background
  if colorscheme == vim.g.colors_name then return end
  if type(colorscheme) == 'string' then return vim.cmd('colorscheme ' .. colorscheme) end
  colorscheme:apply()
  -- Trigger to indicate actual color scheme application
  vim.cmd('doautocmd ColorScheme')
end)
```

This can be simplified. There shouldn't be any harm in doing this all inside `vim.schedule()`.

echasnovski (changes requested):
Yeah, let's try variant 1 first. There indeed shouldn't be too much harm (other than couple of milliseconds of delay) in applying color scheme.

lua/mini/extra.lua
```diff
@@ -384,6 +384,67 @@ MiniExtra.pickers.buf_lines = function(local_opts, opts)
...
+---@param local_opts __extra_pickers_local_opts
+---   Not used at the moment.
+---@param opts __extra_pickers_opts
+---
+---@return __extra_pickers_return
+MiniExtra.pickers.colorschemes_1 = function(local_opts, opts)
+  local pick = H.validate_pick('colorschemes')
+
+  local original_bg = vim.o.background
+  local selected_cs = H.get_colorscheme()
```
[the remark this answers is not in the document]
pkazmier: I love it when you talk about saving 5 lines! Makes me smile every time 😄  In all seriousness, I had a version where I called it more than once, which is how it ended up as a function. This is just a remnant.

lua/mini/extra.lua
```diff
@@ -384,6 +384,67 @@ MiniExtra.pickers.buf_lines = function(local_opts, opts)
...
+
+  local choose = function(item) selected_cs = item end
```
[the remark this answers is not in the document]
pkazmier: I forgot about your earlier comment about handling the marked scenario.  Thanks!

lua/mini/extra.lua
pkazmier: Thanks. I did not know about `schedule_wrap`.

To be honest, I don't really have a good intuition about when one needs to schedule something or not. For me, it's one of those things where if weird stuff happens, try schedule.

pkazmier (commented):
Thanks for the feedback and suggestions!

lua/mini/extra.lua
echasnovski: > For me, it's one of those things where if weird stuff happens, try schedule.

Seems pretty accurate to me :)

tests/test_extra.lua
```diff
@@ -905,6 +905,111 @@ T['pickers']['buf_lines()']['validates arguments'] = function()
...
+
+local pick_colorschemes = forward_lua_notify('MiniExtra.pickers.colorschemes')
+
+T['pickers']['colorschemes()']['works'] = function()
+  child.set_size(10, 80)
+  child.cmd('colorscheme default')
+
+  child.lua_notify('_G.return_item = MiniExtra.pickers.colorschemes()')
+  validate_picker_name('Colorschemes')
+  type_keys('ha')
```
echasnovski, line 918: Let's test it against something that is owned by 'mini.nvim'. Typing `mini` should be more robust.

pkazmier:
Wow. Your 'mini.test' library is very cool and the 'TESTING.md' document was perfect to get me started. I've now added some initial tests, but I need help the "cancel without mini.colors" test. When that test runs, I need to make sure that 'mini.colors' is not available or loaded. I tried different things but could not get it working. The point of this test is to prove that a custom change to a highlight group is not persisted when mini.colors is not present.

tests/test_extra.lua
pkazmier: Sure. I'll change all of them later tonight.

tests/test_extra.lua
```diff
@@ -905,6 +905,111 @@ T['pickers']['buf_lines()']['validates arguments'] = function()
...
+  eq(child.g.colors_name, 'habamax')
+
+  -- Cancel picker and validate restoration
+  type_keys('<C-c>')
+  eq(child.g.colors_name, 'default')
+  eq(child.api.nvim_get_hl(0, { name = 'Normal' }), custom_normal_hl)
+end
+
+T['pickers']['colorschemes()']['cancel without mini.colors'] = function()
+  -- FIXME: How do I unload mini.colors to test without it?
```
echasnovski, line 988: The only way seems to be to mock `require` directly to error when the input is `'mini.colors'`. So something like this:

```lua
child.lua([[
  local require_orig = require
  require = function(modname)
    if modname == 'mini.colors' then error("module 'mini.colors' not found") end
    return require_orig(modname)
  end
]])
```

tests/test_extra.lua
```diff
@@ -905,6 +905,111 @@ T['pickers']['buf_lines()']['validates arguments'] = function()
   validate({ scope = '1' }, '`pickers%.buf_lines`.*"scope".*"1".*one of')
 end
 
+T['pickers']['colorschemes()'] = new_set()
+
+local pick_colorschemes = forward_lua_notify('MiniExtra.pickers.colorschemes')
+
+T['pickers']['colorschemes()']['works'] = function()
+  child.set_size(10, 80)
+  child.cmd('colorscheme default')
```
echasnovski, line 914: All tests already have a fixed color scheme. This makes testing on several Neovim versions possible (as Neovim<0.10 and Neovim>=0.10 have different default color schemes). Is there a particular reason to reset to default here?

pkazmier: Yes, the default mini.hues scheme you set in the minimal script does not set `vim.g.colors_name`, so I just chose `default` so it was set correctly.

echasnovski: > Yes, the default mini.hues scheme you set in the minimal script does not set `vim.g.colors_name`, so I just chose `default` so it was set correctly.

Then either setting some custom `vim.g.colors_name` in the `pre_case` hook for `T['pickers']['colorschemes()']` set or use `minischeme` should do the trick.

tests/test_extra.lua
echasnovski: A tip: to regenerate a screenshot without having to delete files manually, pass `force = true` to `child.expect_screenshot()`.

tests/test_extra.lua
pkazmier: Will fix tonight.

echasnovski:
Thanks, I will take a closer look after 0.16.0 release.

Approved CI to run. Don't worry about commit style checks, those will be squashed. Nice job for making tests pass on all supported versions!

pkazmier:
Thanks for giving me the opportunity to collaborate with you. I learned a LOT even with this very small contribution. And I always thought testing in Neovim would be a nightmare, but I was pleasantly surprised. The screenshot capability is very cool as is the use of the child.

Now back to revamping my mini configuration, which I’m redoing based on yours. I like using the top-level directories in the config folder. There is nothing much left in my Lua directory, which is so different than any config I’ve ever had since moving to vim from emacs three years ago.

echasnovski:
> Thanks for giving me the opportunity to collaborate with you. I learned a LOT even with this very small contribution. And I always thought testing in Neovim would be a nightmare, but I was pleasantly surprised. The screenshot capability is very cool as is the use of the child.

Thanks for taking interest! There are more "good-first-issue" type of feature requests, if you have time and want to continue (which I'll gladly accept).

> Now back to revamping my mini configuration, which I’m redoing based on yours. I like using the top-level directories in the config folder. There is nothing much left in my Lua directory, which is so different than any config I’ve ever had since moving to vim from emacs three years ago.

Yeah, using 'plugin/' for user config is somewhat counter-intuitive, but nice overall. It is not without its issues, though. For example, `nvim -u path/to/init.lua` won't work (as it will still source '~/.config/nvim/plugin/') and needs `nvim --noplugin -u path/to/init.lua`. And "bisecting config" is a bit more verbose as in its strictest form requires commenting out all lines in all files in 'plugin/' one-by-one. The other approach is to `:source` files one-by-one, but that is not 100% replacement.

pkazmier:
> There are more "good-first-issue" type of feature requests, if you have time and want to continue

Yes. I would like to help where I can after I've revamped my configuration. When done I'll take a look at your issues list for something that looks easy. Obviously, your definition of easy will be quite different from mine 😄

echasnovski:
> When done I'll take a look at your issues list for something that looks easy. Obviously, your definition of easy will be quite different from mine 😄

#651 and #1790 look fairly straightforward. Others - not so much.

pkazmier:
Shoot. I meant to submit that last commit as part of a different PR. I'll revert later as I have to head out now.

lua/mini/extra.lua
```diff
@@ -384,6 +384,41 @@ MiniExtra.pickers.buf_lines = function(local_opts, opts)
...
+---   Not used at the moment.
+---@param opts __extra_pickers_opts
+---
+---@return __extra_pickers_return
+MiniExtra.pickers.colorschemes = function(local_opts, opts)
+  local pick = H.validate_pick('colorschemes')
+
+  local original_bg = vim.o.background
+  local has_colors, colors = pcall(require, 'mini.colors')
+  local selected_cs = has_colors and colors.get_colorscheme() or vim.g.colors_name
```
abeldekat, line 404: When using the `mini.nvim` repository, `mini.colors` is always present. Perhaps it is better to only use `mini.colors` if it has been setup. Otherwise, `mini.colors` always parses the current `colorscheme`, which seems unnecessary.

```suggestion
  local selected_cs = MiniColors ~= nil and MiniColors.get_colorscheme() or vim.g.colors_name
```

abeldekat:
I am looking forward to using this new picker!

Would it be possible to have a `local_opt`, specifying a list of strings that represent colors I don't want to see? I currently [remove](https://github.com/abeldekat/nvim_pde/blob/436b928207e6a574c7126e709977b2257767f28e/lua/ak/config/editor/pick.lua#L209) builtin colors.

echasnovski:
> Would it be possible to have a `local_opt`, specifying a list of strings that represent colors I don't want to see? I currently [remove](https://github.com/abeldekat/nvim_pde/blob/436b928207e6a574c7126e709977b2257767f28e/lua/ak/config/editor/pick.lua#L209) builtin colors.

Hmm... Interesting idea. I wonder if the `ignore_builtins` option is better here instead of custom array of color scheme names. Although with a more flexible name array approach the `ignore_builtins` can be a documented example (and easily transformable into "show *only* builtins").

I plan to polish (still need to decide if there needs to be a more informative preview text) and merge this week.

echasnovski:
(A version of) This is now [merged into latest `main`](https://github.com/echasnovski/mini.nvim/blob/6c6ac83b59e8918e42a3d395282b2d555c288a92/doc/mini-extra.txt#L252). I decided to add couple of `local_opts` that I personally find useful. This visibly increased complexity and line count, but I am certain I'll use these features in the future when creating new highlight groups to test against popular color schemes.

Thanks again @pkazmier for the initial implementation and PR!

PR #1793, opened 2025-05-13
docs(colors): fix typo in MiniColors.setup()

abeldekat:

"animte" -> "animate"


echasnovski:
Thanks for the PR!

Note for future: if it is clear which module is talked about (like in case of commit messages for a single module), there is no need to use "module prefix". Only function name inside `` ` `` is enough.

PR #1798, opened 2025-05-14
fix(align): handle empty list and empty strings properly

pkazmier:
This fixes a bug introduced in b83127c36cc7a9f3cb9e16f5d9db1231d30c33f9. In summary, these calls started to fail after that commit:

```
MiniAlign.align_strings({})
MiniAlign.align_strings({""})
MiniAlign.align_strings({"", ""})
```

I noticed that a custom picker of mine broke recently, but hadn't had time to chase down what happened until this morning. I use `MiniAlign.align_strings` in a picker.source.show function to pretty things up with live grep. At the start of that picker, there are no items, so `align_strings` was being passed `{}` and erroring out. Prior to the change mentioned above, my picker worked without error.

The following PR addresses the issue. I also added tests (now that I recently learned how to do so with your awesome 'mini.test'—thank you).

### DIFF
--- a/doc/mini-colors.txt
+++ b/doc/mini-colors.txt
@@ -669,7 +669,7 @@ Module setup
 Calling this function creates a `:Colorscheme` user command. It takes one or
 more registered color scheme names and performs animated transition between
 them (starting from currently active color scheme).
-It uses |MiniColors.animte()| with default options.
+It uses |MiniColors.animate()| with default options.
 
 Parameters ~
 {config} `(table|nil)` Module config table. See |MiniColors.config|.
--- a/lua/mini/align.lua
+++ b/lua/mini/align.lua
@@ -1617,11 +1617,13 @@ H.default_action_merge = function(parts, opts)
   end
 
   -- Do not change indentation
-  local first_parts_are_indent = true
-  for i = 1, dims.row do
-    first_parts_are_indent = first_parts_are_indent and H.is_whitespace(parts[i][1])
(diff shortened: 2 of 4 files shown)


pkazmier:
if your curious, this is my custom live grep picker that aligns the output:

```
  local function map_gsub(items, pattern, replacement)
    return vim.tbl_map(function(item)
      item, _ = string.gsub(item, pattern, replacement)
      return item
    end, items)
  end

  local align_on_nul = function(buf_id, items, query)
    -- Because items is an array of blobs (contains a NUL byte), align_strings
    -- will not work because it expects strings. So, convert the NUL bytes to a
    -- unique (hopefully) separator, then align, and revert back.
    items = map_gsub(items, "%z", "#|#")
    items = MiniAlign.align_strings(items, {
      justify_side = { "left", "right", "right" },
      merge_delimiter = { "", " ", "", " ", "" },
      split_pattern = "#|#",
    })
    items = map_gsub(items, "#|#", "\0")

    -- Back to the regularly scheduled program :-)
    MiniPick.default_show(buf_id, items, query)
  end

  MiniPick.registry.grep_live_align = function()
    MiniPick.builtin.grep_live({}, {
      source = { show = align_on_nul },
    })
  end
```

And screenshot:
image

echasnovski:
Thanks for the PR and tests!

I've merged this in temporary branch to clean it up a bit before merging into `main`.

pkazmier:
Thanks!

PR #1803, opened 2025-05-16
feat(jump): add on_key listener for esc to stop jumping

cosmicbuffalo:

Hey, big fan. I noticed that sometimes I want to jump to something, then realize I wanted to jump to something else. Stopping the jump on cursor move works, but I happen to have muscle memory to mash escape whenever I want to "cancel" something. This PR adds a simple on_key listener for escape to stop jumping in a similar vein to the on_cursormoved callback

### DIFF
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -350,6 +350,7 @@ H.create_autocommands = function()
   au('CursorMoved', '*', H.on_cursormoved, 'On CursorMoved')
   au({ 'BufLeave', 'InsertEnter' }, '*', MiniJump.stop_jumping, 'Stop jumping')
   au('ColorScheme', '*', H.create_default_hl, 'Ensure colors')
+  vim.on_key(H.on_key)
 end
 
 H.create_default_hl = function() vim.api.nvim_set_hl(0, 'MiniJump', { default = true, link = 'SpellRare' }) end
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

This is already possible with custom mapping:

```lua
local jump_stop = function()
  if not MiniJump.state.jumping then return '<Esc>' end
  MiniJump.stop_jumping()
end
vim.keymap.set({ 'n', 'x', 'o' }, '<Esc>', jump_stop, { expr = true, desc = 'Stop jumping' })
```

It might be a good idea to have something like this built-in or documented as an example. I'll need to think about it.

---

In the future, please follow the 'CONTRIBUTING.md' and create an issue first before making a PR. This usually saves time for both parties.

echasnovski:
This is now [documented in help](https://github.com/echasnovski/mini.nvim/blob/18889a4993a2c6de64eb07f91e7ad8666793441a/doc/mini-jump.txt#L115-L121).

echasnovski:
> I think there might be a problem with the approach of using a custom key bind for this though, which is there may be multiple attempts to create a custom bind for esc.

There is currently a solution within 'mini.nvim' via [linked `MiniKeymap.map_multistep()`](https://github.com/echasnovski/mini.nvim/blob/18889a4993a2c6de64eb07f91e7ad8666793441a/doc/mini-jump.txt#L113). The 'mini.jump' behavior can be done via a separate step inside a single `<Esc>` mapping done by the user.

PR #1804, opened 2025-05-16
feat(jump): add get_target_callback option

cosmicbuffalo:

Another little PR to add an optional callback that triggers when mini.jump calls `get_target`. I'm mainly hoping to add this so that I can more easily integrate `eyeliner.nvim` like so:

```lua
{
	'echasnovski/mini.jump',
	dependencies = {
		"jinh0/eyeliner.nvim",
	},
	opts = {
		get_target_callback = function(backward)
			require("eyeliner").highlight({ forward = not backward })
		end
	},
	config = function(_, opts)
		require('mini.jump').setup(opts)
	end
}

### DIFF
--- a/doc/mini-jump.txt
+++ b/doc/mini-jump.txt
@@ -97,6 +97,11 @@ Default values:
     -- This also affects (purely informational) helper messages shown after
     -- idle time if user input is required.
     silent = false,
+
+    -- Optional callback to be triggered when waiting for target input
+    -- Useful for combining mini.jump with other plugins that also want
+    -- to modify f/t/F/T behavior.
+    get_target_callback = function(_backward) end
   }
 <
 ------------------------------------------------------------------------------
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -127,6 +127,11 @@ MiniJump.config = {
   -- This also affects (purely informational) helper messages shown after
   -- idle time if user input is required.
   silent = false,
+
+  -- Optional callback to be triggered when waiting for target input
+  -- Useful for combining mini.jump with other plugins that also want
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by echasnovski** [comment]
Thanks for the PR!

I don't mind adding the way to hook into the 'mini.jump' process here, but adding new config setting is not the best way here. It is probably better to trigger custom event, something like `MiniJumpStart` (and `MiniJumpStop` for completeness). I'll take a look.

**Comment by cosmicbuffalo** [comment]
I think a custom event would work too, but for my planned use case I would hope there would be some way to differentiate between jump starting in either the forward or backward direction. I thought a decent way to get that argument in there would be with a function, but either two different start events, or a single event with two different patterns passed could achieve the same result I was going for.

**Comment by echasnovski** [comment]
> I think a custom event would work too, but for my planned use case I would hope there would be some way to differentiate between jump starting in either the forward or backward direction. I thought a decent way to get that argument in there would be with a function, but either two different start events, or a single event with two different patterns passed could achieve the same result I was going for.

The usual Lua way to solve this is to populate a `data` field that is forwarded to Lua callbacks of the events.

Making sure that `MiniJump.state` has proper value when used inside those potential events is another possibility.

**Comment by echasnovski** [comment]
This now should be possible with the introduction of [events](https://github.com/echasnovski/mini.nvim/blob/14ce72476995255fc5cb919da8067f65865e1225/doc/mini-jump.txt#L56-L66).

The 'eyelilner' setup works something like this:

```lua
local eyeline = function() require('eyeliner').highlight({ forward = not MiniJump.state.backward }) end
vim.api.nvim_create_autocmd('User', { pattern = 'MiniJumpGetTarget', callback = eyeline })
```

Thanks again for the idea and PR!

PR #1826, opened 2025-05-26
docs(jump2d): fix typo

abeldekat:

I read through the code because of issue 1818 and found two spelling errors.

If you'd like, I can also push an extra commit:

- Change line 22 into:

```txt
---       before/at/after cursor line, etc.
---     Example: user can configure to look for spots only inside current
---     window at or after cursor line.

```

- Change line 79 into:

```txt
--- - Jump to line start using combination of options supplied in
```

- Change line 343 into:

```txt
---@param opts table|nil Configuration of jumping, overriding global and buffer
---   local values. Has the same structure as |MiniJump2d.config|
---   without <mappings> field. Extra allowed fields:
```

- Change line 360 into:

```txt
---   -- Jump to single character from user input
---   MiniJump2d.start(MiniJump2d.builtin_opts.single_character)
```

- Change line 561 into:

```txt
  if #spotters == 0 then
    return function() return {} end
  end

```

Line 561 always auto formats.

### DIFF
--- a/doc/mini-jump2d.txt
+++ b/doc/mini-jump2d.txt
@@ -241,7 +241,7 @@ Vimscript functions that operate only inside current window.
 ## View ~
 
 Option `view.n_steps_ahead` controls how many steps ahead to show along
-with the currently required label. Those future steps are showed with
+with the currently required label. Those future steps are shown with
 different (less visible) highlight group ("MiniJump2dSpotAhead"). Usually
 it is a good idea to use this with a spotter which doesn't result into many
 jump spots (like, for example, |MiniJump2d.builtin_opts.word_start|).


echasnovski:
Thanks for the PR!

I am currently (on and off) looking for an automated solution for typo detection in CI and local fixes. This is mostly to not have a ton of `docs: fix typo` commits, which pollutes the changelog during plugin updates.

So in the meantime I'd only have changes that fix factually incorrect documentation (like "Jump to word start" instead of intended "Jump to line start"). Preferably as a single `docs(jump2d): improve documentation` commit if there are several such changes.

echasnovski:
> * line 22: Corrects indentation.

Current formatting is correct as the example is for `Allowed lines:` and not overall customization.

> * line 360: I don't understand `(follow by typing one character)`.

It describes that after executing the call user should type one character.

> * line 561: Somehow, this single line always auto formats to 3 lines.

It does not. Probably you have StyLua version newer than 0.19.0 (which is [what is used for formatting](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md#formatting)). I've looked and there is now a stable version 2.1.0, so will have to change to use it in all modules.

> * line 79: The spotter in example is `line`, not `word`.
> 
> * line 343: Renders like `local values.config|` in help.

Those are indeed not intended. Please, include only them in a single commit together with typos already present in this PR.

PR #1842, opened 2025-06-03
fix(keymap): check if Blink menu is visible before accepting selection

toupeira:
## Description

`get_selected_item()` still returns the last selected item even after the completion menu is closed, which causes the `blink_accept` action to always be chosen instead of any other following steps.

## Context

I use Blink with the `super-tab` preset where `<Tab>` accepts the completion, and noticed that the multi-step keymap stopped working after the completion menu was shown at least once.

`mini.keymap` example setup:

```lua
      keymap.map_multistep('i', '<Tab>', {
        'blink_accept',
        'vimsnippet_next',
        'increase_indent',
        'jump_after_close',
    })
```

`blink.cmp` example setup:

```lua
    keymap = {
      preset = 'super-tab',

      -- mapped with `mini.keymap`
      ['<Tab>'] = {},
      ['<S-Tab>'] = {},
    }
```

### DIFF
--- a/lua/mini/keymap.lua
+++ b/lua/mini/keymap.lua
@@ -806,7 +806,7 @@ H.steps_builtin.cmp_prev   = { condition = H.is_visible_cmp,  action = H.make_cm
 H.steps_builtin.cmp_accept = { condition = H.is_selected_cmp, action = H.make_cmp_action('confirm') }
 
 H.is_visible_blink  = function() return H.has_module('blink.cmp') and require('blink.cmp').is_menu_visible() end
-H.is_selected_blink = function() return H.has_module('blink.cmp') and require('blink.cmp').get_selected_item() ~= nil end
+H.is_selected_blink = function() return H.has_module('blink.cmp') and require('blink.cmp').is_menu_visible() and require('blink.cmp').get_selected_item() ~= nil end
 H.make_blink_action = function(action) return H.make_cmd_lua_action('require("blink.cmp").' .. action .. '()') end
 
 H.steps_builtin.blink_next   = { condition = H.is_visible_blink,  action = H.make_blink_action('select_next') }


toupeira:
@echasnovski I noticed that the `pmenu_accept` and  `cmp_accept` actions don't seem to have the same issue, let me know if you think this should be fixed in `blink.cmp` instead :grinning: 

(I tested `pmenu_accept` by disabling Blink, and `cmp_accept` is a guess based on the `visible()` calls at https://github.com/hrsh7th/nvim-cmp/blob/b5311ab3ed9c846b585c0c15b7559be131ec4be9/lua/cmp/view/custom_entries_view.lua#L326 and https://github.com/hrsh7th/nvim-cmp/blob/b5311ab3ed9c846b585c0c15b7559be131ec4be9/lua/cmp/view/native_entries_view.lua#L120)

echasnovski:
Thanks for the PR!

Although the change looks okay, I *think* it worked as is a month or so ago. Maybe there were changes in 'blink.cmp' that made it not work. I'll take a closer look.

toupeira:
Had a quick look too: Nothing stood out in the Git log for `blink.cmp`, and I can also reproduce the problem with older versions like 98a9d4f from March, as well as with the current `main` branch (vs. the `1.*` tag I normally use).

Just to clarify: The multistep-mapping does work correctly until the first time the completion menu is shown, it's only afterwards that it stops working.

echasnovski:
Yes indeed it does look like an issue with 'blink.cmp' (Saghen/blink.cmp#1856), either with docs or code. And indeed possibly a long standing one.

I've pushed a fix with proper test updates. Should now work properly on `main` branch. Thanks again for the PR!

toupeira:
Thank you too! :bow:

PR #1843, opened 2025-06-04
fix(keymap): pmenu still visible when jump steps are activated

### LINKED ISSUE #1839: [mini.keymap] ensure that pmenu is closed before jump steps are triggered
### Contributing guidelines

### Module(s)

mini.keymap

### Description

https://github.com/user-attachments/assets/17b4b556-ee7a-4168-a9df-b94398d3c9ac

<details>
<summary>init.lua</summary>

```lua
---@diagnostic disable: undefined-global
local function setup_deps(path_to_site)
  local mini_path = path_to_site .. "pack/deps/start/mini.nvim"
  if not vim.uv.fs_stat(mini_path) then
    vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/echasnovski/mini.nvim", mini_path })
    vim.cmd("packadd mini.nvim | helptags ALL")
    vim.cmd('echo "Installed `mini.nvim`" | redraw')
  end
  require("mini.deps").setup({ path = { package = path_to_site } })
end
setup_deps(vim.fn.stdpath("data") .. "/site/")

vim.o.completeopt = "menuone,noselect,fuzzy,nosort"
vim.o.shortmess = "CFOSWaco"
vim.g.mapleader = " "
MiniDeps.now(function()
  require("mini.basics").setup()
  require("mini.notify").setup()
  require("mini.icons").setup()
  MiniIcons.tweak_lsp_kind()

  require("mini.snippets").setup({
    --
  })
  require("mini.completion").setup({ lsp_completion = { auto_setup = false, source_func = "omnifunc" } })

  MiniDeps.add("neovim/nvim-lspconfig")
  vim.api.nvim_create_autocmd("LspAttach", {
    group = vim.api.nvim_create_augroup("my.lsp", {}),
    callback = function(args) vim.bo[args.buf].omnifunc = "v:lua.MiniCompletion.completefunc_lsp" end,
  })
  vim.lsp.config("*", { capabilities = MiniCompletion.get_lsp_capabilities() })
  vim.lsp.config("lua_ls", {
    settings = {
      Lua = {
        completion = { callSnippet = "Replace" },
        diagnostics = { workspaceDelay = -1 }, -- don't analyze
        runtime = { version = "LuaJIT", path = vim.split(package.path, ";") },
        telemetry = { enable = false },
        workspace = { library = { vim.env.VIMRUNTIME, "${3rd}/luv/library" } },
      },
    },
  })
  vim.lsp.enable({ "lua_ls" })

  local map_multistep = require("mini.keymap").map_multistep

  local tab_steps = {
    "minisnippets_next",
    "minisnippets_expand",
    -- "pmenu_next",
    "jump_after_tsnode",
    "jump_after_close",
  }
  map_multistep("i", "<Tab>", tab_steps)

  local shifttab_steps = {
    "minisnippets_prev",
    -- "pmenu_prev",
    "jump_before_tsnode",
    "jump_before_open",
  }
  map_multistep("i", "<S-Tab>", shifttab_steps)
end)

```
</details>

I added a mapping based on the "extra smart" mappings as described in the [help](https://github.com/echasnovski/mini.nvim/blob/c39f7f506939f065d0d1af237e2ea3f252edd3d4/doc/mini-keymap.txt#L104)

```lua
local steps_ctrl_l = { "minisnippets_next", "jump_after_tsnode", "jump_after_close" }
```

I removed the `pmenu` step in order to keep the `<c-p>/<c-n>` navigation. However as the pmenu is not "captured" in the steps, it remains open when "jumping before/after".

In my [config](https://github.com/abeldekat/nvim_pde/blob/cf61da0cfee2810944f7bddf2ae8f1feb4d1c331/lua/ak/config/editing/snippets.lua#L26) I added the following to solve the problem:

```lua
local no_pmenu = { -- make sure pmenu is closed after snippets and before jump
  condition = function() return vim.fn.pumvisible() == 1 end,
  action = function()
    vim.schedule(function() -- see mini.snippets, H.hide_completion
      if vim.fn.mode() == "i" then vim.cmd('silent noautocmd call complete(col("."), [])') end
    end)
    return false
  end,
}
local steps_ctrl_l = { "minisnippets_next", no_pmenu, "jump_after_tsnode", "jump_after_close" }
MiniKeymap.map_multistep("i", "<C-l>", steps_ctrl_l)
local steps_ctrl_h = { "minisnippets_prev", no_pmenu, "jump_before_tsnode", "jump_before_open" }
MiniKeymap.map_multistep("i", "<C-h>", steps_ctrl_h)

```

The problem does not occur when using `blink` instead of `mini.completion`.

Do you think a builtin `no_pmenu` step would be appropriate in `mini.keymap`? If so, I would like to create a PR.

abeldekat:

Fixes #1839

### DIFF
--- a/lua/mini/keymap.lua
+++ b/lua/mini/keymap.lua
@@ -450,7 +450,10 @@ MiniKeymap.gen_step.search_pattern = function(pattern, flags, opts)
 
   local act = function()
     local had_match = vim.fn.search(pattern, flags, stopline(), opts.timeout, opts.skip)
-    if had_match ~= 0 then adjust_cursor() end
+    if had_match == 0 then return end
+
+    adjust_cursor()
+    H.hide_completion()
   end
 
   return { condition = function() return true end, action = function() return act end }
@@ -724,7 +727,7 @@ H.make_jump_tsnode = function(side)
       -- Output of `get_node_range` is 0-indexed with "from" data inclusive and
       -- "to" data exclusive. This is exactly what is needed here:
       -- - For "before" direction exact left end is needed. This will be used
-      --   in Insert mode and cursor weill be between target and its left cell.
+      --   in Insert mode and cursor will be between target and its left cell.
       -- - For "after" direction the one cell to right (after normalization) is
       --   needed because cursor in Insert mode will be just after the node.
       local from_row, from_col, to_row, to_col = vim.treesitter.get_node_range(node)
@@ -735,6 +738,7 @@ H.make_jump_tsnode = function(side)
       -- for "before" direction and non-Insert mode.
       if not (new_pos[1] == pos[1] and new_pos[2] == pos[2]) then
         pcall(vim.api.nvim_win_set_cursor, 0, new_pos)
+        H.hide_completion()
         new_pos = vim.api.nvim_win_get_cursor(0)
       end
       if not (new_pos[1] == pos[1] and new_pos[2] == pos[2]) then break end
@@ -868,6 +872,18 @@ H.get_row_cols = function(row) return vim.fn.getline(row + 1):len() end
 
 H.get_tsnode = function() return vim.treesitter.get_node() end
 if vim.fn.has('nvim-0.9') == 0 then H.get_tsnode = function() return vim.treesitter.get_node_at_cursor() end end
+
+H.hide_completion = function()
+  -- NOTE: `complete()` instead of emulating <C-y> has immediate effect
+  -- (without the need to `vim.schedule()`). The downsides are that `fn.mode(1)`
+  -- returns 'ic' (i.e. not "i" for clean Insert mode) and <C-n>/<C-p> act as if
+  -- there is completion active (thus not allowing them as custom mappings).
+  -- Appending ` | call feedkeys("\\<C-y>", "n")` removes that, but still would
+  -- require workarounds to work in edge cases.
+  -- NOTE: Use `silent` to not show "Pattern not found" messages. It also hides
+  -- '--INSERT--' temporarily when 'showmode' is active, but seems acceptable.
+  if vim.fn.mode() == 'i' then vim.cmd('silent noautocmd call complete(col("."), [])') end
+end
 
 -- TODO: Remove after compatibility with Neovim=0.9 is dropped
 H.islist = vim.fn.has('nvim-0.10') == 1 and vim.islist or vim.tbl_islist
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

Couple of notes:
- The 'mini.snippets' usage of `H.hide_completion()` is done before there is a guaranteed cursor move. Let's preserve this here, just in case. So just before `adjust_cursor()` (but probably still after `search()` as it is possible there'll be no movement) and before `pcall(nvim_win_set_cursor, ...)`.
- In tests it is more informative to have `eq(is_pumvisible(), true)` just before typing a key to make a jump. This both makes a test more robust and shows test's intention more clearly.

abeldekat:
>  ... before there is a guaranteed cursor move.

 I thought about that. However, is there a *guaranteed* cursor move? With treesitter, there is a `pcall`, and there won't be a movement if there is no target to jump to. Without treesitter, there is a check on `had_match`, after the jump.

> ...  it is more informative to have ...

Yes, that indeed shows intention more clearly. Thanks.

echasnovski:
> > ... before there is a guaranteed cursor move.
> 
> I thought about that. However, is there a _guaranteed_ cursor move? With treesitter, there is a `pcall`, and there won't be a movement if there is no target to jump to. Without treesitter, there is a check on `had_match`, after the jump.

The `pcall` in tree-sitter is mostly to guard from edge cases. Like a position after the last line which is valid for tree-sitter nodes but not for placing cursor. The cursor usually moves there.

And if there is no obvious difference or necessity of different order, making the as uniform decisions and code placement as possible usually helps in the long run.

abeldekat:
I added a new commit. Would it be an idea to refactor the following  lines that are duplicated in all new tests:

```lua
  type_keys('i<C-n>')
  eq(is_pumvisible(), true)
  type_keys('<Tab>')
  eq(is_pumvisible(), false)

```

Perhaps a `H.validate_jump_closes_pmenu(key)`

echasnovski:
> Would it be an idea to refactor the following lines that are duplicated in all new tests:

I found that the threshold for wrapping common test steps in the wrapper to be used across several cases is usually higher than in the "regular" code. I.e. in regular code I'd agree that these lines would be a candidate for a separate function, but in tests I'd leave them as is. Creating a wrapper that leaves "near" tested code (i.e. `local validate = function() ... end` inside test case itself) is usually fine.

The reasons are mostly more direct error traceback (without extra last entry leading inside the helper function) and not having to write complex logic with lots of parameters.

abeldekat:
Thanks for the opportunity to create this PR!

echasnovski:
I've adjusted a bit (d21238061803642be4d474e6bbe0e739d9d35dc1) and this should now be on latest `main`. Thanks again for the PR!

PR #1850, opened 2025-06-10
feat(ai): take matches with multiple nodes into account

TheLeoP:

This PR makes two things:

- `mini.ai` is now aware of matches with multiple matching nodes
- `mini.ai` now parses on it's own the buffer when there's a need to

Some comments:

> `mini.ai` is now aware of matches with multiple matching nodes

This means that for captures like https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/fa32a45fdbab9c9c3bda9ecec9b12dddb221b927/queries/lua/textobjects.scm#L12-L17 (present on the `main` branch of nvim-treesitter-textobjects) the behavior before this change is the following

```lua
--   | the cursor is here
--   v 
a(b, c, d)

-- the user configuration looks like
require("mini.ai").setup {
  custom_textobjects = {
    F = require("mini.ai").gen_spec.treesitter { a = "@call.outer", i = "@call.inner" },
  },
}

-- the user does `viF` and only the `d` is visually selected
```

the behavior after the change is the following

```lua
--   | the cursor is here
--   v 
a(b, c, d)

-- the user configuration looks the same

-- the user does `viF` and everything inside of the parenthesis is visually selected
```

> `mini.ai` now parses on it's own the buffer when there's a need to

This means calling `parser:parse(true)` instead of relying on other plugins (or the treesitter builtin highlighter of Neovim) to do it.

We probably don't need to parse the whole buffer, but this is also a configuration setting different enough from `n_lines` to not use it. Also, we may wan't to only parse the (possibly injected) `language_tree` under cursor instead of all of the langue trees (`language_tree:parse`, instead of `parser:parse`). We also probably may want to use the new async parsing capabilities of treesitter by providing a callback as a second argument for `parser:parse()`

I'm opening this PR mostly for discussion and to hear what others think about the topic. I also already had this code implemented in my own config (https://github.com/TheLeoP/nvim-config/blob/1a1d3fef733b59731faecfb854cb00def1f96c6a/lua/plugins/mini.lua#L43-L57), so I may as well see if it's welcomed upstream. I can add tests and update the documentation if the change is deemed acceptable.

### DIFF
--- a/lua/mini/ai.lua
+++ b/lua/mini/ai.lua
@@ -1566,12 +1566,30 @@ H.get_matched_ranges_builtin = function(captures)
   -- Compute ranges of matched captures
   local capture_is_requested = vim.tbl_map(function(c) return vim.tbl_contains(captures, '@' .. c) end, query.captures)
 
+  parser:parse(true)
+
   local res = {}
   for _, tree in ipairs(lang_tree:trees()) do
-    for capture_id, node, metadata in query:iter_captures(tree:root(), 0) do
-      if capture_is_requested[capture_id] then
-        metadata = (metadata or {})[capture_id] or {}
-        table.insert(res, vim.treesitter.get_range(node, buf_id, metadata))
+    for _, match, metadata in query:iter_matches(tree:root(), 0) do
+      for capture_id, nodes in pairs(match) do
+        if capture_is_requested[capture_id] then
+          metadata = (metadata or {})[capture_id] or {}
+          local first_range ---@type Range6
+          local last_range ---@type Range6
+
+          for _, node in ipairs(nodes) do
+            local range = vim.treesitter.get_range(node, buf_id, metadata)
+            local current_start_byte = first_range and first_range[3] or math.huge
+            local current_end_byte = last_range and last_range[6] or -1
+            if range[3] < current_start_byte then first_range = range end
+            if range[6] > current_end_byte then last_range = range end
+          end
+
+          table.insert(
+            res,
+            { first_range[1], first_range[2], first_range[3], last_range[4], last_range[5], last_range[6] }
+          )
+        end
       end
     end
   end
--- a/lua/mini/ai.lua
+++ b/lua/mini/ai.lua
@@ -1569,8 +1569,10 @@ H.get_matched_ranges_builtin = function(captures)
   lang_tree:parse(false)
 
   local res = {}
+  -- TODO: Remove after compatibility with Neovim=0.10 is dropped
+  local opts = { all = true }
   for _, tree in ipairs(lang_tree:trees()) do
-    for _, match, metadata in query:iter_matches(tree:root(), 0) do
+    for _, match, metadata in query:iter_matches(tree:root(), 0, nil, nil, opts) do
       for capture_id, nodes in pairs(match) do
         if capture_is_requested[capture_id] then
           metadata = (metadata or {})[capture_id] or {}
--- a/lua/mini/ai.lua
+++ b/lua/mini/ai.lua
@@ -1566,7 +1566,7 @@ H.get_matched_ranges_builtin = function(captures)
   -- Compute ranges of matched captures
   local capture_is_requested = vim.tbl_map(function(c) return vim.tbl_contains(captures, '@' .. c) end, query.captures)
 
-  parser:parse(true)
+  lang_tree:parse(false)
 
   local res = {}
   for _, tree in ipairs(lang_tree:trees()) do
--- a/lua/mini/surround.lua
+++ b/lua/mini/surround.lua
@@ -1545,7 +1545,7 @@ H.get_matched_range_pairs_builtin = function(captures)
   local query = vim.treesitter.query.get(lang, 'textobjects')
   if query == nil then H.error_treesitter('query') end
 
-  parser:parse(true)
+  lang_tree:parse(false)
 
   -- Compute matches for outer capture
   local outer_matches = {}
--- a/lua/mini/surround.lua
+++ b/lua/mini/surround.lua
@@ -1568,7 +1568,9 @@ end
 H.get_builtin_matches = function(capture, root, query, buf_id)
   local res = {} ---@type Range6[]
 
-  for _, match, metadata in query:iter_matches(root, 0) do
+  -- TODO: Remove after compatibility with Neovim=0.10 is dropped
+  local opts = { all = true }
+  for _, match, metadata in query:iter_matches(root, 0, nil, nil, opts) do
     for capture_id, nodes in pairs(match) do
       if query.captures[capture_id] == capture then
         metadata = (metadata or {})[capture_id] or {}
--- a/lua/mini/surround.lua
+++ b/lua/mini/surround.lua
@@ -1515,12 +1515,16 @@ H.get_matched_range_pairs_plugin = function(captures)
   outer_matches = vim.tbl_filter(function(x) return x.node.tree ~= nil end, outer_matches)
 
   -- Pick inner range as the biggest range for node matching inner query. This
-  -- is needed because query output is not quaranteed to come in order, so just
+  -- is needed because query output is not guaranteed to come in order, so just
   -- picking first one is not enough.
   return vim.tbl_map(function(m)
     local inner_matches = ts_queries.get_capture_matches(0, captures.inner, 'textobjects', m.node, nil)
+    local inner_ranges = vim.tbl_map(
+      function(match) return vim.treesitter.get_range(match.node, buf_id, match.metadata) end,
+      inner_matches
+    )
     local outer = vim.treesitter.get_range(m.node, buf_id, m.metadata)
-    local inner = H.get_biggest_range(buf_id, inner_matches)
+    local inner = H.get_biggest_range(inner_ranges)
     return { outer = outer, inner = inner }
   end, outer_matches)
 end
@@ -1541,35 +1545,58 @@ H.get_matched_range_pairs_builtin = function(captures)
   local query = vim.treesitter.query.get(lang, 'textobjects')
   if query == nil then H.error_treesitter('query') end
 
+  parser:parse(true)
+
   -- Compute matches for outer capture
   local outer_matches = {}
   for _, tree in ipairs(lang_tree:trees()) do
-    vim.list_extend(outer_matches, H.get_builtin_matches(captures.outer:sub(2), tree:root(), query))
+    vim.list_extend(outer_matches, H.get_builtin_matches(captures.outer:sub(2), tree:root(), query, buf_id))
   end
 
   -- Pick inner range as the biggest range for node matching inner query
-  return vim.tbl_map(function(m)
-    local inner_matches = H.get_builtin_matches(captures.inner:sub(2), m.node, query)
-    local outer = vim.treesitter.get_range(m.node, buf_id, m.metadata)
-    local inner = H.get_biggest_range(buf_id, inner_matches)
+  return vim.tbl_map(function(outer)
+    local node = lang_tree:named_node_for_range(
+      { outer[1], outer[2], outer[4], outer[5] },
+      { ignore_injections = false }
+    )
+    local inner_matches = H.get_builtin_matches(captures.inner:sub(2), node, query, buf_id)
+    local inner = H.get_biggest_range(inner_matches)
     return { outer = outer, inner = inner }
   end, outer_matches)
 end
 
-H.get_builtin_matches = function(capture, root, query)
-  local res = {}
-  for capture_id, node, metadata in query:iter_captures(root, 0) do
-    if query.captures[capture_id] == capture then
-      table.insert(res, { node = node, metadata = (metadata or {})[capture_id] or {} })
+H.get_builtin_matches = function(capture, root, query, buf_id)
+  local res = {} ---@type Range6[]
+
+  for _, match, metadata in query:iter_matches(root, 0) do
+    for capture_id, nodes in pairs(match) do
+      if query.captures[capture_id] == capture then
+        metadata = (metadata or {})[capture_id] or {}
+        local first_range ---@type Range6
+        local last_range ---@type Range6
+
+        for _, node in ipairs(nodes) do
+          local range = vim.treesitter.get_range(node, buf_id, metadata)
+          local current_start_byte = first_range and first_range[3] or math.huge
+          local current_end_byte = last_range and last_range[6] or -1
+          if range[3] < current_start_byte then first_range = range end
+          if range[6] > current_end_byte then last_range = range end
+        end
+
+        table.insert(
+          res,
+          { first_range[1], first_range[2], first_range[3], last_range[4], last_range[5], last_range[6] }
+        )
+      end
     end
   end
+
   return res
 end
 
-H.get_biggest_range = function(buf_id, match_arr)
+H.get_biggest_range = function(ranges)
   local best_range, best_byte_count = nil, -math.huge
-  for _, match in ipairs(match_arr) do
-    local range = vim.treesitter.get_range(match.node, buf_id, match.metadata)
+  for _, range in ipairs(ranges) do
     local byte_count = range[6] - range[3] + 1
     if best_byte_count < byte_count then
       best_range, best_byte_count = range, byte_count


echasnovski:
Thanks for the PR! Making 'mini.ai' work better with `main` branch of both 'nvim-treesitter' and 'nvim-treesitter-textobjects' is certainly welcome.

I am afraid I have too many questions:
- Is this approach of working with queries *explicitly* documented somewhere? I.e. that several captures within one match define one region instead of separate?

    If yes, then:
    - Does it work only on `main` 'nvim-treesitter'/'nvim-treesitter-textobjects' branch or on `master` branch also?
    - Is the order of ranges inside metadata guaranteed? I.e. that start of first range and end of last range defines the target range? Or maybe it should be "the earliest range start" and "the latest range end"?

- What earliest version of Neovim does this work on? Both `iter_matches` solution and `parser:parse(true)`? Currently 'mini.nvim' supports Neovim>=0.9, but for tree-sitter related stuff I am willing to unconditionally relax this to Neovim>=0.10. If this is only Neovim>=0.11, then there needs to be separate code paths, unfortunately.

---

On a side note, this type of complexity is the final drop for me to actually move away from mocking tree-sitter in tests for both 'mini.ai' and 'mini.surround'. But that is not the topic of this PR.

TheLeoP:
> I am afraid I have too many questions:

Don't worry, I love questions c: .

> Is this approach of working with queries explicitly documented somewhere? 

I don't think it is :/ . All the information I know regarding the topic is:

- https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/612 moved `nvim-treesitter-textobjects` from using custom directives like `make-range` to use quantified treesitter captures on their `main` branch
- `nvim-treesitter` also seems to use them for folds and indents on their `main` branch (i.e. https://github.com/nvim-treesitter/nvim-treesitter/blob/42fc28ba918343ebfd5565147a42a26580579482/queries/c/folds.scm#L19).
- The internal implementation of `foldexpr` using treesitter in Neovim core uses this same logic (https://github.com/neovim/neovim/blob/6a71239cd5ea2a8e993bddd8e7765d1afe3e79e4/runtime/lua/vim/treesitter/_fold.lua#L111-L120)

So, I guess is more an interpretation of how to use quantified captures rather than an officially documented feature (?.

> Does it work only on main 'nvim-treesitter'/'nvim-treesitter-textobjects' branch or on master branch also?

It only works on the `main` branch. The `master` branch still uses `make-range` in captures.

> Is the order of ranges inside metadata guaranteed? I.e. that start of first range and end of last range defines the target range? Or maybe it should be "the earliest range start" and "the latest range end"?

The internal Neovim core implementation of `foldexpr` using treesitter also assumes that nodes are ordered by range (https://github.com/neovim/neovim/blob/6a71239cd5ea2a8e993bddd8e7765d1afe3e79e4/runtime/lua/vim/treesitter/_fold.lua#L111-L120), but I couldn't find this documented anywhere.

I also found that the PR that added that change (https://github.com/neovim/neovim/pull/29344) mentions this not being documented anywhere, but a test was added for it. So, I guess is more of an internal behavior instead of an officially guaranteed one.

> What earliest version of Neovim does this work on?

Good question, I forgot that `mini.nvim` has a strict requirement of backwards compatibility, sorry for that.

> iter_matches

The behaviour changed from returning only the last matching node to returning all the matching nodes in Neovim `0.10`, I think. `:h news-0.10` mentions

```markdown
  • |Query:iter_matches()|, |vim.treesitter.query.add_predicate()|, and
    |vim.treesitter.query.add_directive()| accept a new `all` option which
    ensures that all matching nodes are returned as a table. The default option
    `all=false` returns only a single node, breaking captures with quantifiers
    like `(comment)+ @comment`; it is only provided for backward compatibility
    and will be removed after Nvim 0.10.
```

> parser:parse(true)

Since `0.11`, using `vim.treesitter.get_parser()` no longer parses the tree implicitly. From `:h news-0.11`

```markdown
• |vim.treesitter.get_parser()| and |vim.treesitter.start()| no longer parse the
  tree before returning. Scripts must call |LanguageTree:parse()| explicitly. >lua
    local p = vim.treesitter.get_parser(0, 'c')
    p:parse()
```

. But, if the tree is already parsed, calling `parser:parse` is a noop, so it shouldn't be a problem . async parsing (that is not being used in this PR, but I mentioned earlier as a possible improvement) was also added in `0.11`.

The `true` argument (or a range argument) to ensure parsing injected languages was added in the commit https://github.com/neovim/neovim/commit/2ca076e45fb3f1c08f6a1a374834df0701b8d778 , which is 2 years old and seems to be first included in Neovim `0.10`.

> If this is only Neovim>=0.11, then there needs to be separate code paths, unfortunately

Understandable. There's no problem on my side for this. What worries me the most is that treating quantified treesitter captures as a range is not officially documented anywhere.

> On a side note, this type of complexity is the final drop for me to actually move away from mocking tree-sitter in tests for both 'mini.ai' and 'mini.surround'. But that is not the topic of this PR.

I'm curious, what would you prefer doing instead of mocking treesitter? I must admit that it took me a non trivial amount of time to get the mocked tests working as expected, lol. So, understandable.

echasnovski:
That's a great answer! As expected from the person who loves questions :)

Yes, not having it (the query construction rule and assumption about `nodes` order in matches during iteration) documented is a problem. The query one is the main issue, because the `nodes` one is probably worth explicitly finding the first and last one anyways. I'll have to ask people responsible for queries about the behavior.

> > Does it work only on main 'nvim-treesitter'/'nvim-treesitter-textobjects' branch or on master branch also?
> 
> It only works on the `main` branch. The `master` branch still uses `make-range` in captures.

I mostly meant to ask if it doesn't break actually working behavior of `master`. The `make-range` in captures is a known "can not support" case.

> > parser:parse(true)
> 
> Since `0.11`, using `vim.treesitter.get_parser()` no longer parses the tree implicitly.

Yeah, I know about that. I only wanted to know if extra call to `:parse()` works on earlier versions. It looks to be present on Neovim=0.9, but the `:parse(true)` is supported only on Neovim>=0.10. Not a big deal to adjust.

> > On a side note, this type of complexity is the final drop for me to actually move away from mocking tree-sitter in tests for both 'mini.ai' and 'mini.surround'. But that is not the topic of this PR.
> 
> I'm curious, what would you prefer doing instead of mocking treesitter? I must admit that it took me a non trivial amount of time to get the mocked tests working as expected, lol. So, understandable.

To use built-in Lua and Vimscript parsers plus custom test queries, probably. Something like in ['test_snippets'](https://github.com/echasnovski/mini.nvim/blob/35e1767f4cd7dde51256eabae7349a5283a43cba/tests/test_snippets.lua#L1288-L1298) or [tests for `MiniExtra.pickers.treesitter()`](https://github.com/echasnovski/mini.nvim/blob/35e1767f4cd7dde51256eabae7349a5283a43cba/tests/test_extra.lua#L3587-L3594).

Let's wait a bit until I'll rework tree-sitter testing across 'mini.nvim' (will try to finish this week) and then get back to this. Does this sound good?

TheLeoP:
> Let's wait a bit until I'll rework tree-sitter testing across 'mini.nvim' (will try to finish this week) and then get back to this. Does this sound good?

Sounds good. Also, as mentioned in chat, I'll also wait until this feature is more stable/documented. Thanks for your time

echasnovski:
> Also, as mentioned in chat, I'll also wait until this feature is more stable/documented. Thanks for your time

Nah, I think it is okay to use now. Just want to err on side of caution and be at least a bit more sure that this won't break on `master`. Which more direct tests should help with.

echasnovski:
@TheLeoP, I've finally found enough determination to touch tree-sitter testing :sweat_smile: Both 'mini.ai' and 'mini.surround' now use real tree-sitter parser and `vim.treesitter` (which allowed to find issues on Nightly, by the way).

If you still have time and aspiration to have this functionality (which is *very* welcome), it requires conflict resolution and different testing. I'd also appreciate if there is a similar change to 'mini.surround' and its `gen_spec.input.treesitter()`, as it should have as much in common as `gen_spec.treesitter()` from 'mini.ai'.

If it is a "no" to any of these, I'd completely understand and will do this myself (hopefully, again, by the end of this week).

TheLeoP:
>  I'd also appreciate if there is a similar change to 'mini.surround' and its gen_spec.input.treesitter(), as it should have as much in common as gen_spec.treesitter() from 'mini.ai'.

I'd love to take a look and see what I can do

TheLeoP:
Using `parser:parse(true)` would parse the whole buffer, including injections, if there's a need to every time it's used. I'd be a noop if the buffer is already parsed. Would this be a concern regarding performance? Treesitter parsing that includes injections can be slow for big files.

An alternative would be to parse a range around the cursor instead, we could use `n_lines` for both `mini.ai` and `mini.surround`, or a new option entirely.

echasnovski:
> Using `parser:parse(true)` would parse the whole buffer, including injections, if there's a need to every time it's used. I'd be a noop if the buffer is already parsed. Would this be a concern regarding performance? Treesitter parsing that includes injections can be slow for big files.

I am not particularly knowledgeable in this area. My guess would be that tree-sitter parsing is now (on Neovim>=0.11) done async after loading and debounced/throttled after every text change. If that is correct, then it should be fine to include force parsing *if* it is indeed a noop on already parsed file.

Nevertheless, it is probably a good idea to parse only what might be needed. I'd like to not reuse `n_lines` (as it was a somewhat bad early API decision). Instead parse only a single language tree at cursor, which might remove injections slowdown (I think).

---

But I'd ask you to focus this PR only on making both 'mini.ai' and 'mini.surround' account for matches with multiple nodes. Mostly because it is a higher priority (there are consistently not working use cases without workarounds) and to not entangle these changes together. I am also skeptical it will be easy to write a test for async parsing (which I'd like to have).

TheLeoP:
> I am not particularly knowledgeable in this area. My guess would be that tree-sitter parsing is now (on Neovim>=0.11) done async after loading and debounced/throttled after every text change. If that is correct, then it should be fine to include force parsing if it is indeed a noop on already parsed file.

treesitter parsing can be done async on Neovim>=0.11 by giving `parser:parse` a second argument with a callback. From `:h LanguageTree:parse()`

```vimdoc
    Parameters: ~
      • {range}     (`boolean|Range?`) Parse this range in the parser's
                    source. Set to `true` to run a complete parse of the
                    source (Note: Can be slow!) Set to `false|nil` to only
                    parse regions with empty ranges (typically only the root
                    tree without injections).
      • {on_parse}  (`fun(err?: string, trees?: table<integer, TSTree>)?`)
                    Function invoked when parsing completes. When provided and
                    `vim.g._ts_force_sync_parsing` is not set, parsing will
                    run asynchronously. The first argument to the function is
                    a string representing the error type, in case of a failure
                    (currently only possible for timeouts). The second
                    argument is the list of trees returned by the parse (upon
                    success), or `nil` if the parse timed out (determined by
                    'redrawtime').

                    If parsing was still able to finish synchronously (within
                    3ms), `parse()` returns the list of trees. Otherwise, it
                    returns `nil`.
```

Also, every plugin (including the builtin highlighting implementation) that uses treesitter needs to take care of parsing the buffer after if it has changes. From `:h treesitter-languagetree`

```vimdoc
Whenever you need to access the current syntax tree, parse the buffer: >lua
    local tree = parser:parse({ start_row, end_row })
<

This returns a table of immutable |treesitter-tree| objects representing the
current state of the buffer. When the plugin wants to access the state after a
(possible) edit it must call `parse()` again.
```

the same section also mentions

```vimdoc
If the buffer wasn't edited, the
same tree will be returned again without extra work. If the buffer was parsed
before, incremental parsing will be done of the changed parts.
```

> Nevertheless, it is probably a good idea to parse only what might be needed. I'd like to not reuse n_lines (as it was a somewhat bad early API decision). Instead parse only a single language tree at cursor, which might remove injections slowdown (I think).

Got it. `parser:parse(false)` should do exactly that.

---
 
> But I'd ask you to focus this PR only on making both 'mini.ai' and 'mini.surround' account for matches with multiple nodes. Mostly because it is a higher priority (there are consistently not working use cases without workarounds) and to not entangle these changes together.

Understood. The PR should be working right now. CI is failing in 0.9 and 0.10 because of treesitter related changes. Is that ok? Or should I handle those failures more gracefully? My first guess would be showing a friendlier error message stating that treesitter builtin support requires Neovim>=0.11.

I'm also used to typing all of my functions, so I added a couple of typings annotations to some private functions in `mini.surround`, which is what's causing the Document Generation CI check to fail. Would you prefer me to remove those type annotations? Or to ignore them? I haven't used `mini.doc` a lot, but I seem to recall there being a way to avoid generating documentation for specific functions.

> I am also skeptical it will be easy to write a test for async parsing (which I'd like to have).

I haven't used `mini.test`, but if it allows async tests via some kind of callback mechanism, this should be possible

echasnovski:
> Understood. The PR should be working right now. CI is failing in 0.9 and 0.10 because of treesitter related changes. Is that ok? Or should I handle those failures more gracefully? My first guess would be showing a friendlier error message stating that treesitter builtin support requires Neovim>=0.11.

If it were Neovim>=0.10 I'd be fine with hard requiring it for tree-sitter textobjects and surroundings, but only Neovim>=0.11 is a bit too much. So I'd opt for it to work on Neovim<0.11 in without errors but also possibly without extra features from this PR. And, of course, as concisely as reasonably possible.

> I'm also used to typing all of my functions, so I added a couple of typings annotations to some private functions in `mini.surround`, which is what's causing the Document Generation CI check to fail. Would you prefer me to remove those type annotations? Or to ignore them? I haven't used `mini.doc` a lot, but I seem to recall there being a way to avoid generating documentation for specific functions.

Yes, I'd prefer to remove type annotations for local functions. Those are okay only for the complicated local helpers without intuitive arguments (and which I tended to add early on, but not anymore).

The way to ignore type annotations in 'mini.doc' is to add `---@private` somewhere in the same annotation block.

> I haven't used `mini.test`, but if it allows async tests via some kind of callback mechanism, this should be possible

Of course, in a way. It tests a lot of async methods in 'mini.nvim' after all :) I tend to prefer writing such tests to use `sleep()` for a known amount (plus a bit more for robustness) and then checking the state.

TheLeoP:
> If it were Neovim>=0.10 I'd be fine with hard requiring it for tree-sitter textobjects and surroundings, but only Neovim>=0.11 is a bit too much. So I'd opt for it to work on Neovim<0.11 in without errors but also possibly without extra features from this PR. And, of course, as concisely as reasonably possible.

I just re-read  my previous comment and realized that 0.10 can indeed be supported with a simple change :man_facepalming: . Well, that's fixed now, CI is green for 0.10.

> Yes, I'd prefer to remove type annotations for local functions. 

Done.

echasnovski:
Thanks again for all the work! I've merged into temporary branch and will take it from here (main credit will still be yours). The actual merge to `main` will probably be tomorrow, though.

echasnovski:
@TheLeoP, I force pushed into [`backlog`](https://github.com/echasnovski/mini.nvim/tree/backlog) two commits that address quantified captures in both 'mini.ai' and 'mini.surround'. They also make it work on Neovim=0.9 (except quantified capture part) at a tolerable performance penalty of adjusting variable type.

The 'mini.surround' code also required a bit more careful design decisions. Using `named_node_for_range` didn't really work when I interactively tested (not exactly sure why, but the approach is not quite suitable regardless). So decided to opt for working directly with ranges: find all possible outer and inner ranges (each with a single `iter_matches` per tree) and then for every outer range find the biggest nested inner range. Seems to work better and might even be more performant (as it does less `iter_matches()` in exchange of filtering out more inner ranges manually).

I plan to look at them with fresh eyes tomorrow before merging. If you have time and aspiration, could you also take a look before then? Maybe something will stand out to you.

---

I also decided to at least postpone the force parsing. Maybe until there is a reported issue about it or maybe until some time later. It feels like this should be done universally across all 'mini.nvim' modules that use `vim.treesitter`. That is not trivial (if adding tests is also a concern) and has performance implications. But we'll see.

TheLeoP:
> If you have time and aspiration, could you also take a look before then? Maybe something will stand out to you.

Besides the lack of parsing (that you removed on purpose), there's nothing that stands out to me at first sight.

> I also decided to at least postpone the force parsing. Maybe until there is a reported issue about it or maybe until some time later. It feels like this should be done universally across all 'mini.nvim' modules that use vim.treesitter. That is not trivial (if adding tests is also a concern) and has performance implications. But we'll see.

Understandable. Let me know if there's anything I can help with.

echasnovski:
Thanks for the extra look!

This should now be part of the latest `main`.

PR #1866, opened 2025-06-19
fix(pick): prevent clamping of `config.width` if there is no border

dmun:

This PR fixes the issue of the Pick width not matching the width of the editor as shown in the clip below (see right side).

This is a pretty bandaid fix, but I wanted to create this PR first to ask whether you think it's worth it to handle all different border cases, like only having a right border. Or maybe even remove the `math.min` clamping entirely since neovim already handles values larger than the window size.

https://github.com/user-attachments/assets/96baa0fb-ade8-4c51-bd8e-e68458d1b1e3

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -2285,7 +2285,7 @@ H.picker_compute_win_config = function(win_config, is_for_open)
   if config.border == 'none' then config.border = { '', ' ', '', '', '', ' ', '', '' } end
   -- - Account for border
   config.height = math.max(math.min(config.height, max_height - 2), 1)
-  config.width = math.max(math.min(config.width, max_width - 2), 1)
+  config.width = math.max(math.min(config.width, max_width), 1)
 
   return config
 end


echasnovski:
Thanks for the PR!

> This PR fixes the issue of the Pick width not matching the width of the editor as shown in the clip below (see right side).

Could you please give an example 'mini.pick' config to reproduce the first example. ~~I am mostly curious about why the window is skewed to the left with respect to its horizontal borders.~~ Ah, it is not. It is just overlapping with statusline which is the same color. But I'd still would like to see the config, please.

dmun:
Heres my config:

```lua
require("mini.pick").setup({
  options = { use_cache = true },
  window = {
    config = function()
      return {
        border = "none",
        col = 0,
        row = vim.o.lines - 1,
        height = math.floor(vim.o.lines / 3),
        width = vim.o.columns,
        relative = "editor",
      }
    end,
    prompt_caret = "█",
    prompt_prefix = "> ",
  },
})
```

But I see that you've committed a fix thank you!

echasnovski:
> Heres my config:

Thanks. I remembered that this is the way how `border = 'none'` looks only after looking at code :facepalm:

And yes this should now be resolved on latest `main` in a more structured (and hopefully robust) way.

PR #1923, opened 2025-08-06
fix(completion): ensure doc is a string when normalizing item doc

wenjinnn:
Hi echasnovski.

The `doc` variable inside `H.normalize_item_doc` function could be a table in some case, like when `lsp_item.documentation= { kind = "markdown" }` , which it should always be a string, otherwise it will throw an error while calling `doc:find`.
fixed by adjust the `doc` variable assignment statement.

I encountered this problem when using mini.completion with [kulala.nvim](https://github.com/mistweaverco/kulala.nvim), this plugin include a kulala LSP:
Snipaste_2025-08-06_14-20-28

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -1896,7 +1896,7 @@ H.normalize_item_doc = function(lsp_item, fallback_info)
 
   -- Extract string content. Treat markdown and plain kinds the same.
   -- Show both `detail` and `documentation` if the first provides new info.
-  detail, doc = detail or '', (type(doc) == 'table' and doc.value or doc) or ''
+  detail, doc = detail or '', type(doc) == 'table' and doc.value or ''
   -- Wrap details in language's code block to (usually) improve highlighting
   -- This approach seems to work in 'hrsh7th/nvim-cmp'
   detail = (H.is_whitespace(detail) or doc:find(detail, 1, true) ~= nil) and '' or (H.wrap_in_codeblock(detail) .. '\n')


echasnovski:
Thanks for the PR!

By the quick looks of it, the current implementation is more safe, because:
- Supplying documentation as a table with only `kind` and no `value` does not follow the LSP specification. The value should be [`MarkupContent`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition) which requires both fields present.
- The `documentation` (if present) [may also be string directly](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem), which the change in the PR stops account for.

Could you please provide a minimal reproduction example so that I can see in more details on what is going on? My current understanding is that it is an issue with either LSP server or 'kulala.nvim'.

echasnovski:
The following diff looks like a more appropriate solution:

```diff
diff --git a/lua/mini/completion.lua b/lua/mini/completion.lua
index 64e3db61..d0287256 100644
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -1896,7 +1896,7 @@ H.normalize_item_doc = function(lsp_item, fallback_info)
 
   -- Extract string content. Treat markdown and plain kinds the same.
   -- Show both `detail` and `documentation` if the first provides new info.
-  detail, doc = detail or '', (type(doc) == 'table' and doc.value or doc) or ''
+  detail, doc = detail or '', type(doc) == 'table' and (doc.value or '') or doc
   -- Wrap details in language's code block to (usually) improve highlighting
   -- This approach seems to work in 'hrsh7th/nvim-cmp'
   detail = (H.is_whitespace(detail) or doc:find(detail, 1, true) ~= nil) and '' or (H.wrap_in_codeblock(detail) .. '\n')
```

Does this solve the issue? Although I am pretty sure that LSP server would benefit from fixing this on their side.

wenjinnn:
Thanks for your quick reply!

> Thanks for the PR!
> 
> By the quick looks of it, the current implementation is more safe, because:
> 
>     * Supplying documentation as a table with only `kind` and no `value` does not follow the LSP specification. The value should be [`MarkupContent`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition) which requires both fields present.
> 
>     * The `documentation` (if present) [may also be string directly](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem), which the change in the PR stops account for.

And thanks for the refer, you're right, I'll report this issue to the kulala.nvim team.

> 
> Could you please provide a minimal reproduction example so that I can see in more details on what is going on? My current understanding is that it is an issue with either LSP server or 'kulala.nvim'.

Here's the minimal `init.lua` and reproduction step:
```lua
-- Put this at the top of 'init.lua'
local path_package = vim.fn.getcwd() .. '/.nvim/site'
local mini_path = path_package .. '/pack/deps/start/mini.nvim'
if not vim.loop.fs_stat(mini_path) then
  local clone_cmd = {
    'git',
    'clone',
    '--filter=blob:none',
    -- Uncomment next line to use 'stable' branch
    -- '--branch', 'stable',
    'https://github.com/echasnovski/mini.nvim',
    mini_path,
  }
  vim.fn.system(clone_cmd)
  vim.cmd('packadd mini.nvim | helptags ALL')
end
require('mini.deps').setup({
  path = {
    package = path_package,
  },
})
require('mini.completion').setup()
local add = MiniDeps.add

add('mistweaverco/kulala.nvim')
require('kulala').setup()
```
1. save it to some temporary dir
2. switch to this dir
3. launch neovim with this `init.lua` and edit a new http file with `nvim -u init.lua test.http`
4. type something like `Au` to trigger mini.completion,  and then type `C-n` to see the error.

> The following diff looks like a more appropriate solution:
> 
> ```diff
> diff --git a/lua/mini/completion.lua b/lua/mini/completion.lua
> index 64e3db61..d0287256 100644
> --- a/lua/mini/completion.lua
> +++ b/lua/mini/completion.lua
> @@ -1896,7 +1896,7 @@ H.normalize_item_doc = function(lsp_item, fallback_info)
>  
>    -- Extract string content. Treat markdown and plain kinds the same.
>    -- Show both `detail` and `documentation` if the first provides new info.
> -  detail, doc = detail or '', (type(doc) == 'table' and doc.value or doc) or ''
> +  detail, doc = detail or '', type(doc) == 'table' and (doc.value or '') or doc
>    -- Wrap details in language's code block to (usually) improve highlighting
>    -- This approach seems to work in 'hrsh7th/nvim-cmp'
>    detail = (H.is_whitespace(detail) or doc:find(detail, 1, true) ~= nil) and '' or (H.wrap_in_codeblock(detail) .. '\n')
> ```
> 
> Does this solve the issue? Although I am pretty sure that LSP server would benefit from fixing this on their side.

This indeed solves the problem and also complies with the relevant LSP specifications. So are you planning to make these changes in mini.completion? I’m fine with whatever you decide, but I personally prefer making these changes. Some LSP implementations may not handle null and empty strings well, and  these changes many make the handling of `doc.value` more robust.

echasnovski:
> This indeed solves the problem and also complies with the relevant LSP specifications. So are you planning to make these changes in mini.completion? I’m fine with whatever you decide, but I personally prefer making these changes. Some LSP implementations may not handle null and empty strings well, and these changes many make the handling of `doc.value` more robust.

The general idea is that 'mini.completion' will not make *substantial* extra efforts to account for LSP servers that don't follow the specification. Otherwise it will become to convoluted. Specification exists to introduce a standard which simplifies implementations of both LSP server (no need to invent custom solutions and data structures) and LSP client (it can assume specific data structures in specific situations).

But since in this particular instance it was already accounted for (just not covering all the issues), I'll adjust it soon (probably tomorrow).

wenjinnn:
Good to know. I just close this then.

echasnovski:
This should now be fixed on latest `main`.

Issue #1928, opened 2025-08-07
Arduino completion item bug


D4-CTM:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [ ] I have updated 'mini.nvim' to latest version of the `main` branch

### Module(s)

mini.completion

### Neovim version

0.11.x

### Description

I was trying out the `arduino-language-server` as I wanted to test it in nvim, when I had the lsp working and the completion seemed to be fine I got hit by the next error message while typing:
```
E5108: Error executing lua .../share/nvim/lazy/mini.completion/lua/mini/completion.lua:1820: attempt to index local 'item' (a boolean value)                                                    
stack traceback:                                                                                                                                                                                
        .../share/nvim/lazy/mini.completion/lua/mini/completion.lua:1820: in function 'get_lsp_edit_range'                                                                                      
        .../share/nvim/lazy/mini.completion/lua/mini/completion.lua:1796: in function 'get_completion_start'                                                                                    
        .../share/nvim/lazy/mini.completion/lua/mini/completion.lua:450: in function <.../share/nvim/lazy/mini.completion/lua/mini/completion.lua:425>
```

This occurred while I was typing `std::` after placing the first colon but didn't happen for the second. As the `arduino-language-server` uses `clangd` from behind I was expecting, well, nothing to happen, but this is not exclusive to the `std::` it is something that happens when typing anything that is not a keyword nor a constant.

### Reproduction

 1.  Use `arduino-language-server` with `mini-completion`. Requires clangd and arduino-cli, this is the [arduino_language_server.txt](https://github.com/user-attachments/files/21653197/arduino_language_server.txt) I used (I couldn't upload as `.lua`).
> [!NOTE]
> Requires mason

2. Create a `.ino` sketch and try typing anything that could trigger the completion to popup or write random characters, it will also trigger this.

echasnovski:
Thanks for the issue!

I indeed was able to reproduce this (after significant amount of struggle to set up Arduino project). The culprit here is not 'mini.completion', but the LSP server.

[...]

PR #1929, opened 2025-08-07
fix(#1928)/arduino-completion

D4-CTM:
- [ ] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)

I created an issue before checking the bug myself. This PR fixes an issue when working with the `arduino-language-server` in which typing anything that isn't a keyword will throw an error message.

[Screencast From 2025-08-07 00-25-19.webm](https://github.com/user-attachments/assets/c9b72d59-5968-4e0c-a376-6ea52cc21695)

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -1815,7 +1815,9 @@ H.get_lsp_edit_range = function(response_data)
   -- Try using all items to find the first one with edit range
   local items = response_data.result.items or response_data.result
   for _, item in pairs(items) do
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Unfortunately, this is not really a good change since it introduces extra checks (which affect performance) that should not be needed in the first place.

This should now be resolved on latest `main` with more details described [here](https://github.com/echasnovski/mini.nvim/issues/1928#issuecomment-3163320727).

PR #1954, opened 2025-08-18
feat(misc): make `zoom()` return whether zoomed in or out

loichyan:
This PR makes `MiniMisc.zoom` return the zoomed state of current buffer: `true` if zoomed in and `false` if zoomed out. This is useful for conditionally applying additional configurations to the window zoomed in. This feature was suggested in <https://github.com/echasnovski/mini.nvim/discussions/1953#discussioncomment-14138773> and also provides a more robust way to check zoomed state compared to that suggested in <https://github.com/echasnovski/mini.nvim/issues/1911#issuecomment-3112985891>. Basic usage:

```lua
local my_zoom = function()
  -- Do nothing if zoomed out.
  if not MiniMisc.zoom() then return end
  vim.wo.winbar = 'my winbar'
  -- ...other configurations
end
vim.keymap.set('n', '<Leader>z', my_zoom, { desc = 'Zoom' })
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,12 @@ There are following change types:
 
 - Add `gen_spotter.vimpattern()` that can generate spotter based on Vimscript (not Lua) pattern.
 
+## mini.misc
+
+### Expand
+
+- Update `zoom()` to return whether current buffer is zoomed in.
+
 ## mini.pick
 
 ### Expand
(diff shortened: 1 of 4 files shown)


echasnovski:
Thanks for the PR!

Could you incorporate testing of returned value into existing tests, please? Like [here](https://github.com/echasnovski/mini.nvim/blob/5d938b3a078a7eeae23ba9f04f180066ab8bd9ef/tests/test_misc.lua#L907) (use `child.lua_get()` to get the returned value) and also later in the test for closing (which is weirdly not already tested there :thinking: ).

loichyan:
Sure! 👀 I have added checks to each call of `MiniMisc.zoom()`, as well as checks for the closing state (except for those closed manually with `:quit`).

echasnovski:
Thanks again for the PR! This should now be part of latest `main`.

I've made several changes (see c5f2b25bfffbfb793602fa091fca0074871d30fe):
- There should be no need to add extra testing for returned values in every `zoom()` call.
- There should be a space between `@param` entries and `@return` for nicer help generation.

PR #1956, opened 2025-08-19
fix(misc): add handling for a nil background

### LINKED ISSUE #1955: `setup_termbg_sync()` does nothing on transparent `Normal` background
### Contributing guidelines

### Module(s)

mini.misc

### Neovim version

0.11.x

### Description

When a transparent theme (Normal.bg is set to "none", which become nil), the function that sets the background ignoes it, this can cause transparent themes to retain the color of the last non-transparent theme. 

### Reproduction

    ```lua
    vim.fn.system(clone_cmd)

    vim.cmd('packadd mini.nvim')
    require('mini.deps').setup()

    ```

4. Replace this with description of interactive reproduction steps along with the behavior you observe.
   Feel free to include images/videos/etc, this helps a lot.

</details>

mintoya:
fixes #1955 , this pr just sets the normal.background to 0 if it is nil inside of setup_termbg_sync's internal sync function

### DIFF
--- a/lua/mini/misc.lua
+++ b/lua/mini/misc.lua
@@ -324,19 +324,22 @@ MiniMisc.setup_termbg_sync = function()
     had_proper_response = true
     pcall(vim.api.nvim_del_autocmd, track_au_id)
 
-    -- Set up sync
-    local sync = function()
-      local normal = vim.api.nvim_get_hl_by_name('Normal', true)
-      normal.background = normal.background or 0
-      -- NOTE: use `io.stdout` instead of `io.write` to ensure correct target
-      -- Otherwise after `io.output(file); file:close()` there is an error
-      io.stdout:write(string.format('\027]11;#%06x\007', normal.background))
-    end
-    vim.api.nvim_create_autocmd({ 'VimResume', 'ColorScheme' }, { group = augroup, callback = sync })
-
     -- Set up reset to the color returned from the very first call
     H.termbg_init = H.termbg_init or bg_init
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

How does it work for you? Could you, please, provide a demo of before this change and after. Alongside which terminal emulator you use (and ideally with the relevant to transparency config values).

At first glance, this doesn't look like an intended solution because it will set terminal background to #000000 (i.e. pure black) which might contradict the whole "transparent background" idea.

mintoya:
> At first glance, this doesn't look like an intended solution because it will set terminal background to #000000 (i.e. pure black) which might contradict the whole "transparent background" idea.

to my knowlage, most "transparent" themes set the background to "none", and the terminal will print its default background. because of this, if a transparent theme is used with a non-transparent terminal, the background will just be the color of the terminal. You're right that setting it to black will set the color of the terminal to black, but the transparency will still be in effect.
 I've tested this with kitty, alacritty( non-transparent background), and my main terminal emulator, ghostty. The only difference is that in kitty, it seems all backgrounds, even for the non-transparent themes, have the opacity effect applied onto them.
 
 in my last commit i moved the reset function above sync and call that when there is no background, which should fix the pure black background
 
 first result from setting the theme to catpuccin, then to the transparent theme 
2025-08-19-042735_hyprshot
same actions with the first "fix" 
2025-08-19-042719_hyprshot
same actions with my last commit
image

echasnovski:
Yes, doing `reset` seems to make more sense. I've merged into a temporary branch to polish this a bit.

**Edit**: Should now be part of latest `main`.

PR #1968, opened 2025-08-27
fix(deps): Git environment variables would break mini.deps

diego-velez:

As part of my workflow, I use a Git bare repository. For this, I have a Lua script for Neovim to automatically detect and set up environment variables so that it can work with other Git integration plugins. In particular, I define `vim.env.GIT_WORK_TREE` and `vim.env.GIT_DIR`.

When I run `DepsUpdate`, the update fails and the remote origin for the git repository that Neovim was set up to follow with the aforementioned environment variables changes to the last plugin I added with mini.deps.

This PR aims to fix this issue by explicitly not using any environment variables when running git from mini.deps.

### DIFF
--- a/lua/mini/deps.lua
+++ b/lua/mini/deps.lua
@@ -1489,7 +1489,10 @@ H.cli_run = function(jobs)
     -- Prepare data for `vim.loop.spawn`
     local executable, args = command[1], vim.list_slice(command, 2, #command)
     local process, stdout, stderr = nil, vim.loop.new_pipe(), vim.loop.new_pipe()
-    local spawn_opts = { args = args, cwd = cwd, stdio = { nil, stdout, stderr } }
+    local env = { GIT_WORK_TREE = cwd, GIT_DIR = vim.fs.joinpath(cwd, '.git') }
+    if not vim.uv.fs_stat(env.GIT_DIR) then env = {} end
+
+    local spawn_opts = { args = args, cwd = cwd, env = env, stdio = { nil, stdout, stderr } }
 
     local on_exit = function(code)
       -- Process only not already closing job


echasnovski:
Thanks for the PR!

Although sounds reasonable, I *think* there might be some environment variables related to `git clone` and or `git fetch` that is worth respecting.

So I have couple of questions:
- Does this change (not using any environment variables) indeed solve your particular issue? I.e. have you verified it locally?
- Can you take a look if there is a more targeted way to address your use case? For example, maybe setting only those variables to some explicit value? It should be enough to fix your use case, but in a way that works for regular usages.

diego-velez:
>     * Does this change (not using any environment variables) indeed solve your particular issue? I.e. have you verified it locally?

Yes, it does solve my issue, and I have tested it locally. `DepsUpdate` works again when applying this PR, and I also tested `DepsAdd`.

>     * Can you take a look if there is a more targeted way to address your use case? For example, maybe setting only those variables to some explicit value? It should be enough to fix your use case, but in a way that works for regular usages.

I tested setting those environment variables for Git, and it worked fine both for `DepsUpdate` and `DepsAdd`.

Let me know what you think.

echasnovski:
> I tested setting those environment variables for Git, and it worked fine both for `DepsUpdate` and `DepsAdd`.
> 
> Let me know what you think.

The local `env` variable was not actually used. Could you test locally after applying my suggestion?

diego-velez:
Yes, sorry. I tested both with the environment variables and without to make sure everything worked as it should, but I forgot to change it back to use the variables.

echasnovski:
> Yes, sorry. I tested both with the environment variables and without to make sure everything worked as it should, but I forgot to change it back to use the variables.

I see, thanks. Now, with knowledge that this solves your use case, I'll take (later) a closer look if this is okay for regular use cases. I'll also adjust tests myself, as they are quite finicky here :)

echasnovski:
@diego-velez, so after re-educating myself about `GIT_DIR` and `GIT_WORK_TREE` I was able to reproduce this issue pretty easily.

After experiments with `opts.env` in `vim.system()` and reading its source code, I think the cleanest way to deal with this is to unset problematic variables directly while explicitly supplying all other relevant variables. This *might* introduce a bit of an overhead with extra `vim.fn.environ()`, but it shouldn't be too bad.

This should now be fixed on latest `main` and I plan to push a similar fix to the Neovim's `vim.pack`. Thanks again for finding this and helping two projects :)

Issue #917, opened 2024-05-26
how to sponsor you?


looshch:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.nvim

### Description

`mini.nvim` is fantastic. First i was skeptical how one person can maintain a reasonable level of quality for so many sub-packages, but you proved yourself beyond my expectations. I want to support you, my Ukrainian brother. How can i do this?

echasnovski:
I sincerely thank you for the kind words and willingness for financial support!

Unfortunately, at the moment GitHub Sponsors is not available in Ukraine. As it seems to be the easiest way enable support, I did not look any further yet. Maybe I should.

------

> .... how one person can maintain a reasonable level of quality for so many sub-packages.

The answer is mostly in the name and [general principles](https://github.com/echasnovski/mini.nvim?tab=readme-ov-file#general-principles). Most modules are already finished in terms of feature set. So for them only maintenance for breaking changes on Nightly (like latest `gra`/`grn` built-in mappings) and possibly reimplementing features (like the latest refactor of 'mini.statusline') is left.

And writing new modules is usually fun for me, so there's that :)

looshch:
it’s a shame it’s not supported there! Can you please post your credit card or at least crypto wallet? I would send money to you directly, that’s not a bother at all

echasnovski:
> it’s a shame it’s not supported there! Can you please post your credit card or at least crypto wallet? I would send money to you directly, that’s not a bother at all

Thanks, but I'd rather not. Let's wait for some more centralized options. I'll let you know by closing this issue as "Resolved" :)

ngortheone:
I was just wondering about this same. Here are some additional options to consider:
- Upwork
- Patreon
- Paypal

I am willing to support you too! Mini is the best set of plugins for neovim

echasnovski:
> I was just wondering about this same. Here are some additional options to consider:
> 
>     * Upwork
> 
>     * Patreon
> 
>     * Paypal
> 
> 
> I am willing to support you too! Mini is the best set of plugins for neovim

Thank you **so** much for your enthusiasm!

[...]

PR #1990, opened 2025-09-07
feat(move): allow shifting char horizontally in linewise mode

Groctel:

Hi!

First of all: I love your plugin suite. Thank you very much for your work!

I would like to propose extending `mini-move` to allow shifting a single character left or right in linewise mode as a toggleable functionality. This is something you already documented as not implemented, so I know you already thought of if, but not if you decided to not support it by design. I hope not :-P

I've added a few `NOTE:` comments in the code to call your attention to some points. Please let me know what I need to change to make this ready to be merged. Also, since this is a small change, should I just amend this commit? What's your history tidiness policy?

Thanks!

### DIFF
--- a/lua/mini/move.lua
+++ b/lua/mini/move.lua
@@ -14,7 +14,8 @@
 ---       four directions (left, right, down, up).
 ---     - Special handling of linewise movement:
 ---         - Vertical movement gets reindented with |=|.
----         - Horizontal movement is improved indent/dedent with |>| / |<|.
+---         - Horizontal movement is either improved indent/dedent with |>| / |<|
+---           or single character shifting with left and right.
 ---         - Cursor moves along with selection.
 ---
 --- - Provides both mappings and Lua functions for motions. See
@@ -50,9 +51,7 @@
 ---     - Doesn't support vertical movement of charwise and blockwise selections.
 ---       While 'mini.move' does.
 ---     - Doesn't support horizontal movement of current line in favor of
----       horizontal movement of current character. While 'mini.move' supports
----       horizontal movement of current line and doesn't support such movement
----       of current character.
+---       horizontal movement of current character.
 ---     - Has extra functionality for certain moves (like move by half page).
 ---       While 'mini.move' does not (by design).
 --- - 'booperlv/nvim-gomove':
@@ -152,6 +151,9 @@ MiniMove.config = {
 
   -- Options which control moving behavior
   options = {
+    -- Shift the character under the cursor horizontally instead of indenting
+    -- in linewise mode
+    shift_horizontal_chars = false,
     -- Automatically reindent selection during linewise vertical move
     reindent_linewise = true,
   },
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Can you please expand on what the expected usage here? If I understand correctly, you'd like, for example, to be able to press `2<A-l>` and it will move a single character two columns to the right, correct? If yes, then it is expected to be done with `v2<A-l><Esc>`. For a single column movement the simple `xP` is enough (which I use regularly), although it has some drawbacks regarding dot-repeat.

Groctel:
Yes! The idea is to allow the plugin to do `v<A-l><Esc>` by just pressing `<A-l>` with any `v:count`, which is a bit more ergonomic and aligns with the behaviour I honestly expected from the plugin when I installed it, which I'm sure is the case of others.

I don't think this would help with dot-repeat, as it currently only repeats the last `P`. I see that `H.make_cmd_normal` disables the command from altering the yank register, so pressing `.` will actually put what was in `"0` before you pressed `<A-l>`, am I right? Still, for such a small change as is moving a character left or right, I have never reached for the dot to repeat the motionalthough I'm sure someone will consider this absolute heresy :-P

echasnovski:
> Yes! The idea is to allow the plugin to do `v<A-l><Esc>` by just pressing `<A-l>` with any `v:count`, which is a bit more ergonomic and aligns with the behaviour I honestly expected from the plugin when I installed it, which I'm sure is the case of others.

I see. I am afraid, this goes against the whole *linewise* movement. The charwise movement is still suggested to be done via visual mode. The reasoning is that I'd say the most common charwise movement of a single character is by a single column, in which case `xp` works good enough. For more complex cases, visual selection also provides better feedback about the current state of the move.

Closing as not planned.

Groctel:
I see. Thanks for the well-reasoned feedback :-D

I ended up using this mapping in my config to make it behave the way I want:

```lua
require("mini.move").setup({})

local function move_selection(direction)
    vim.fn.feedkeys("v", "nx")
    MiniMove.move_selection(direction)
    vim.fn.feedkeys("v", "nx")
end

vim.keymap.set("n", "<A-h>", function () move_selection("left")  end, {noremap = true})
vim.keymap.set("n", "<A-l>", function () move_selection("right") end, {noremap = true})
```

My goal with the PR was to offer the community a solution in case they had a need similar to mine. This is the second best solution, so if anyone comes looking for this, they can yank this snippet ;-) Again, thank you for your work! This wouldn't have been possible if you hadn't made your plugins so easy to integrate from the outside.

See you around in the interwebs!

echasnovski:
> I ended up using this mapping in my config to make it behave the way I want:

Yep, that works. Here are couple of extra suggestions:

```lua
-- Make a generator directly
local function make_move_char(direction)
  return function()
    -- Not really necessary, but using `vim.api` functions is usually
    -- preferred over `vim.fn`
    vim.api.nvim_feedkeys('v', 'nx', true)
    MiniMove.move_selection(direction)
    vim.api.nvim_feedkeys('v', 'nx', true)
  end
end

-- Drop the whole `{ noremap = true }` because its behavior is default for
-- `vim.keymap.set()`. And prefer to use `{ remap = true }` for inverse case
-- (make mappings which treats keys as if they are mapped).
vim.keymap.set('n', '<A-h>', make_move_char('left'))
vim.keymap.set('n', '<A-l>', make_move_char('right'))
```

Also, if this type of movement is all users want, then it can (mostly; there are literal edge cases) be implemented without 'mini.move':

```lua
vim.keymap.set('n', '<M-l>', '<Cmd>exe "normal! x".v:count1."lP"<CR>')
vim.keymap.set('n', '<M-h>', '<Cmd>exe "normal! x".v:count1."hP"<CR>')
```

Groctel:
Thanks for the `{ noremap = true }` tip! One last question, if you don't mind: You said that *using `vim.api` functions is usually preferred over `vim.fn`*. Is the reasoning for this documented anywhere? Once again, thank you for your kindness <3

echasnovski:
> You said that _using `vim.api` functions is usually preferred over `vim.fn`_. Is the reasoning for this documented anywhere?

Not to that extent, sorry. The best one is `:h lua-guide-api` which this:

> Through this, any possible interaction can be done through Lua without writing a complete new API from scratch. For this reason, functions are usually not duplicated between layers unless there is a significant benefit in functionality or performance (e.g., you can map Lua functions directly through |nvim_create_autocmd()| but not through |:autocmd|). In case there are multiple ways of achieving the same thing, this guide will only cover what is most convenient to use from Lua.

But generally the idea is that:
- `vim.fn` has to go through extra layer of Lua->Vimscript-Lua conversion, which is usually best avoided if not needed. See also `:h lua-vimscript`.
    If I am not mistaken (which I might here), this for example results into the limitation that some `vim.fn` functions can't be called under certain conditions (like inside `vim.uv` callbacks).
- The `vim.api` are specifically built by Neovim for Neovim, so usually is better.

Also, some functionality is only available through Vimscript functions (`vim.fn`) without (yet) a `vim.api` alternative (like working with folds, window state, virtual columns, etc.). So in this case `vim.fn` are necessary.

**But** this is a very, *very* loose advice. In not performance critical areas use whatever feels better. I often use `vim.fn.line('.')` to get cursor line instead of `vim.api.nvim_win_get_cursor(0)[1]`, for example.

Groctel:
Hey! Thank you again for your help with the last question. I wanted to respond earlier but... I forgor :skull: Do you have a place where I can send you a tip for your work and kindness? I'd like to buy you a coffe or two and finally close this thread :heart:

echasnovski:
> Do you have a place where I can send you a tip for your work and kindness? I'd like to buy you a coffe or two and finally close this thread

Warm thank you for the intent! Currently there is now way, I afraid. But it might be possible in the future.

The best way to keep track of it is to subscribe to #917: it will be updated after there is a method for financial support. It will be completely voluntary, though, of course.

PR #2004, opened 2025-09-12
feat(terminal): open floating terminal or run shell commands in a floating terminal

kvgamedev:

terminal.lua adds a user command called MiniTerm which opens a floating window when there are no arguments present, and when there is an argument (let's say lazygit), it opens that shell command in a floating terminal window.
It also adds the keybind ctrl-q to exit terminal mode

### DIFF
--- a/lua/mini/terminal.lua
+++ b/lua/mini/terminal.lua
@@ -43,7 +43,7 @@ MiniTerm.config = {
   win = {
     height = 0.8,
     width = 0.8,
-    border = 'rounded',
+    border = vim.o.winborder or 'rounded',
   },
 }
 


echasnovski:
Thanks for the PR!

Although 'mini.terminals' is indeed planned, I have own ideas to implement in it. Thank you for the effort, but I'll close this in favor of a future new module.

kvgamedev:
excited to see your implementation. Can you give me feedback on how the plugin was written, are there things I missed or did wrong?

echasnovski:
> Can you give me feedback on how the plugin was written, are there things I missed or did wrong?

From the general plugin developing point of view? I can't see anything particularly bad. Maybe something like:
- I wouldn't create a `<C-q>` global Terminal mode mapping. This feels out of place here.
- I'd probably not export the `state` table directly in favor of `get` function(s) that return copy of locally tracked state data.
- The `config.win` might benefit from being a generic floating window config without dedicated set of fields (which can only be beneficial if not included fields are "harmful" to left to configure to the user).

    It can also be a callable that returns a window config: it is a more robust way to define a config because it can react to terminal resizing. Which is also a good reminder: if there are persistent floating windows, there should be a `VimResized` autocommand (either one global or created temporarily while window is visible) that recomputes window config.

Hope this helps.

Issue #2006, opened 2025-09-15
[mini.ai] cannot use treesitter textobjects inside an injected language


datwaft:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/nvim-mini/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/nvim-mini/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [x] I have updated 'mini.nvim' to latest version of the `main` branch

### Module(s)

mini.ai

### Neovim version

0.12 (!at least latest Nightly build!)

I can also reproduce with v0.11.4.

### Description

`tree-sitter` textobjects don't work inside injected languages.

For example with the following markdown document:

````markdown
# Some document

This document contains some text:

```python
x = 10
```
````

And the following queries:

```
(fenced_code_block
  (code_fence_content) @cell.inner) @cell.outer
```

And the following text object:

```lua
local spec_treesitter = require("mini.ai").gen_spec.treesitter
require("mini.ai").setup({
	custom_textobjects = {
		c = spec_treesitter({ a = "@cell.outer", i = "@cell.inner" }),
	},
})
```

If your cursor is in ` ```python ` and you press `vic` the inside of the codeblock will be selected, but if your cursor is inside the codeblock then the following error will appear:

```
E5108: Lua: (mini.ai) Can not get query for buffer 1 and language "python".
stack traceback:
        [C]: in function 'error'
        ...s/datwaft/.local/share/nvim/lazy/mini.ai/lua/mini/ai.lua:2029: in function 'error'
        ...s/datwaft/.local/share/nvim/lazy/mini.ai/lua/mini/ai.lua:1612: in function 'error_treesitter'
        ...s/datwaft/.local/share/nvim/lazy/mini.ai/lua/mini/ai.lua:1569: in function 'range_querier'
        ...s/datwaft/.local/share/nvim/lazy/mini.ai/lua/mini/ai.lua:1006: in function 'spec'
        ...s/datwaft/.local/share/nvim/lazy/mini.ai/lua/mini/ai.lua:1363: in function 'get_textobject_spec'
        ...s/datwaft/.local/share/nvim/lazy/mini.ai/lua/mini/ai.lua:606: in function 'find_textobject'
        ...s/datwaft/.local/share/nvim/lazy/mini.ai/lua/mini/ai.lua:1061: in function 'select_textobject'
[...]

PR #2009, opened 2025-09-15
feat(ai): query all language_trees for textobjects

TheLeoP:

Addresses #2006 

I open this as a draft because there isn't a consensus on how this should be handled. This PR queries all of the languages in the current buffer, which could lead to unwanted behavior (as mentioned in https://github.com/nvim-mini/mini.nvim/issues/2006#issuecomment-3290841201).

To address a few of the topics mentioned in that comment:

> At the moment I don't think I like the idea because it introduces issues about how the textobject should be searched. Like if there is a non-covering textobject with injected language and covering textobject for parent language, which one should be chosen with cover_or_next search method. I'd probably expect the former, but I can see how the latter can be expected.

With this PR, the current heuristic for choosing between multiple regions is used (which may not be the desired behavior). There's no distinction between textobjects in injected and non-injected languages.

> My current idea is to allow to specify "valid" languages for textobject. Then they can be adjusted specifically for filetype to query only expected parser. This has both upsides and downsides, which mostly are related to expected workflow and use cases. I'll take a look.

This would be an easy change. We could skip the language if it's not in the allow list for a specific filetype in

https://github.com/theleop/mini.nvim/blob/d647be34c34d03b71d7196be497eb2dfd2ebdab6/lua/mini/ai.lua#L1564

But, what would the default behavior be? Allow all languages? Only allow the language under cursor to keep the current behavior?

### DIFF
--- a/lua/mini/ai.lua
+++ b/lua/mini/ai.lua
@@ -1559,20 +1559,19 @@ H.get_matched_ranges_builtin = function(captures)
   local has_parser, parser = pcall(vim.treesitter.get_parser, buf_id, nil, { error = false })
   if not has_parser or parser == nil then H.error_treesitter('parser') end
 
-  -- Get parser (LanguageTree) at cursor (important for injected languages)
-  local pos = vim.api.nvim_win_get_cursor(0)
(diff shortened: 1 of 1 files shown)


echasnovski:
Let's go with #2010 approach.

PR #2011, opened 2025-09-15
fix(files): allow leading // or \\ on windows for UNC paths

cottandr:
I noticed on windows when trying to browse a file share using a UNC path (//server/share or \\server\share) the leading slash was eaten as part of the file normalization. 

This PR modifies the windows normalization function on windows to allow two leading slashes, but otherwise remove any repeats.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2596,7 +2596,7 @@ end
 
 H.fs_normalize_path = function(path) return (path:gsub('/+', '/'):gsub('(.)/$', '%1')) end
 if H.is_windows then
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Looks reasonable. As I don't have easy access to a Windows machine, I'll double check: does this indeed solve the use case described here? I.e. did you specifically check with these changes?

cottandr:
@echasnovski Thanks for looking at this!

Yes, I've tested it accessing file shares on windows, it's my main OS.

I tried to browse files on a home assistant config share and couldn't - when I tried I got a ```'path' is not a valid path``` error which showed the path with a missing slash. After applying this change to a local copy of mini in Lazy.vim, it works fine. 

The other bits of windows file normalization don't seem to cause any issues.

echasnovski:
This should now be part of latest `main`. Thanks again for the change!

Issue #1997, opened 2025-09-10
Option to auto-select first item if that is the only option


yutanagano:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/nvim-mini/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/nvim-mini/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.pick, mini.extra

### Description

I love mini.pick and would like to use it as a wrapper around vim.lsp.buf.definition to jump to definition, or select from one of many options in the case that the lsp detects more than one definition.

I know that mini.extra already provides an LSP picker, but it currently will not automatically jump to definition even if there is one unambiguous one. I think it would be nice to have an option inside mini.pick (at pick.start for example) to specify whether "auto-select-only-option" is enabled, then mini.extra.pickers.lsp can depend on this to enable such behaviour for when querying for lsp definitions.

echasnovski:
Thanks foe the suggestion!

The similar use case was already discussed in #978 with the [following resolution](https://github.com/nvim-mini/mini.nvim/issues/978#issuecomment-2178081844). So I don't think it is a good idea to add direct support for this use case.

However, in case this might help, there is a way to add "choose item if it is single" behavior to any picker:

```lua
local autochoose_single = function()
  local items = MiniPick.get_picker_items()
  if #items ~= 1 then return end
  local choose = MiniPick.get_picker_opts().source.choose or MiniPick.default_choose
  vim.schedule(function() choose(items[1]) end)
  MiniPick.stop()
end

-- NOTE: This relies on items being set synchronously, i.e. before start.
-- To also work with async items setting (with a bit of performance penalty of
-- being called more often), use 'MiniPickMatch' pattern.
vim.api.nvim_create_autocmd('User', { pattern = 'MiniPickStart', callback = autochoose_single })
```

Issue #978, opened 2024-06-19
Feature: add option jump_to_single_result to lsp picker


abeldekat:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.pick, mini.extra

### Description

Recently I started using mini.pick in my [config](https://github.com/abeldekat/nvim_pde/blob/main/lua/ak/config/editor/mini_pick.lua). Thank you for this great plugin!

Regarding the `lsp` picker, I miss the ability to jump directly to a single result, bypassing the picker.
This is standard behavior in `telescope`, and can be configured in `fzf-lua` by supplying `jump_to_single_result` is true.

I frequently use `goto definition`. For now, `fzf-lua` provides the following `lsp` pickers in my config: "definition", "reference", "implementation", "type_definition".

Would it be possible to add this behavior to mini.pick?

echasnovski:
Thanks for the suggestion!

This is, of course, possible. However, I don't think this is a good idea. My reasoning is that starting a picker should *always* result into showing a picker (as stated in #809 with a bit more thoughts in [this comment](https://github.com/echasnovski/mini.nvim/issues/550#issuecomment-1792044864)).

Closing as not planned.

abeldekat:
Oops... I was just writing an apology for this duplicate issue...)

metafates:
I've stumbled upon the same problem and implemented this function that does just that =) Hope it helps

```lua
-- Open LSP picker for the given scope
---@param scope "declaration" | "definition" | "document_symbol" | "implementation" | "references" | "type_definition" | "workspace_symbol"
---@param autojump boolean? If there is only one result it will jump to it.
function M.picker(scope, autojump)
	---@return string
	local function get_symbol_query()
		return vim.fn.input("Symbol: ")
	end

	if not autojump then
		local opts = { scope = scope }

		if scope == "workspace_symbol" then
			opts.symbol_query = get_symbol_query()
		end

		require("mini.extra").pickers.lsp(opts)
		return
	end

	---@param opts vim.lsp.LocationOpts.OnList
	local function on_list(opts)
		vim.fn.setqflist({}, " ", opts)

		if #opts.items == 1 then
			vim.cmd.cfirst()
[...]

PR #2014, opened 2025-09-17
feat(extra): immediately call vim.lsp.buf[scope] if one result

kkanden:

if lsp call results in 1 entry, it immediately calls the appropriate `vim.lsp.buf` handler without displaying the picker. it seemed kind of annoying that if e.g. "go to definition" had only one result, you still had to press enter once more to go there. 

hopefully this is something that's okay to open a PR for. i'm open to any sort of feedback!

### DIFF
--- a/lua/mini/extra.lua
+++ b/lua/mini/extra.lua
@@ -1984,6 +1984,10 @@ H.lsp_make_on_list = function(source, opts)
   end
 
   return function(data)
+    if vim.tbl_count(data.items) == 1 then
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Sorry, but I am afraid this was already discussed several times (the latest one is #1997) with conclusion that I don't think it is a good idea (more details [here](https://github.com/nvim-mini/mini.nvim/issues/978#issuecomment-2178081844)).

PR #2021, opened 2025-09-21
feat(completion): add abbr_hlgroup for deprecated items

telebart:
Use the DiagnosticDeprecated hlgroup to differentiate between deprecated vs normal items in the completion menu.

Took "inspiration" from here https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/completion.lua#L320

before:
image

after:
image

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -1256,6 +1256,15 @@ H.lsp_completion_response_items_to_complete_items = function(items)
 
     local lsp_data = { item = item, item_id = i }
     lsp_data.needs_snippet_insert = needs_snippet_insert
+
+    local abbr_hlgroup = ""
+    if
+      item.deprecated
+      or vim.list_contains((item.tags or {}), vim.lsp.protocol.CompletionTag.Deprecated)
+    then
+      abbr_hlgroup = 'DiagnosticDeprecated'
+    end
+
     table.insert(res, {
       -- Show less for snippet items (usually less confusion), but preserve
       -- built-in filtering capabilities (as it uses `word` to filter).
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

The `abbr_hlgroup` functionality is only available in Neovim>=0.11, while 'mini.nvim' currently supports Neovim>=0.9. So this will require a bit more care on top of improving performance of the if-else check.

But the main reason of why this was not yet added is that I don't particularly like this functionality and find it distracting. Maybe it can be added behind a dedicated 'MiniCompletionXxx' highlight group, not sure.

For future, I'd highly appreciate if issue is created first to discuss this (as described in 'CONTRIBUTING.md').

telebart:
Thanks for taking a look at this and sorry for not reading the CONTRIBUTING.md properly.

 I don't really have a argument of having this feature especially if it seems too heavy for what it provides. I just found it being strange that this was missing when it is part of the default completion.
 
 I guess this can be closed and opened in a new PR if it ever comes relevant in the future.

echasnovski:
> I guess this can be closed and opened in a new PR if it ever comes relevant in the future.

Let's keep this open for a couple of days, maybe I'll change my mind. In that case I'll finish this myself.

echasnovski:
After thinking about it, this does look like a reasonable thing to support. I decided to go with more flexible support and [allow customizing `abbr_hlgroup` as inside `process_items()`](https://github.com/nvim-mini/mini.nvim/blob/256d35c9e24c9cf5f5da568e121e988c6ceb1e3f/doc/mini-completion.txt#L51-L53) (same as for `kind_hlgroup`). By default it [adds new `MiniCompletionDeprected` group](https://github.com/nvim-mini/mini.nvim/blob/256d35c9e24c9cf5f5da568e121e988c6ceb1e3f/doc/mini-completion.txt#L370) for items that are marked as deprecated.

Thanks again for pointing this out!

Issue #474, opened 2023-09-09
mini.ai: tag selection breaks when tag contains an underscore


abeldekat:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

mini.ai

### Description

When the tags in an xml file contain underscores, the selection is not correct.

### Neovim version

0.9.2

### Steps to reproduce

1. Open an xml file, containing the following line: `<first_name>First Name</first_name>`
2. Move the cursor to any position inside the `<first_name>` tag 
3. Press `vit`

### Expected behavior

The text between the pair of tags gets selected: `First Name`

### Actual behavior

Error message:
`(mini.ai) No textobject "it" found covering region within 500 lines and `search_method = 'cover_or_next'`.`

echasnovski:
Thanks for the issue!

Tag textobject has obvious limitations in 'mini.ai' due to it using Lua pattern matching by default. The current implementation [does recognize only alphanumeric characters in tag name](https://github.com/echasnovski/mini.nvim/blob/f158f1905286e023d5848ab0b335a1cf5b8c692a/lua/mini/ai.lua#L1082). You can modify in `setup()` call as [it is suggested in this comment](https://github.com/echasnovski/mini.nvim/issues/314#issuecomment-1529097699) (it will recognize alphanumeric and punctuation as tag name):

```lua
require('mini.ai').setup({
  custom_textobjects = {
    t = { '<([%p%w]-)%f[^<%w][^<>]->.-</%1>', '^<.->().*()</[^/]->$' }
  }
})
```

------

However, specifically for tags a more sustainable and robust approach would be to use tree-sitter parsing and textobjects based on queries from 'nvim-treesitter-textobject': create 'after/ftplugin/html.lua' file and define `vim.b.miniai_config` buffer local variable there:

```lua
local spec_treesitter = require('mini.ai').gen_spec.treesitter
vim.b.miniai_config = {
  custom_textobjects = {
    t = spec_treesitter({ a = '@function.outer', i = '@function.inner' }),
  },
}
```

It will be more accurate but won't work inside comments and strings, though.

abeldekat:
[...]

Issue #314, opened 2023-04-30
Wrong `at` and `it` for html tag with "-"


jqhr:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

ai, surround

### Description

For vue, svelte or other template language, textobj "at" or "it"  only work for common tag without "-". For example div
![图片](https://user-images.githubusercontent.com/23305067/235364482-16fe5b0b-b853-414a-9848-30f482c11a8f.png)
But for "el-dropdown-menu", it dones select  tag "el-dropdown-menu", but select the outer tag "template", because "template" has no "-"
![图片](https://user-images.githubusercontent.com/23305067/235364668-236d51eb-b372-4f65-ae8d-0b36779e58d2.png)

### Neovim version

0.9.0

### Steps to reproduce

use "vat" or "vit" on html tag with "-"

### Expected behavior

![图片](https://user-images.githubusercontent.com/23305067/235364744-f106bd88-45e1-49c6-9f58-40dcef14e326.png)

### Actual behavior

![图片](https://user-images.githubusercontent.com/23305067/235364755-1a6add9f-f2cb-4a82-aadb-0dc7a69486ec.png)

jqhr:
I have not install `nvim-treesitter`.
The nvim default textobj "at" or "it"  was correct. but "mini.ai" or "mini.surrount" was wrong

echasnovski:
Thanks for the issue!

Yes, tag implementation in 'mini.ai' relies on an appropriate Lua patterns to match a tag region. I would strongly suggest using tree-sitter to get more proper information about tag structure. Something like how it this mentioned in [this comment](https://github.com/echasnovski/mini.nvim/issues/110#issuecomment-1212277863).

However, you can still customize tag textobject to include punctuation in tag name by slightly modifying built-in one like this:

```lua
require('mini.ai').setup({
  custom_textobjects = {
    t = { '<([%p%w]-)%f[^<%w][^<>]->.-</%1>', '^<.->().*()</[^/]->$' }
  }
})
```

jqhr:
Thanks.
[...]

PR #2057, opened 2025-10-11
feat(ai): Allow . in tag names

saep:
Before this change, if you were to press `cit` while inside the `tag.with.dot`
tag somewhere inside the `properties` tag in the example below, it would delete 
all other tags and leave an empty `properties` tag.

Example:

``` xml
<properties>
  <a>tag is deleted</a>
  <tag.with.dot>I only want to change this</tag.with.dot>
</properties>
```

Before this change:

``` xml
<properties></properties>
```

Before this change:

``` xml
<properties>
  <a>tag is deleted</a>
  <tag.with.dot></tag.with.dot>
</properties>
```


echasnovski:
Thanks for the PR!

This discussed in several issues: #314, #474. They also contain more feature-rich implementations.

The current behavior is intentional, in part to limit the use of Lua pattern based "tag" textobject in cases where it can be set up via tree-sitter.

Closing as not planned.

PR #2060, opened 2025-10-13
feat(files): add MiniFilesBorderFocused hl

husa:
Introduce new `MiniFilesBorderFocused` highlight for the active MiniFiles window so it can be themed distinctly. 
`MiniFilesBorderModified` takes precedence over "Focused" when the buffer is dirty.

Here's an example with catppuccin theme and some custom hl definitions
Screenshot 2025-10-13 at 14 11 22

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,6 +182,7 @@ Here is a list of all highlight groups defined inside 'mini.nvim' modules. See d
 
 - 'mini.files':
     - `MiniFilesBorder`
+    - `MiniFilesBorderFocused`
     - `MiniFilesBorderModified`
     - `MiniFilesCursorLine`
     - `MiniFilesDirectory`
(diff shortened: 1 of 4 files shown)


echasnovski:
Thanks for the PR! This was already considered and I think the conclusion still holds: it feels too noisy (even if default highlight group is linked to regular border). There is already `MiniFilesTitleFocused`, plus a focused window is usually detectable by width and position.

Closing as not planned. Thank you for your efforts, though!

husa:
@echasnovski thanks for explanation.

For anyone else interested, I'm achieving the same effect with 
```lua
  callback = function(ev)
    local win_id = ev.data.win_id
    if win_id == vim.api.nvim_get_current_win() then
      vim.api.nvim_win_set_option(win_id, "winhighlight", "FloatBorder:MiniFilesBorderFocused")
    end
  end,
})
```

PR #2075, opened 2025-10-19
Add combined change+delete hunk highlighting and document the new highlight group

lvim-tech:
- This helps users discover and customize the new combined hunk highlighting feature in their configuration.
- Add documentation for the new `MiniDiffSignChangeDelete` highlight group, which is used to visualize lines that are both "change" and "delete" (combined hunk visualization).


echasnovski:
Thanks for the PR!

The current behavior is intentional. Mostly for performance, but also to limit complexity of highlighting and signs.

PR #2087, opened 2025-10-26
feat: add width_file_preview option

PeterCardenas:
Support configuring the width of the preview window specifically for files. I wanted the preview window for files to be larger, so I decided to add the `width_file_preview` option. However, setting this (or any other width option) to a large value can make the windows overlap each other, which didn't seem intended to me. Thus, I also added some calculations to reset the widths of the created windows to fit the vim width. This change can be separated out into its own PR if desired, but put them both here as I felt the context was helpful.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -711,6 +711,8 @@ MiniFiles.config = {
     width_nofocus = 15,
     -- Width of preview window
     width_preview = 25,
+    -- Width of preview window if it is for a file
+    width_file_preview = nil,
   },
 }
 --minidoc_afterlines_end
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

The "resize in proportion" is not something I think should be done in 'mini.files'. This breaks config which defines widths *exactly*. If the window doesn't fit, it should be not shown. If this is not happening right now after the file preview option, then something is wrong in the other part of the code.

Also, 'mini.files' is already quite big and the threshold for any new features/configs is "*very* low line count compared to their effect". I am still debating whether a separate file preview width fits this.

In the meantime I think I am going to close this. Thank you for your time and effort, though!

PeterCardenas:
> If this is not happening right now after the file preview option, then something is wrong in the other part of the code.

so the behavior right now is if only the focused window and the preview window exist, if the config is such that the there is not enough space for the preview window, the preview window keeps the size but stays under the focused window overlapping and hiding it. what do you believe the expected behavior should be here?

PeterCardenas:
on your note about contributions, would code cleanup be welcome like adding types (no functionality change)

echasnovski:
> > If this is not happening right now after the file preview option, then something is wrong in the other part of the code.
> 
> so the behavior right now is if only the focused window and the preview window exist, if the config is such that the there is not enough space for the preview window, the preview window keeps the size but stays under the focused window overlapping and hiding it. what do you believe the expected behavior should be here?

I can not reproduce this. The preview window is omitted completely if it doesn't fit the screen (thus only focused window remains). If you can provide reproduction steps for window overlapping, please create an issue.

> on your note about contributions, would code cleanup be welcome like adding types (no functionality change)

Type annotations are currently mostly used for 'mini.doc' documentation. Going beyond that is not (currently) planned. So more rigorous typing is not something I'd like to see in 'mini.nvim' (yet). There are some more thoughts in [this comment](https://github.com/nvim-mini/mini.nvim/pull/1110#issuecomment-2266660524).

PeterCardenas:
> I can not reproduce this. The preview window is omitted completely if it doesn't fit the screen (thus only focused window remains). If you can provide reproduction steps for window overlapping, please create an issue.

it appears you're right, I'm guessing somehow i introduced the bug with my change. well if you change your mind I can take a look further. otherwise, I'll just use my fork 🙏

PR #2106, opened 2025-11-03
feat(pick): add functionality to remove and modify glob patterns during grep_live

acidx27x:

grep_live picker has `<C-o>` custom mapping to add glob to the array. Propose to expand this type of functionality with `<C-S-O>` custom mapping to remove glob from the array and `<C-m>` custom mapping to modify glob inside the array.

Looks like this when press `<C-S-O>`. Modifying with `<C-m>` looks similar but after hit a number, it prompts current glob and allows to change and confirm like `<C-o>`. Uses common for `mini.nvim` functions `H.user_input` and `H.echo` for input and output.
image

### DIFF
--- a/doc/mini-pick.txt
+++ b/doc/mini-pick.txt
@@ -1138,7 +1138,8 @@ Pick from pattern matches with live feedback
 
 Perform pattern matching treating prompt as pattern. Gives live feedback on
 which matches are found. Use |MiniPick-actions-refine| to revert to regular
-matching. Use `<C-o>` to restrict search to files matching glob patterns.
+matching. Use `<C-o>` to restrict search to files matching glob patterns,
+`<C-S-O>` to remove glob patterns, `<C-m>` to modify glob patterns.
 Tries to use one of the CLI tools to create items (see |MiniPick-cli-tools|):
 `rg`, `git`. If none is present, error is thrown (for performance reasons).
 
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

I am afraid, this kind of functionality is a bit too much. Especially since it requires so many new code.

Closing as not planned. Thanks for your time and efforts, though!

acidx27x:
@echasnovski, thanks for quick reply!

> I am afraid, this kind of functionality is a bit too much.

I feel this is useful for big project with mixture of different languages, template files and nested subdirs. Like when i have 3 or more patterns it is more easy to be able to edit, than just to reset picker and write almost same from scratch again.

> Especially since it requires so many new code.

Half of new code is taken from other `mini.nvim` modules (`H.user_input` and `H.echo` functions). I could use regular `vim.fn.input` and `print` but thought that using already written function is more stable approach. The rest of new code is like: get user input, check it and either remove or modify it. IMHO, it does not look like overkill functionality.

echasnovski:
> Half of new code is taken from other `mini.nvim` modules (`H.user_input` and `H.echo` functions). I could use regular `vim.fn.input` and `print` but thought that using already written function is more stable approach. The rest of new code is like: get user input, check it and either remove or modify it. IMHO, it does not look like overkill functionality.

Around 60 new lines are added just for this functionality, which is already too much for the 'mini.pick' at its current state. And occupying two mappings is also a bit too much for my taste. The better initial approach might have been for `<C-o>` to be followed by an "action identifier" (like `a` for "add", `e` for "edit", `d` for "delete"). But it is too big of a change now and adds to many layers to my liking.

I am sorry, but currently I don't think this addition to 'mini.pick' follows the overall design of 'mini.nvim' (prioritize maintenance-features balance instead of adding more features).

PR #2161, opened 2025-12-09
fix(pick/error): `set_picker_items ` don't throw an `error` if `picker is inactive`

AgatZan:

## Why
Everywhere else in the entire module, `MiniPick.is_picker_active` is checked before the assertion section. For example:
```lua
MiniPick.set_picker_items_from_cli = function(command, opts)
  if not MiniPick.is_picker_active() then return end
  local is_valid_command = H.is_array_of(command, 'string') and #command >= 1
  if not is_valid_command then H.error('`command` should be an array of strings.') end
```

The alternative behavior is not described in the documentation for the method.

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -1698,8 +1698,8 @@ MiniPick.get_picker_query = function() return vim.deepcopy((H.pickers.active or 
 ---
 ---@seealso |MiniPick.get_picker_items()| and |MiniPick.get_picker_stritems()|
 MiniPick.set_picker_items = function(items, opts)
-  if not H.islist(items) then H.error('`items` should be an array.') end
   if not MiniPick.is_picker_active() then return end
+  if not H.islist(items) then H.error('`items` should be an array.') end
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Can you, please, elaborate, why this is a problem and needs to be changed? Do you have a specific use case where this causes issues?

AgatZan:
@echasnovski Thanks for your work!


> why this is a problem and needs to be changed


Because that behaviour is different from other functions.

> Do you have a specific use case where this causes issues?


I'm not. But, not hard to suggest that in form like that.
```vim
--" not MiniPick.is_picker_active
:lua MiniPick.set_picker_items("items")          --"     throws an error
:lua MiniPick.set_picker_items_from_cli("items") --" not throws an error
```
If it is not a sufficient reason for a change, then I will close the PR.

echasnovski:
> Because that behaviour is different from other functions.

This should now be addressed on latest `main`, but in the other direction: argument validation in all `set_picker_xxx()` functions is now done before early return in case there is no active picker. This seems like a more "safe" direction to go here.

Issue #2166, opened 2025-12-12
mini.pairs closing fails when there's ghost text


vitaly:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/nvim-mini/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/nvim-mini/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [x] I have updated 'mini.nvim' to latest version of the `main` branch

### Module(s)

mini.pairs

### Neovim version

0.12 (!at least latest Nightly build!)

### Description

When a ghost text (e.g. from copilot) appears between the cursor and the closing pair:

`"some string content <cursor>some ghost text from copilot"`

When I type a closing quote `"`, the current behaviour of just moving forward, instead of typing the closing quote.
In the presence of ghost text, it doesn't work.
It just moves the cursor forward inside the ghost text:

`"some string content s<cursor>ome ghost text from copilot"`

So typing `"` appears to be just moving cursor forward until it reaches the end of the ghost text, at which point it will correctly close the string. The result will be `"some string content "`.

I think the solution is, instead of ignoring the typed character, and moving cursor forward, need to actually type it, and then `<del>` the next character. I tried this locally, and it worked. I'm going to submit a PR soon, though not sure if it does all correctly.

### Reproduction

1. Create separate 'nvim-repro' config directory:
    - '~/.config/nvim-repro/' on Unix
    - '~/AppData/Local/nvim-repro/' on Windows

2. Inside 'nvim-repro' directory create a file named 'init.lua'.
   Populate it with the following content:

```lua
    -- Clone latest 'mini.nvim' (requires Git CLI installed)
vim.cmd('echo "Installing `mini.nvim`" | redraw')
local mini_path = vim.fn.stdpath("data") .. "/site/pack/deps/start/mini.nvim"
local clone_cmd = { "git", "clone", "--depth=1", "https://github.com/nvim-mini/mini.nvim", mini_path }
vim.fn.system(clone_cmd)
vim.cmd('echo "`mini.nvim` is installed" | redraw')

-- Make sure 'mini.nvim' is available
vim.cmd("packadd mini.nvim")
require("mini.deps").setup()
require("mini.pairs").setup()

-- we are going to be able to 'travel' inside ghost text.
vim.opt.virtualedit = "all"

-- this is just a crude way to simulate ghost text from copilot.
[...]

PR #2167, opened 2025-12-12
Fix: interaction with ghost text (#2166)

vitaly:
Ref: nvim-mini/mini.nvim#2166

This is my attempt at fixing  a specific problem. It did fix it, but I'm not an expert on mini.pairs, so hope it didn't break something else ;)

The problem:

When a ghost text (e.g. from copilot) appears between the cursor and the closing pair:

`"some string content <cursor>some ghost text from copilot"`

The current behaviour of just moving forward, instead of typing the closing quote character doesn't work. It just moves the cursor forward inside the ghost text:

`"some string content s<cursor>ome ghost text from copilot"`

The solution is to actually type the end quote, and delete the character ahead.

### DIFF
--- a/lua/mini/pairs.lua
+++ b/lua/mini/pairs.lua
@@ -311,7 +311,7 @@ end
 MiniPairs.close = function(pair, neigh_pattern)
   local close = H.get_close_char(pair)
   local move_right = not H.is_disabled() and H.neigh_match(neigh_pattern) and H.get_neigh('right') == close
-  return move_right and H.get_arrow_key('right', true) or close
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

Always editing text when in 99.99% of cases it is enough to only move cursor is not the best approach.
The only reasonable solution here is to temporarily disable 'virtualedit', for which there is a helper. I'll test this locally and do benchmarks to see if it is worth it.

Meanwhile, I'll close this PR as it is not the solution I'd like to see.

PR #2182, opened 2025-12-19
feat(files): add the possibility to change the working directory

jsecchiero:

i find it very useful to change the cwd from `mini.files` and feels natural to me. I would like to hear from you if is in line with your vision before doing optimizations and fixing the tests

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -146,6 +146,8 @@ Available built-in actions (see "Details" for more information): >
 
  | Action      | Keys | Description                                    |
  |-------------|------|------------------------------------------------|
+ | Change cwd  |  *   | Change current working directory               |
+ |-------------|------|------------------------------------------------|
  | Close       |  q   | Close explorer                                 |
  |-------------|------|------------------------------------------------|
  | Go in       |  l   | Expand entry (show directory or open file)     |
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

The "change working directory mapping" is already [present as an example](https://nvim-mini.org/mini.nvim/doc/mini-files.html#minifiles-examples-createmappingswhichusedatafromentryundercursor). I think it is more suitable that way.

However, somehow indicating the current working directory is interesting. Not sure about an extra sign, but a dedicated highlight group sounds interesting. I'll think about it.

But anyway, I think this PR is a bit too much, so I am going to close this in the meantime. Thank you for your time, though!

jsecchiero:
Thanks for looking into this! I followed your callback suggestion and added the `cwd` info to the lualine instead to give me at least a feedback of the change. Just to clarify, i didn't use `get_fs_entry` like in the example. Instead, i set it up so the `cwd` changes to whatever folder i’m currently browsing.

It’s more ergonomic for me e.g. to switch back the `cwd` to the project root, there is no need to browse to the parent folder, also most of the times there’s no 'extra cost' associated with "browse the directory" + "change cwd" because whenever i want to do this, i want to see the files inside anyway.

It might be less intuitive for some, but it feels better to me. I'll test it out more and see how it goes! Digression over

echasnovski:
> However, somehow indicating the current working directory is interesting. Not sure about an extra sign, but a dedicated highlight group sounds interesting. I'll think about it.

Adding a dedicated highlight group for current working directory will conflict with `MiniFilesTitleFocused`, i.e. only one can be shown. I think it will be confusing, so unfortunately it won't be added to 'mini.files' directly. Still should be possible to adjust via autocommands.

PR #2196, opened 2025-12-27
fix(starter): close buffer before eval to prevent mapping interferences

themarcel:

**Problem:** 
mini.starter buffer mappings interfere with opened files when using `evaluate_single = true` with query based selection.

**Solution:** 
Close starter buffer before executing action in `eval_current_item`.

**Changes:**
- Use `close()` before action execution
- Fix `H.make_query()` buffer ID and use it in the closing

**Testing:** 
With `evaluate_single = true`, press 0 really fast to select item and navigation now works correctly.

### DIFF
--- a/lua/mini/starter.lua
+++ b/lua/mini/starter.lua
@@ -894,9 +894,10 @@ MiniStarter.eval_current_item = function(buf_id)
   if not H.validate_starter_buf_id(buf_id, 'eval_current_item()') then return end
 
   -- Reset query before evaluation without query echo (avoids hit-enter-prompt)
-  H.make_query(vim.api.nvim_get_current_buf(), '', false)
+  H.make_query(buf_id, '', false)
 
   local data = H.buffer_data[buf_id]
+  MiniStarter.close(buf_id)
   H.eval_fun_or_string(data.items[data.current_item_id].action, true)
 end
 


echasnovski:
Thanks for the PR!

> With `evaluate_single = true`, press 0 really fast to select item and navigation now works correctly.
> ...
> mini.starter buffer mappings interfere with opened files when using `evaluate_single = true` with query based selection.

I do not really understand *how* it interferes. Although at glance changes look fairly reasonable, could you please start with providing a full description of the issue this might fix? Including the 'mini.starter' config and steps I'd have to do?

It is needed to write the test for this.

marcelmanz:
> I do not really understand _how_ it interferes. Although at glance changes look fairly reasonable, could you please start with providing a full description of the issue this might fix? Including the 'mini.starter' config and steps I'd have to do?
> 
> It is needed to write the test for this.

Hi! Yes completely understandable, I am still trying to find a way to reproduce the issue. It happens to me *sometimes*.

Also I thought that was the only fix needed but it seems to persist, so I will still need to do some more changes.

Is it ok to keep the PR as draft till I find the solution? Thanks!

echasnovski:
> Is it ok to keep the PR as draft till I find the solution? Thanks!

I think it is better to close this in favor of a future issue that contains a description of a problem and reproduction steps.
Beware that this might be an issue introduced not by 'mini.starter' itself but with combination of some other plugin (maybe the one that deals with buffer local mappings?).

PR #2203, opened 2026-01-01
fix(files): allow case-only renames on case-insensitive filesystems

oskarkook:
On case-insensitive filesystems like macOS APFS, renaming a file or directory from `Foo` to `foo` would fail with "Target path already exists" error, forcing users to perform a two-step rename (e.g., `Foo` → `temp` → `foo`).

Now, a check is done to determine if two paths point to the same file. The move operation allows case-only renames when the source and target are the same file, while still preventing accidental overwrites of different files.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2633,6 +2633,14 @@ H.fs_is_imaginary_path = function(path) return path:sub(-1) == '\000' end
 
 H.fs_is_present_path = function(path) return vim.loop.fs_stat(path) ~= nil and not H.fs_is_imaginary_path(path) end
 
+H.fs_is_same_file = function(path1, path2)
+  if path1 == path2 then return true end
+  local stat1 = vim.loop.fs_stat(path1)
+  local stat2 = vim.loop.fs_stat(path2)
+  if stat1 == nil or stat2 == nil then return false end
+  return stat1.ino == stat2.ino and stat1.dev == stat2.dev
+end
+
 H.fs_child_path = function(dir, name) return H.fs_normalize_path(string.format('%s/%s', dir, name)) end
 
 H.fs_full_path = function(path) return H.fs_normalize_path(vim.fn.fnamemodify(path, ':p')) end
@@ -2777,8 +2785,10 @@ H.fs_do.delete = function(from, to)
 end
 
 H.fs_do.move = function(from, to, skip_buf_rename)
-  -- Don't override existing path
-  if H.fs_is_present_path(to) then return H.warn_existing_path(from, 'move or rename') end
+  -- Don't override existing path (unless it's a case-only rename of the same file)
+  if H.fs_is_present_path(to) and not H.fs_is_same_file(from, to) then
+    return H.warn_existing_path(from, 'move or rename')
+  end
 
   -- Move while allowing to create directory
   vim.fn.mkdir(H.fs_get_parent(to), 'p')


echasnovski:
Thanks for the PR!

Although I *kind of* get the use case, the combination of factors doesn't quite make this change suitable for 'mini.files':
- It adds code complexity (reliance on rather cryptic `ino` and `dev` fields that are probably not as ubiquitous as `libuv` describes) *and* performance penalty (two extra `fs_stat` calls) for a case that I'd imagine is quite rare.
- It is still a bit odd for me that changing case in case-insensitive system is a necessary operation to take into account.

So for the time being I'd close this PR (and behavior) as "probably not planned" ("probably" - because it might happen that both described above factors will change with time). But thank you for your time to make this PR! It made me learn what those fields are: `dev` is "Device number" (presumably unique identifier of a device; like disk or partition) and `ino` is "Inode Number" (presumably unique identifier of a file inside file system device).

oskarkook:
I understand your points, but I'd still like to challenge them.

1. As I understand, the complexity emerges from the reliance on `ino` and `dev`. The usage of these fields can certainly be described in the commit body or in a code comment to make it easier to understand for future readers. I can also explore other options if you still believe it's too complex to accept.
2. Re: performance. I imagine this plugin is not used in an environment where a move operation is executed _a lot_. It's an occasional operation. Is performance truly a concern? If you're willing to accept _a fix_ (doesn't have to be this one), I'd be ready to explore other solutions if you could describe in more detail what the use case is where performance would be a concern.
3. This plugin supports renaming, but it simply doesn't work on my filesystem (the most common one used on Macs) in this use case. I've run into this often enough by now that I decided to open this PR. Maybe I'm particularly weird in that I have to do this type of rename :D, but I've experienced this _many times_ by now. I wouldn't invest my own time into this if I didn't have a use case where I felt the pain.

Up to you, but I'm ready to adjust the PR and work for a solution, because I really do care about fixing this (as said, I've ran into this issue many times by now).

oskarkook:
When looking for alternative solutions, I found that `fs_realpath` works as well to remove the reliance on `ino` and `dev`, but I realised that it would not work for symbolic links. In the process, I also realised that `fs_lstat` should be used in place of `fs_stat` (same issue with symbolic links). I updated my branch to use `fs_lstat` and I'll leave it up to you to decide if you think the concerns you pointed out are surmountable or not.

Thank you for your time in any case! (and still very happy with `mini.nvim` as a long-time user).

echasnovski:
After looking around, it does seem that even dedicated system file explorers might require two-stage changes. And having to use `fs_lstat` (which I have never needed before) is also not reassuring.

But I did get an idea of a more straightforward approach to possibly fixing this. As I do not have access to case insensitive system right now, could you please try if the following patch allows case-only renames?

```diff
diff --git a/lua/mini/files.lua b/lua/mini/files.lua
index ffbdad83..8b5d5eeb 100644
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2633,6 +2633,11 @@ H.fs_is_imaginary_path = function(path) return path:sub(-1) == '\000' end
 
 H.fs_is_present_path = function(path) return vim.loop.fs_stat(path) ~= nil and not H.fs_is_imaginary_path(path) end
 
+H.fs_is_case_change = function() return false end
+if vim.fn.has('fname_case') == 0 then
+  H.fs_is_case_change = function(from, to) return from ~= to and vim.fn.tolower(from) == vim.fn.tolower(to) end
+end
+
 H.fs_child_path = function(dir, name) return H.fs_normalize_path(string.format('%s/%s', dir, name)) end
 
 H.fs_full_path = function(path) return H.fs_normalize_path(vim.fn.fnamemodify(path, ':p')) end
@@ -2777,8 +2782,10 @@ H.fs_do.delete = function(from, to)
 end
 
 H.fs_do.move = function(from, to, skip_buf_rename)
-  -- Don't override existing path
-  if H.fs_is_present_path(to) then return H.warn_existing_path(from, 'move or rename') end
+  -- Don't override existing path, if not a case change (like on Windows+MacOS)
+  if H.fs_is_present_path(to) and not H.fs_is_case_change(from, to) then
+    return H.warn_existing_path(from, 'move or rename')
+  end
 
   -- Move while allowing to create directory
   vim.fn.mkdir(H.fs_get_parent(to), 'p')
```

oskarkook:
That indeed works! Much simpler. Thank you!

oskarkook:
I'll point out that the docs aren't super-reassuring on `fname_case`:
> Case in file names matters (for Darwin and MS-Windows this is not present).

If it just checks whether I'm on Darwin and not the filesystem itself, then it's not strictly correct. [Zed for example handles it correctly](https://github.com/zed-industries/zed/blob/e1a09e290c48fc02d07aaf2150856d77996414df/crates/fs/src/fs.rs#L1160-L1196) (although the solution is quite... interesting :D).

oskarkook:
Maybe it is indeed a hairy-enough issue to leave it unsolved. _Failing_ the rename is more safe than risking overwriting another file in case the user is renaming things on a case-sensitive filesystem on Darwin/Windows.

echasnovski:
> If it just checks whether I'm on Darwin and not the filesystem itself, then it's not strictly correct.

Yes, [indeed](https://github.com/neovim/neovim/blob/612cd99a00174fc8c12a439781e7108a6efa8bea/build.zig#L215) it [does](https://github.com/neovim/neovim/blob/612cd99a00174fc8c12a439781e7108a6efa8bea/CMakeLists.txt#L119-L122).

> Maybe it is indeed a hairy-enough issue to leave it unsolved. _Failing_ the rename is more safe than risking overwriting another file in case the user is renaming things on a case-sensitive filesystem on Darwin/Windows.

It is definitely safer. And as far as I could tell, even more capable system file explorers/managers don't handle this very well. Probably for a good reason.

PR #2208, opened 2026-01-04
feat(pick): add manpages picker

Dook97:
A manpages picker inspired by [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim)'s implementation of the same.

Here it is in action:

https://github.com/user-attachments/assets/5a49e5ee-816a-4a1b-aa1f-ae7adb372aa1

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -1485,6 +1485,78 @@ MiniPick.builtin.help = function(local_opts, opts)
   return MiniPick.start(opts)
 end
 
+--- Man pages picker
+---
+--- Pick from system manpages.
+---
+---@param local_opts __extra_pickers_local_opts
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

I don't think it is a good fit for 'mini.pick' itself, which is meant to contain only minimum amount of essential pickers. It *might* be a good idea for 'mini.extra', but I am a bit hesitant to increase its size even more at this point. It *might* happen, but need some time to weigh some pros and cons here.

Could you please create a [Show&Tell discussion](https://github.com/nvim-mini/mini.nvim/discussions/categories/show-and-tell) instead? This is a good example of a custom picker and we'll probably make it more concise. It will also probably be part of the (hopefully future) wiki.

PR #2219, opened 2026-01-11
feat(pick): add _with_focus_lock function to support text jump

aaronlifton:
Title: Add _with_focus_lock function to support text jump

Description:
This enables usage of leap, and probably flash too, to jump to results, making the picker more on par with snacks.picker/fzf-lua. This was a missing feature I had to patch in to make leaping work.

Here is an example action for mini.pick:
```lua
MiniPick.setup({
  mappings = {
    move_down = '<C-n>',
    move_start = '<C-g>',
    leap = {
      char = '<M-s>',
      func = function()
        local state = MiniPick.get_picker_state()
        if not state or not state.windows or not state.windows.main then return end
        local win_id = state.windows.main
        local run = function()
          require('leap').leap({
            target_windows = { win_id },
            action = function(target)
              local current_line = vim.api.nvim_win_get_cursor(win_id)[1]
              local delta = target.pos[1] - current_line
              if delta == 0 then return end
              local key = delta >= 0 and '<C-n>' or '<C-p>'
              for _ = 1, math.abs(delta) do
                vim.api.nvim_input(key)
              end
            end,
          })
        end
        MiniPick._with_focus_lock(run)
      end,
    },
  },
})

```

https://github.com/user-attachments/assets/0268a5c4-369b-49bf-bba1-323e2d12992b

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -2199,6 +2199,16 @@ H.picker_new = function(opts)
   return picker
 end
 
+MiniPick._with_focus_lock = function(fn)
+  if type(H.cache) ~= 'table' then return fn() end
+  local prev = H.cache.is_in_getcharstr
+  H.cache.is_in_getcharstr = true
+  local ok, result = pcall(fn)
+  H.cache.is_in_getcharstr = prev
+  if not ok then error(result) end
+  return result
+end
+
 H.picker_advance = function(picker)
   vim.schedule(function() vim.api.nvim_exec_autocmds('User', { pattern = 'MiniPickStart' }) end)
 


echasnovski:
Thanks for the PR!

About the proposed function: I am not keen on adding a functionality for a pretty narrow scope and which intentionally deviates from the overall 'mini.pick' custom key query process.

About the use case: the suggested approach (which in my opinion is at least as equally efficient) is to use ["Refine" action](https://nvim-mini.org/mini.nvim/doc/mini-pick.html#minipick-actions-built-in-refine) (`<C-Space>` by default) to narrow down items based on the different query. This is more composable (since multiple refines in a row are possible) and has less "context switching".

Closing as not planned.

abeldekat:
I find "hints" very useful when the most relevant results of the picker are listed first, or when the results are predictable.

This is possible without modifying mini.pick, see [this](https://github.com/echasnovski/mini.nvim/discussions/1109) discussion.

I have been [using](https://github.com/abeldekat/nvim_pde/blob/b710e97a341c8ca77f61f8208d288dc9fc188ec4/lua/ak/mini/pick.lua#L18) that [code](https://github.com/abeldekat/nvim_pde/blob/b710e97a341c8ca77f61f8208d288dc9fc188ec4/lua/akextra/pick_hinted.lua#L1) daily for more than a year now. Perhaps it's useful for you too.

PR #2227, opened 2026-01-17
mini.pick: allow cancel of grep

ldelossa:
Prior to this commit there was no graceful way to cancel the pattern
input of the grep picker.

Hitting `ctrl-c` created an error.
Any other natural form of cancellation returned an empty string, which
blocked the picker in large repos for quite some time.

Replace `vim.fn.input` with `vim.ui.input` which properly distinguishes
a cancellation of the prompt from an empty input string.

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -1346,14 +1346,21 @@ MiniPick.builtin.grep = function(local_opts, opts)
   local default_opts = { source = { name = string.format('Grep (%s%s)', tool, name_suffix), show = show } }
   opts = vim.tbl_deep_extend('force', default_opts, opts or {})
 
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

All 'mini.nvim' modules (well, except a single occasion) intentionally use `vim.fn.input` for consistent experience. Replacing it with `vim.ui.input` (or a more narrow alternative) is planned, but only if there is a way to force its execution to be synchronous. This is important in modules like 'mini.ai' and 'mini.surround'.

Until then, replacing with `vim.ui.input` is not planned.

---

As per respecting cancel in `MiniPick.builtin.grep` (which I'd appreciate if it first would have been an issue), this should now be resolved on latest `main`. Thanks for reporting this!

PR #2248, opened 2026-01-28
fix(operators): correctly handle newlines at start of textobject

### LINKED ISSUE #2247: [mini.operators] some operators have issues with newlines at start/end of textobject
### Contributing guidelines

### Module(s)

mini.operators

### Neovim version

0.11.x

### Description

Using both `mini.ai` (not required, but allows better control/knowledge of how the textobjects for the operators are selected) and `mini.operators` default configuration, given the buffer

```lua
foo(
  a,
  b,
  c
)
```

, and with the cursor above `a`, doing `yi)` and then `gri)` results in the buffer

```lua
foo
  a,
  b,
  c

)
```

while I would expect it to result in the same buffer. Similar results can be seen with the multiply operator. `gmi)` results in

```lua
foo(
  a,
  b,
  c(
  a,
  b,
  c
)
```

note the `(` after `c` that should not be part of the textobject.

This happens because in

https://github.com/nvim-mini/mini.nvim/blob/9b935c218ddba02e5dc75c94f90143bce1f7c646/lua/mini/operators.lua#L1167-L1171

`keys` ends up being something similar to `` `["_dv`] ``. Then, when Neovim moves the cursor to the `[` mark, since it isn't on visual mode and `virtualedit` default value is empty, instead of putting the cursor on the newline after `(`, the cursor is placed on the parenthesis itself.

This issue can be (partially) solved by temporarily changing `virtualedit` to `onemore` inside of 

https://github.com/nvim-mini/mini.nvim/blob/9b935c218ddba02e5dc75c94f90143bce1f7c646/lua/mini/operators.lua#L1149

(like it is already done on `mini.ai` for selecting a textobject). After that change, the first example no longer deletes the `(`, but it still duplicates the newline after `c`. So the end result is

```lua
foo(
  a,
  b,
  c

)
```

I've been look at it for a while, but I'm not sure as to why. Yanking (`yi)`) includes the trailing newline after `c` and `mini.ai` tries to move the cursor to `{4, 4}` when selecting the textobject. But, replacing `gri)`does not include the trailing newline after `c` when deleting the text. `vim.api.nvim_buf_get_mark(0, ']')` returns `{4, 2}` and `` `] `` moves the cursor up to `c`. So, my guess is that this second part of the issue may have something to do with Neovim internals (?

### Reproduction

    ```lua
    local mini_path = vim.fn.stdpath("data") .. "/site/pack/deps/start/mini.nvim"
    local clone_cmd = { "git", "clone", "--depth=1", "https://github.com/nvim-mini/mini.nvim", mini_path }
    vim.fn.system(clone_cmd)

    vim.cmd("packadd mini.nvim")
    require("mini.deps").setup()

    require("mini.ai").setup()
    require("mini.operators").setup()
    ```

4. Insert the following text in the buffer

```
foo(
  a,
  b,
  c
)
```

5. Put the cursor above `a`, `yi)` and `gri)`

The resulting buffer will be

```
foo
  a,
  b,
  c

)

```

TheLeoP:

Partially fixes #2247

### DIFF
--- a/lua/mini/operators.lua
+++ b/lua/mini/operators.lua
@@ -375,8 +375,13 @@ MiniOperators.exchange = function(mode)
     H.cache.exchange.step_two = H.exchange_set_region_extmark(mode, false)
     if H.cache.exchange.step_two == nil then return end
 
+    local cache_virtualedit = vim.o.virtualedit
+    vim.o.virtualedit = 'onemore'
+
     -- Do exchange
     H.exchange_do()
+
+    vim.o.virtualedit = cache_virtualedit
 
     -- Stop exchange
     H.exchange_stop()
@@ -420,6 +425,9 @@ MiniOperators.multiply = function(mode)
   if data == nil then return end
   local mark_from, mark_to, submode = data.mark_from, data.mark_to, data.submode
 
+  local cache_virtualedit = vim.o.virtualedit
+  vim.o.virtualedit = 'onemore'
+
   H.with_temp_context({ registers = { 'x', '"' } }, function()
     -- Yank to temporary "x" register
     local yank_data = { mark_from = mark_from, mark_to = mark_to, submode = submode, mode = mode, register = 'x' }
@@ -442,6 +450,8 @@ MiniOperators.multiply = function(mode)
     -- already is at first non-blank, while this moves to first column.
     if submode ~= 'V' then vim.cmd('normal! `[') end
   end)
+
+  vim.o.virtualedit = cache_virtualedit
 end
 
 --- Replace text with register
@@ -484,7 +494,12 @@ MiniOperators.replace = function(mode)
   data.register = register
   data.reindent_linewise = H.get_config().replace.reindent_linewise
 
+  local cache_virtualedit = vim.o.virtualedit
+  vim.o.virtualedit = 'onemore'
+
   H.replace_do(data)
+
+  vim.o.virtualedit = cache_virtualedit
 
   return ''
 end
--- a/tests/test_operators.lua
+++ b/tests/test_operators.lua
@@ -1205,6 +1205,17 @@ T['Exchange']['works with `make_mappings()`'] = function()
   validate_edit1d('aa bb', 0, { 'viwcx', 'w', 'viwcx' }, 'bb aa', 3)
 end
 
+T['Exchange']['works on visual selection with leading newline'] = function()
+  set_lines({ 'a(', '  b', ')', 'c{', '  d,', '}' })
+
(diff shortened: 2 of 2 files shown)


echasnovski:
If it is a change in `do_between_marks`, then I think it is reasonable to also add tests for all operators that use it. Namely, "multiply" and "exchange".

Also, I am a bit hesitant to yet merge this as it doesn't address both problems of #2247.

TheLeoP:
> If it is a change in do_between_marks, then I think it is reasonable to also add tests for all operators that use it. Namely, "multiply" and "exchange".

On it.

> Also, I am a bit hesitant to yet merge this as it doesn't address both problems of https://github.com/nvim-mini/mini.nvim/issues/2247.

Fair. I'll try to take a look at what may be causing the issue with the trailing newline tomorrow

TheLeoP:
I changed the commit message to include newlines at the `end` of textobject because I added a test for a different edge case on `MiniOperators.replace` that happened only on visual mode and only when replacing a region with a trailing newline. But, the operator still duplicates the trailing newline (it can't delete it because, as explained in https://github.com/nvim-mini/mini.nvim/issues/2247#issuecomment-3832318672 , custom operators can't see trailing newlines).

We would be accepting that this is a current limitation of Neovim and the behavior is expected.

echasnovski:
Could you please move cache-set-unset of 'virtualedit' at least inside `H.exchange_do` *or* explore if only moving it inside `H.replace_do` is enough . Preferably closer to commands when they are actually needed (i.e. not at the top and bottom).

Also, please add the same one-line comment above each 'virtualedit' caching describing why this is needed.

For tests, it looks like the behavior is Visual mode specific. So could you please move them after `'works in Visual mode'` cases for each operator and be more general like `'works with newline in region'`. For trailing newlines in `exchange` and `multiply` having a comment about why it doesn't work (like `Doesn't work due to Neovim limitations\nSee https://github.com/neovim/neovim/issues/37664`).

echasnovski:
I currently have time to finish nit-picking parts myself. Hope to merge into `main` soon-ish.

echasnovski:
The first version of this (with only `do_between_marks`) should now be part of latest `main`.

After looking more closely, having all "cache-set-unset" separate didn't gain much. The only case is for "replace" to act less wrong on trailing newline. The problem is that the resulting behavior is still wrong (replacing a region with itself should have produced the same text, not add another line).

So as the first version is much simpler and to the point, let's go with that one.

---

Thanks for all the deep research and PR to make 'mini.operators' better in this very narrow use case :)

PR #2257, opened 2026-02-03
docs(git): Add note explaining how to force a vertical split

briandipalma:

Some documentation that would have helped me, might help someone else.


echasnovski:
Thanks for the PR!

The example above the added line already shows how to open a vertical split and explains it. I think essentially duplicating an example is not a good idea here.

Issue #2201, opened 2025-12-31
`MiniPick.builtin.grep_live` is slow on very large code bases


mbfoss:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/nvim-mini/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/nvim-mini/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.pick

### Description

I am having issues with live grep on very large code bases.
The issue I believe is that rg is spawned for each key press (I may be wrong)
 if the result are in the hundreds of thousands. the plugin takes a lot of time to load the results into memory and the UI becomes non-responsive for few seconds.
This cycle repeats until the query string becomes longs enough and the result list get's shorter.

Unfortunately, I don't think rg have a switch to limit the total number of results.
What about implementing a throttle mechanism and a limit on the maximum number of results.

1 - throttle rg spawning while typing (for example: max 1 spawn per second)
2 - if rg takes longer than the throttle delay or if the result exceeds certain number of lines, then stop the process and show only the results so far and if the query changed meanwhile then start another rg.

mbfoss:
I can attempt a PR if this is something that makes sense to you.

echasnovski:
Thanks for the issue!

I can see that happening, but I'd like to ask for the reproduction. If the delay is couple of seconds for 100K+ matches, I'd consider this okay-ish.

Optimizing speed for 100K+ matches is not really a goal for 'mini.pick', as at that range it is responsiveness that matters: user should be able to type the query and each key will get registered and processed (eventually).

Although `rg` indeed spawns after every query change, the previous process should be terminated immediately. If I know that there are a lot of files to search through, I tend to type initial keys fast so that to reach the state when there are manageable number of matches.

> Unfortunately, I don't think rg have a switch to limit the total number of results.

Total - no. But there is `--max-count` which might be occasionally useful.

mbfoss:
I reproduced using the following steps

- git clone https://github.com/torvalds/linux.git
- NVIM_APPNAME=nvim-minimax nvim
- :Pick grep_live
- press down and don't release a keyboard key (to generate a sequence of about 10 or more)
- Neovim becomes unresponsive for about 1 to 3 seconds 

This is a Mac M1, so mileage may vary depending on the hardware.
It was a little worse on my work computer.

[...]

PR #2259, opened 2026-02-04
fix(pick): don't cancel setting file picker items on user input

somnam:

Typing matching input in the file picker before file list is loaded results in an empty picker. This can happen in folders with many files e.g. $HOME dir.

Fix: the CLI feed shouldn't abort because the user typed. Change the throttle call in `H.set_picker_items_from_feed` to not pass `H.querytick`.

### DIFF
--- a/lua/mini/pick.lua
+++ b/lua/mini/pick.lua
@@ -3406,7 +3406,7 @@ H.set_picker_items_from_feed = function(feed, pattern, opts)
   if not MiniPick.is_picker_active() then return end
 
   -- NOTE: For intended effect, relies on presence of the active coroutine
-  local poke_picker = H.poke_picker_throttle(H.querytick)
+  local poke_picker = H.poke_picker_throttle()
 
   -- Realign feed so that each one ends in a pattern
   for i = 2, #feed do


echasnovski:
Thanks for the PR!

> Typing matching input in the file picker before file list is loaded results in an empty picker. This can happen in folders with many files e.g. $HOME dir.

I can reproduce but only if typing not immediately after the picker is started. Probably indeed has to be after CLI has finished producing output but before 'mini.pick' finished processing it.

Removing querytick checking doesn't  feel like a good approach. It it there for a reason, to resolve #2201. I'll have to take a closer look.

somnam:
Purely anecdotal, I did a few live_grep tests as described [here](https://github.com/nvim-mini/mini.nvim/issues/2201#issuecomment-3702753498) with & without this change in $HOME and /usr and the experience was basically the same. This is probably hardware dependent anyway so it's hard to tell if this is anything reliable.

echasnovski:
I have a pretty good idea of how to adjust this properly, but need to do a couple of more tests.

echasnovski:
This should now be resolved on latest `main`.

The trick was to reuse the `opts.set_items_opts.querytick`. The `builtin.files` doesn't set this (as computing CLI items should not stop after query changes, only if the picker itself has changed), so it is called with `nil`. But `builtin.grep_live` passes querytick, which has the same value as `H.querytick`.

Thanks again for narrowing down the problem to one particular line!

PR #2272, opened 2026-02-17
fix(files): improved path normalization on windows

nickvonkaenel:
Adds vim.api.nvim_win_is_valid() check before accessing window buffers to prevent errors when window IDs become invalid, such as with empty folders.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -1061,11 +1061,9 @@ MiniFiles.get_explorer_state = function()
   H.explorer_ensure_target_window(explorer)
   local windows = {}
   for _, win_id in ipairs(explorer.windows) do
-    if vim.api.nvim_win_is_valid(win_id) then
-      local buf_id = vim.api.nvim_win_get_buf(win_id)
-      local path = (H.opened_buffers[buf_id] or {}).path
-      table.insert(windows, { win_id = win_id, path = path })
-    end
+    local buf_id = vim.api.nvim_win_get_buf(win_id)
+    local path = (H.opened_buffers[buf_id] or {}).path
+    table.insert(windows, { win_id = win_id, path = path })
   end
 
   return {
@@ -2634,7 +2632,13 @@ end
 
 H.fs_normalize_path = function(path) return (path:gsub('/+', '/'):gsub('(.)/$', '%1')) end
 if H.is_windows then
-  H.fs_normalize_path = function(path) return (path:gsub('\\', '/'):gsub('([^/])/+', '%1/'):gsub('(.)[\\/]$', '%1')) end
+  H.fs_normalize_path = function(path)
+    path = path:gsub('\\', '/'):gsub('///+', '//')
+    local prefix = path:sub(1, 2)
+    local rest = path:sub(3):gsub('/+', '/'):gsub('/+$', '')
+    if prefix:match('%a:') then prefix = prefix .. '/' end
+    return prefix .. rest
+  end
 end
 
 H.fs_is_imaginary_path = function(path) return path:sub(-1) == '\000' end


echasnovski:
Thanks for the PR!

> ... to prevent errors when window IDs become invalid, such as with empty folders.

Could you please elaborate on this? How can I (and future test) reproduce this error?

nickvonkaenel:
Admittedly most of the time I get this error is due to an autocmd I set up on MiniFilesWindowUpdate which calls MiniFiles.get_explorer_state() but I believe this could happen in other situations.

Occasionally there will be an invalid window id and it causes an error. I noticed this mostly when working on Windows. You can ignore the empty folders comment, I did some more testing and I discovered that previewing files where you don't have read access causes the error consistently.

So to repro:
Use Windows (sorry)
Set up an autocmd on MiniFilesWindowUpdate that calls MiniFiles.get_explorer_state()
Try to preview a file like a system file where you don't have read access.

echasnovski:
> You can ignore the empty folders comment, I did some more testing and I discovered that previewing files where you don't have read access causes the error consistently.

I think this might be the culprit that is worth fixing.

'mini.files' already [does the check](https://github.com/nvim-mini/mini.nvim/blob/ca9adcfc4e796b292cc9072f4952318ef5a7f6c0/lua/mini/files.lua#L2310-L2311) of whether a file is readable, which I thought was enough.

As I don't have Windows access, could you please test on Windows for the file without read access:
- `:=vim.uv.fs_open('unreadable-file', 'r', 1)` . I think it should show `nil`.
- `:=vim.fn.filereadable('unreadable-file')` . I think it should show 0.

Also, could you try applying the following patch and see if it fixes the initial problem you have:

```diff
diff --git a/lua/mini/files.lua b/lua/mini/files.lua
index 331bead2..bee195e2 100644
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2305,6 +2305,9 @@ H.buffer_update_directory = function(buf_id, path, opts, is_preview)
 end
 
 H.buffer_update_file = function(buf_id, path, opts, _)
+  if vim.fn.filereadable(path) == 0 then
+    return H.set_buflines(buf_id, { '-No-access' .. string.rep('-', opts.windows.width_preview) })
+  end
   -- Work only with readable text file. This is not 100% proof, but good enough.
   -- Source: https://github.com/sharkdp/content_inspector
   local fd, width_preview = vim.loop.fs_open(path, 'r', 1), opts.windows.width_preview
```

nickvonkaenel:
vim.uv.fs_open('unreadable-file', 'r', 1) is returning nil and returning an operation not permitted string
vim.fn.filereadable('unreadable-file') is returning 1

The patch doesn't fix the problem. I don't think that early return there would fix the problem even if filereadable returned 0 correctly. I tried setting the compare value to 1 just to see and I still got the error.

echasnovski:
> vim.uv.fs_open('unreadable-file', 'r', 1) is returning nil and returning an operation not permitted string vim.fn.filereadable('unreadable-file') is returning 1
> 
> The patch doesn't fix the problem. I don't think that early return there would fix the problem even if filereadable returned 0 correctly. I tried setting the compare value to 1 just to see and I still got the error.

Hmm... Then I don't really see how the fact that the file is unreadable can result in this problem.

Anyway, it seems to me that a simple validation from this PR doesn't resolve the culprit of the problem. As I can not reproduce (maybe yet), I'd ask you to dig a bit deeper. Ideally, all windows in `explorer.windows` should always be valid. If this can be adjusted in this particular use case, that would be the best.

I'll also try to reproduce the issue.

nickvonkaenel:
Ok I dug further and found the culprit.

With preview enabled in C:/, hovering files like:
DumpStack.log.tmp
pagefile.sys
swapfile.sys

shows -No-fs-entry-, and trying to open them says they don’t exist. But they do exist, they’re just protected/locked by Windows.

In these cases vim.loop.fs_stat() returns nil, so fs_is_present_path() treats them as not present. That makes fs_get_type() return nil, which leads to -No-fs-entry-. Feels like we should distinguish:

truly missing: -No-fs-entry-

present but inaccessible: -No-access-

One fix that worked for me is: if fs_stat() fails on Windows, fall back to scanning the parent dir for the basename:

```lua
H.fs_is_present_path = function(path)
  if H.fs_is_imaginary_path(path) then return false end
  if vim.loop.fs_stat(path) ~= nil then return true end

  if H.is_windows then
    local parent = H.fs_get_parent(path)
    if not parent then return false end

    local base = H.fs_get_basename(path)
    local fs = vim.loop.fs_scandir(parent)
    if not fs then return false end

    while true do
      local name = vim.loop.fs_scandir_next(fs)
      if not name then break end
      if name == base then return true end
    end
  end

  return false
end
```

This also seemed related to the “invalid win_id” I was seeing: MiniFilesWindowUpdate can fire mid-refresh (called inside H.explorer_refresh_depth_window()), so explorer.windows isn’t necessarily stable if an integration calls get_explorer_state() right there. If that’s expected, maybe a MiniFilesRefreshComplete event at the end of H.explorer_refresh() would give integrations a safe point to do layout/state work without vim.schedule().

echasnovski:
> shows -No-fs-entry-, and trying to open them says they don’t exist. But they do exist, they’re just protected/locked by Windows.
> 
> In these cases vim.loop.fs_stat() returns nil, so fs_is_present_path() treats them as not present. That makes fs_get_type() return nil, which leads to -No-fs-entry-.

That is relatively strange. Could you please share what is the error part of `vim.loop.fs_stat(path)` for such paths?

> One fix that worked for me is: if fs_stat() fails on Windows, fall back to scanning the parent dir for the basename:

That is a bit too much of an adjustment. I'd rather declare as "known to not work" if that is the only solution. Maybe falling back to `vim.fn.filereadable()` could work? Could you please try the following patch and see if it fixes the issue for you:

```diff
diff --git a/lua/mini/files.lua b/lua/mini/files.lua
index 331bead2..424934b7 100644
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2638,6 +2638,11 @@ end
 H.fs_is_imaginary_path = function(path) return path:sub(-1) == '\000' end
 
 H.fs_is_present_path = function(path) return vim.loop.fs_stat(path) ~= nil and not H.fs_is_imaginary_path(path) end
+if H.is_windows then
+  H.fs_is_present_path = function(path)
+    return (vim.loop.fs_stat(path) ~= nil or vim.fn.filereadable(path) == 1) and not H.fs_is_imaginary_path(path)
+  end
+end
 
 H.fs_child_path = function(dir, name) return H.fs_normalize_path(string.format('%s/%s', dir, name)) end
 
```

echasnovski:
Also, after looking around for `libuv` issues, could you please share the `libuv` and Windows versions on your machine? Execute `:=vim.uv.version_string()` to see the `libuv` version.

nickvonkaenel:
I'm on Windows 11, libuv 1.51.0

I tried your patch and it still shows displays no-fs-entry for those specific files.

vim.loop.fs_stat returns:

```
nil
"ENOENT: no such file or directory: C:\\swapfile.sys"
"ENOENT"
```

nickvonkaenel:
Ok actually I just noticed something weird kinda by accident

vim.loop.fs_stat('C:\\swapfile.sys') returns nil as shown above

vim.loop.fs_stat('C:\\\\swapfile.sys') returns this:

```
{
  atime = {
    nsec = 578239700,
    sec = 1771430020
  },
  birthtime = {
    nsec = 390903900,
    sec = 1753380132
  },
  blksize = 4096,
  blocks = 32768,
  ctime = {
    nsec = 578239700,
    sec = 1771430020
  },
  dev = 43224006,
  flags = 0,
  gen = 0,
  gid = 0,
  ino = 2.8147497674633e+14,
  mode = 33206,
  mtime = {
    nsec = 578239700,
    sec = 1771430020
  },
  nlink = 1,
  rdev = 0,
  size = 16777216,
  type = "file",
  uid = 0
}

```

If I do the same thing on a normal file in the C directory, it returns valid info regardless of how many backslashes I add. So I think maybe the issue is we aren't adding the proper amount of backslashes to the file paths and this is handled fine for most files but causes issues on others!

Edit: github is formatting my backslashes for some reason. I'm actually escaping them properly doing two then four in my tests.

nickvonkaenel:
Ok I just confirmed that this fixes the issue and properly returns no access for those files. Not suggesting this as an actual fix but it points to a possible solution
```lua

if H.is_windows then
  H.fs_is_present_path = function(path)
    path = path:gsub('/', '\\')
    path = path:gsub(':\\', ':\\\\')
    return (vim.loop.fs_stat(path) ~= nil or vim.fn.filereadable(path) == 1) and not H.fs_is_imaginary_path(path)
  end
end

```

echasnovski:
> I'm on Windows 11, libuv 1.51.0
> 
> I tried your patch and it still shows displays no-fs-entry for those specific files.
> 
> vim.loop.fs_stat returns: ... `ENOENT`

That is very strange. As in it should either already be resolved on 1.51.0 (libuv/libuv#4566) or a `libuv` issue (I haven't seen it associated with `ENOENT` error).

> vim.loop.fs_stat('C:\swapfile.sys')

This treats the `\s` as a single character in Lua. It has to be `'C:\\swapfile.sys'` if you want to insert literal `\` after `:`.

All paths in 'mini.files' [are normalized](https://github.com/nvim-mini/mini.nvim/blob/dbb073bd2ed4a7bb35daafc7989567f0ff1426ee/lua/mini/files.lua#L2633-L2636) to use forward slashes, though. Could you check which *exactly* path 'mini.files' uses in this case? Adding logging with something like `table.insert(_G.log, path)` in proper place with one extra `_G.log = {}` is what I usually do (except with ['mini.misc' `log_xxx()` functions](https://nvim-mini.org/mini.nvim/doc/mini-misc.html#minimisc.log_add)).

For example, what is the exact input for `H.fs_is_present_path()` for what should be `C:\\swapfile.sys` file? Maybe 'mini.files' normalizes it in the wrong way.

echasnovski:
> Ok I just confirmed that this fixes the issue and properly returns no access for those files.

That is nice. Could you please follow the suggestion in my previous comment and show what is the `path` exactly? This sounds like a path normalization issue.

And after that please check if dropping `vim.fn.filereadable(path) == 1` would still work after making `path:gsub('/', '\\'):path:gsub(':\\', ':\\\\')` adjustments?

nickvonkaenel:
> This treats the \s as a single character in Lua. It has to be 'C:\\swapfile.sys' if you want to insert literal \ after :.

I understand. That is what I wrote, for some reason the github changed it (I made an edit to point it out)

So to be clear:

```lua
vim.loop.fs_stat('C:\\swapfile.sys') -- returns nil

vim.loop.fs_stat('C:\\\\swapfile.sys') -- returns a table
```

The path is being passed in as
C:/swapfile.sys

so it is normalized with forward slashes but only one.

> And after that please check if dropping vim.fn.filereadable(path) == 1 would still work after making path:gsub('/', '\\'):path:gsub(':\\', ':\\\\') adjustments?

I can confirm that it still works without vim.fn.filereadable with these adjustments

TheLeoP:
I managed to reproduce the original issue, but I'm seeing differences on how `mini.files` behaves on my Windows installation regarding future comments.

With the configuration

```lua
require("mini.files").setup {
  windows = {
    preview = true,
  },
}

  callback = function()
    MiniFiles.get_explorer_state()
  end,
})
```

doing `cd C:/` and hovering `swapfile.sys` `mini.files` correctly recognizes that the file does exists and that I do not have access to read it

image

(there seems to be an issue regarding how the title is shown, probably because Windows absolute paths start with a letter, instead of `/`. I think I also found a similar issue regarding `MiniFiles.open()`, but I'll look into it and open a separated issue later)

If I then use the `h` keybind I get the error mentioned in the description of the PR

image

Regarding the issue with `fs_stat`, I can confirm that `vim.uv.fs_stat('C:/swapfile.sys')` returns `nil` and `ENOENT: no such file or directory` as an error. However, it doesn't seem to be a matter of forward slashes vs backslashes since `vim.uv.fs_stat('C://swapfile.sys')` returns the correct stat table just like the version with a double backslash at the beginning. As I said before, I personally don't know if this is related to the original error as I don't even see the`-No-fs-entry-` also mentioned

echasnovski:
> So to be clear:
> 
> ```lua
> vim.loop.fs_stat('C:\\swapfile.sys') -- returns nil
> 
> vim.loop.fs_stat('C:\\\\swapfile.sys') -- returns a table
> ```
> 
> The path is being passed in as C:/swapfile.sys

Okay, so it might be a path normalization issue. Which is a bit weird, since [specs for Windows path](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats) all show that `C:\something` is a fully valid path.

@nickvonkaenel, does the following patch specifically for normalization (i.e. no other changes applied to 'lua/mini/files.lua') fix the issue and doesn't introduce new ones:

```diff
diff --git a/lua/mini/files.lua b/lua/mini/files.lua
index 331bead2..1d493919 100644
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2632,7 +2632,9 @@ end
 
 H.fs_normalize_path = function(path) return (path:gsub('/+', '/'):gsub('(.)/$', '%1')) end
 if H.is_windows then
-  H.fs_normalize_path = function(path) return (path:gsub('\\', '/'):gsub('([^/])/+', '%1/'):gsub('(.)[\\/]$', '%1')) end
+  H.fs_normalize_path = function(path)
+    return (path:gsub('\\', '/'):gsub('([^/])/+', '%1/'):gsub('^(%a:)/([^/])', '%1//%2'):gsub('(.)[\\/]$', '%1'))
+  end
 end
 
 H.fs_is_imaginary_path = function(path) return path:sub(-1) == '\000' end
```

---

@TheLeoP, as I don't have Windows, could you also please take a deeper dive here and in the other issue you noticed?

TheLeoP:
As far as I can tell, the issue comes from the fact that the root path `C:/` is being normalized as `C:/C:`, so 

https://github.com/nvim-mini/mini.nvim/blob/ca9adcfc4e796b292cc9072f4952318ef5a7f6c0/lua/mini/files.lua#L2669

returns `false` and `mini.files` tries to go up a directory, failing to do so. I even think that the lack of access to a file is a red hearing, as I can reproduce the original issue without hovering such a file.

This happens because

https://github.com/nvim-mini/mini.nvim/blob/ca9adcfc4e796b292cc9072f4952318ef5a7f6c0/lua/mini/files.lua#L2634-L2636

normalizes `C:\\` as `C:` (removing the trailing slash because of the pattern in the last `:gsub('(.)[\\/]$', '%1')`, which doesn't happen for Linux because `/` does not match the equivalent `gsub('(.)/$', '%1')`) which isn't recognized as the top level directory and then gets expanded to `C:/C:`.

Removing the last `gsub` (as a proof of concept, because it should be fixed instead to work in all cases) does indeed remove the issue for me @echasnovski

TheLeoP:
@nickvonkaenel could you check if #2277 fixes your issues?

nickvonkaenel:
@TheLeoP You are definitely on to something there, as that does fix the issue. I think there needs to be a more robust function like you mentioned because there are other bugs that present due to the change, for example:

```lua
E5108: Lua: (mini.files) `branch` contains not a parent-child pair: "C:/" and "C:///test.txt"
stack traceback:
	[C]: in function 'error'
	...vim-data\site\pack\core\opt\mini.nvim/lua/mini/files.lua:2896: in function 'error'
	...vim-data\site\pack\core\opt\mini.nvim/lua/mini/files.lua:2888: in function 'validate_branch'
	...vim-data\site\pack\core\opt\mini.nvim/lua/mini/files.lua:1112: in function 'set_branch'
	...vim-data\site\pack\core\opt\mini.nvim/lua/mini/files.lua:1004: in function 'reveal_cwd'
```

TheLeoP:
Thanks for the feedback! @nickvonkaenel Could you give some more details on how you got that new error? Are you using #2277 or the temporary fix I mentioned in my comment?

---

Regarding the previous discussion about `swapfile.sys`, I can't find any official documentation about it, but, empirically, I found that it seems to be a special file specifically located in the path `C://swapfile.sys` and not the path `C:/swapfile.sys`.

nickvonkaenel:
I can get the new error with your change by running
```lua
MiniFiles.open(vim.api.nvim_buf_get_name(0), false)
```
with one of the files in the C:/ directory open

I made an AI slop robust function and this fixes all the errors. I can try to make something that's less lines of code once I have time if that helps.
```lua
if H.is_windows then
  H.fs_normalize_path = function(path)
    if type(path) ~= 'string' or path == '' then return path end

    -- 1) Normalize separators
    path = path:gsub('\\', '/')

    -- 2) Extended-length / device paths (keep the prefix exactly)
    --    \\?\C:\foo  -> //?/C:/foo
    --    \\.\COM1    -> //./COM1
    if path:sub(1, 4) == '//?/' or path:sub(1, 4) == '//./' then
      local prefix = path:sub(1, 4)
      local rest = path:sub(5)
      -- collapse slashes in rest
      rest = rest:gsub('/+', '/')
      -- strip trailing slash unless the rest is empty (just the prefix)
      if rest ~= '' then rest = rest:gsub('/+$', '') end
      return prefix .. rest
    end

    -- 3) UNC paths (must keep the initial double slash)
    --    \\server\share -> //server/share
    if path:sub(1, 2) == '//' then
      local rest = path:sub(3)
      rest = rest:gsub('^/+', '') -- don't allow more than 2 leading slashes total
      rest = rest:gsub('/+', '/')
      rest = rest:gsub('/+$', '')
      return '//' .. rest
    end

    -- 4) Drive-letter paths
    -- Absolute:  C:\Users -> C:/Users
    -- Root:      C:\ or C:// -> preserve root form
    -- Relative:  C:foo  (leave as C:foo, but still normalize slashes inside)
    if path:match('^[A-Za-z]:') then
      local drive = path:sub(1, 2) -- "C:"
      local rest = path:sub(3) -- includes whatever after "C:"

      -- If it looks absolute/rooted (next char was a slash), keep exactly one slash
      if rest:sub(1, 1) == '/' then
        -- preserve "C://..." (if user provided multiple slashes right after the colon)
        local m = rest:match('^(/*)')
        local lead = m or '/'
        local tail = rest:sub(#lead + 1)

        tail = tail:gsub('/+', '/') -- collapse internal doubles
        -- trailing slash rules: keep drive root slashes, strip otherwise
        if tail == '' then
          return drive .. lead -- e.g. "C:/" or "C://"
        else
          tail = tail:gsub('/+$', '')
          return drive .. lead .. tail
        end
      else
        -- Relative drive path (C:foo). Normalize inside but don't invent a slash.
        rest = rest:gsub('/+', '/')
        rest = rest:gsub('/+$', '')
        return drive .. rest
      end
    end

    -- 5) Posix-ish paths
    -- Preserve leading "/" (root), collapse duplicates, strip trailing unless root.
    local is_root = (path == '/') or (path:match('^/+$') ~= nil)
    if is_root then return '/' end

    local leading = path:sub(1, 1) == '/' and '/' or ''
    local rest = leading == '/' and path:sub(2) or path

    rest = rest:gsub('/+', '/')
    rest = rest:gsub('/+$', '')

    return leading .. rest
  end
end
```

TheLeoP:
> I can get the new error with your change by running
> 
> ```lua
> MiniFiles.open(vim.api.nvim_buf_get_name(0), false)
> ```

Sadly, I can't reproduce your new issue. For me, `mini.files` explorer is open normally when doing that

nickvonkaenel:
> > I can get the new error with your change by running
> > ```lua
> > MiniFiles.open(vim.api.nvim_buf_get_name(0), false)
> > ```
> 
> Sadly, I can't reproduce your new issue. For me, `mini.files` explorer is open normally when doing that

Interesting. Are you on 0.12? I get this error if I remove any one of the gsub calls

nickvonkaenel:
> > I can get the new error with your change by running
> > ```lua
> > MiniFiles.open(vim.api.nvim_buf_get_name(0), false)
> > ```
> 
> Sadly, I can't reproduce your new issue. For me, `mini.files` explorer is open normally when doing that

Actually try this:
```lua
MiniFiles.open(vim.api.nvim_buf_get_name(0), false)
MiniFiles.reveal_cwd()
```

nickvonkaenel:
I've changed this PR to reflect the root cause of the issue and also committed my solution for the issue.

```lua
if H.is_windows then
  H.fs_normalize_path = function(path)
    path = path:gsub('\\', '/'):gsub('///+', '//')
    local prefix = path:sub(1, 2)
    local rest = path:sub(3):gsub('/+', '/'):gsub('/+$', '')
    if prefix:match('%a:') then prefix = prefix .. '/' end
    return prefix .. rest
  end
end
```
This new function handles UNC and Extended-length paths (starting with //) and also gives improved behavior with forward slashes after drive letters. I've tested it and it's resolved all of my issues.

echasnovski:
Thank you both for taking a closer look!

As far as I understand, the summary for problematic Windows path normalization is:
- For `C://` path it should preserve trailing slash(es). All of [this](https://github.com/nvim-mini/mini.nvim/pull/2272#issuecomment-3931219484), #2277, and built-in `vim.fs.normalize()` result in `C:/`.
- The `C://file` should be `C://file`.
- The `C:///file` should reduce the number of slashes. Probably, to `C://file`.
- It should not matter whether `/` or `\` is used.

The hypothetical questions is: should `C:/file` (one slash) be normalized as `C://file`? My hope and general logic says "no". Plus Neovim's `vim.fs.fnamemodify(path, ':p')` and later `vim.fs.abspath()` might already result in paths with `C:\\` and not `C:\`.

The current implementation in this PR does always ensure `C://` even if the input starts only with `C:/`.

---

Taking into account that `:` is a not supported filename character on Windows, I have a possibly more concise solution.

```diff
diff --git a/lua/mini/files.lua b/lua/mini/files.lua
index 331bead2..78b8f185 100644
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2632,7 +2632,9 @@ end
 
 H.fs_normalize_path = function(path) return (path:gsub('/+', '/'):gsub('(.)/$', '%1')) end
 if H.is_windows then
-  H.fs_normalize_path = function(path) return (path:gsub('\\', '/'):gsub('([^/])/+', '%1/'):gsub('(.)[\\/]$', '%1')) end
+  H.fs_normalize_path = function(path)
+    return (path:gsub('\\', '/'):gsub('([^/:])/+', '%1/'):gsub('([^:])/+$', '%1'):gsub('^(%a):/+([^/])', '%1://%2'))
+  end
 end
 
 H.fs_is_imaginary_path = function(path) return path:sub(-1) == '\000' end
```

@nickvonkaenel and @TheLeoP, could you please test this approach and see if it fixes the issues you separately found?

---

Side notes:
- 'mini.files' does not use `vim.fs.normalize` for better performance.
- All online documentation says that `C:/file` should work just fine and there should be no need for `C://file` (which is somewhat incorrect even). So maybe this is partially a `libuv` issue and forcing `C://` in file paths is wrong and only `C://` -> `C:/` should be taken into account.

    I would appreciate also testing the variant of the above patch with `gsub('^(%a):/+([^/])', '%1:/%2')` as the last `gsub`. I.e. to ensure a `/` and not `//` as the volume prefix.

TheLeoP:
>  could you please test this approach and see if it fixes the issues you separately found?

If fixes all of the issues I found. It even (maybe unintentionally?) solves the issue @nickvonkaenel mentioned about `MiniFiles.reveal_cwd()` (that is also an issue with #2277).

> I would appreciate also testing the variant of the above patch with `gsub('^(%a):/+([^/])', '%1:/%2')` as the last `gsub`. I.e. to ensure a `/` and not `//` as the volume prefix.

I'm seeing the error that I mentioned in https://github.com/nvim-mini/mini.nvim/pull/2277#issuecomment-3930589118 with this change. I also realized that it only happens if the explorer gets in a state where `explorer.branch` looks like `{ "C://", "C:/vfcompat.dll" }` (the root must have 2 trailing slashes for the error to happen), the cursor is moved on top of a files that requires `C://` and then on top of a file that doesn't require it. I can reliably get the explorer to said state by ` :lua MiniFiles.open('C:/vfcompat.dll')` where `vfcompat.dll` is any file under `C:/` that does not require `C://` (weirdly enough).

It also doesn't solve the issue with `MiniFiles.reveal_cwd()` because

https://github.com/nvim-mini/mini.nvim/blob/b70d9f932240b69be98ac46f1056581825ed0523/lua/mini/files.lua#L2881

has the values

```
┆H.validate_branch#for┆ ╎(parent .. '/' .. child:match('[^/]+$'))╎: "C://$Recycle.Bin"
┆H.validate_branch#for┆ ╎res[i] ╎: "C:/$Recycle.Bin"
```

which do not match unless `C:/` is always normalized to `C://`. My guess is that that's not the intention of the code, but I don't fully understand the intended logic, so I may be wrong.

---

> So maybe this is partially a libuv issue and forcing C:// in file paths is wrong and only C:// -> C:/ should be taken into account.

That may be true. My empirical conclusion that certain files require `C://` came from the fact that `:edit` autocompleted them only for `C://` and not for `C:/` (and it also only gave a permission warning when trying to open them with `C://`). But Neovim may be indeed using libuv under the hood to handle file system access on the C side (?

`:e c:/`, some files do not show up

image

`:e c://` all of the files that (I think) require `C://` do show up

image

nickvonkaenel:
I can confirm what @TheLeoP is reporting. The function you provided works for me, but if I change the last gsub function then there is an error with reveal_cwd / validate_branch and 'branch contains not a parent-child pair'

echasnovski:
Thank you both! I'll fix it soon-ish with some additional testing to ensure that this behavior on Windows persists.

echasnovski:
All right. I've update `main` branch to follow Windows path normalization on Windows as in [this comment](https://github.com/nvim-mini/mini.nvim/pull/2272#issuecomment-3933044987). It required a couple of tweaks (mostly because this *does* look like not a standard way of normalizing paths, since current directory uses `C:\dir` notation). Plus there is now a (first of its kind) test of internal helper: mostly to document this conversation and make the future changes at least more educated.

@nickvonkaenel and @TheLeoP, huge thanks to you both for helping me debug this! Let's hope this change won't trigger some other problems down the road.

PR #2277, opened 2026-02-19
fix(files): properly normalize root directory on Windows

TheLeoP:

Fixes the issue described in #2272 based on the findings of https://github.com/nvim-mini/mini.nvim/pull/2272#issuecomment-3930187872

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2632,7 +2632,11 @@ end
 
 H.fs_normalize_path = function(path) return (path:gsub('/+', '/'):gsub('(.)/$', '%1')) end
 if H.is_windows then
-  H.fs_normalize_path = function(path) return (path:gsub('\\', '/'):gsub('([^/])/+', '%1/'):gsub('(.)[\\/]$', '%1')) end
+  H.fs_normalize_path = function(path)
+    path = path:gsub('\\', '/'):gsub('([^/])/+', '%1/')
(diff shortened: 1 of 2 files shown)


TheLeoP:
Actually, after this PR, simply hovering one of `C://pagefile.sys`, `C://swapfile.sys` or `C://DumpStack.log.tmp` and then moving the cursor on top of any other file on the `C:/` directory will result in a similar error to the one described in #2272 with the same minimal reproduction

```lua
    require("mini.files").setup {
      windows = {
        preview = true,
      },
    }
      callback = function()
        local s = MiniFiles.get_explorer_state()
      end,
    })
```

TheLeoP:
This PR already handles the most common case of root directory normalization in Windows, but after a few iterations I ended up with

```lua
if H.is_windows then
  H.fs_normalize_path = function(path)
    path = path:gsub('\\', '/')
    local head, tail = path:match('^(%w:/+)(.*)$')
    -- preserve number of slashes if 1 or 2, normalize more than 2 slashes to 1
    if #head > 4 then head = head:gsub('/+', '/') end
    tail = tail:gsub('(.)/$', '%1')
    return head .. tail
  end
end
```

instead that also handles the aforementioned special paths that require specifically `C://` instead of just `C:/`, but I'm not sure if complicating the normalization logic even more is worth it for such a niche case. It's your call @echasnovski

nickvonkaenel:
This fix results in an error for me when using MiniFiles.reveal_cwd(). Please see the fix in my PR.

PR #2278, opened 2026-02-20
docs(git): fixed left/right quotation marks instead of apostrophes

Gotoro:

In all of READMEs in the copy-and-paste snippet code blocks sections instead of normal apostrophes `'` there were left and right quotes surrounding the code, which caused the copy-and-paste to not work:

`‘` U+2018 : LEFT SINGLE QUOTATION MARK {single turned comma quotation mark}
`’` U+2019 : RIGHT SINGLE QUOTATION MARK {single comma quotation mark}

Now they're all replaced with normal apostrophes.

I've used regex to replace them, but I've checked them all manually before submitting a pr.

The only places where those quotes are left in the project are in the `*Follow recommended ‘mini.deps’ installation*` sections and one right quote `’` in `tests/dir-git/help-output:31`: `Remove remote branches that don’t have a local counterpart.` somehow (lol).

### DIFF
--- a/readmes/mini-ai.md
+++ b/readmes/mini-ai.md
@@ -76,8 +76,8 @@ Here are code snippets for some common installation methods (use only one):
 
     | Branch | Code snippet                                                 |
     |--------|--------------------------------------------------------------|
-    | Main   | `add(‘nvim-mini/mini.ai’)`                                   |
-    | Stable | `add({ source = ‘nvim-mini/mini.ai’, checkout = ‘stable’ })` |
+    | Main   | `add('nvim-mini/mini.ai')`                                   |
+    | Stable | `add({ source = 'nvim-mini/mini.ai', checkout = 'stable' })` |
 
 </details>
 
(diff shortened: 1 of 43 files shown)


echasnovski:
Thanks for the PR! Very strange that I haven't noticed this before. I guess it proves that nobody used this part of the README before or that there is not so many 'mini.deps' users :)

One thing, though. All previous changes that touch all modules were done under with the main maintainer as commit author. And I'd prefer to keep it that way, so I'll do these changes myself soon-ish (hopefully tomorrow).

Gotoro:
Okay, cheers! Yeah, I was confused at first, why the snippet wouldn't work

echasnovski:
> Okay, cheers! Yeah, I was confused at first, why the snippet wouldn't work

Thanks for understanding! And congratulations for possibly being the first to actually use them :)

echasnovski:
This should now be resolved on the latest `main`. The `Follow recommended 'mini.deps' installation` was also a mistake and is fixed now. The 'mini.git' test example remains the same since it is what the manual page looks like.

Huge thanks again for noticing!

PR #2284, opened 2026-02-26
feat(jump)!: make dot-repeat jumping separate from regular jumping

### LINKED ISSUE #2054: [mini.jump] Breaks dot-repeat when jumping between repeats
### Contributing guidelines

### Module(s)

mini.jump

### Neovim version

0.11.x

### Description

When dot-repeating a command which includes a jump (e.g. `dt]`), and using a different jump between repeats, the original jump is lost and instead the other jump is repeated.

This works correctly with `mini.jump` disabled.

### Reproduction

    ```lua
    vim.fn.system(clone_cmd)

    vim.cmd('packadd mini.nvim')
    require('mini.deps').setup()

    require('mini.jump').setup()
    ```

4. Enter the following text:
    ```
    [foo, 1]
    [foo, 2]
    [foo, 3]
    ```

5. Move to the beginning of the file: `gg0`
6. Search for `foo`: `/foo<CR>`
7. Jump to the first comma: `f,`
8. Delete until the closing bracket: `dt]`
9. Go to the next search result and jump to the second comma: `nf,`
10. Repeat the deletion: `.`

**Expected result:** Everything up to the next closing bracket should get deleted:

```
[foo]
[foo]
[foo, 3]
```

**Actual result:** Everything up to the next comma gets deleted instead:

```
[foo]
[foo, 3]
```

abeldekat:
Resolve #2054

### DIFF
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -181,6 +181,11 @@ MiniJump.state = {
 MiniJump.jump = function(target, backward, till, n_times)
   if H.is_disabled() then return end
 
+  -- If dot-repeating the state needs to be restored after the jump
+  local needs_restore = MiniJump._is_expr and MiniJump.state.mode ~= 'no'
+  MiniJump._is_expr = nil
+  local state_snapshot = needs_restore and vim.deepcopy(MiniJump.state) or nil
+
   -- Cache inputs for future use
   H.update_state(target, backward, till, n_times)
 
@@ -230,6 +235,12 @@ MiniJump.jump = function(target, backward, till, n_times)
   -- Track cursor position to account for movement not caught by `CursorMoved`
   H.cache.latest_cursor = H.get_cursor_data()
   H.cache.has_changed_cursor = not vim.deep_equal(H.cache.latest_cursor, init_cursor_data)
+
+  -- If dot-repeating restore the state to its value from before this jump
+  if needs_restore then
+    state_snapshot.jumping = true
+    MiniJump.state = state_snapshot
+  end
 end
 
 --- Make smart jump
@@ -383,7 +394,8 @@ H.make_expr_jump = function(backward, till)
   return function()
     if H.is_disabled() then return '' end
 
-    H.update_state(nil, backward, till, vim.v.count1)
+    local count = vim.v.count1
+    H.update_state(nil, backward, till, count)
 
     -- Ask for `target` for non-repeating jump as this will be used only in
     -- operator-pending mode. Dot-repeat is supported via expression-mapping.
@@ -398,7 +410,10 @@ H.make_expr_jump = function(backward, till)
       if H.cache.has_changed_cursor then return end
       vim.cmd('undo!')
     end)
-    return 'v<Cmd>lua MiniJump.jump()<CR>'
+
+    -- Encode state in expression for dot-repeat
+    local args = string.format('%s,%s,%s,%s', vim.inspect(target), backward, till, count)
+    return 'v<Cmd>lua MiniJump._is_expr=true; MiniJump.jump(' .. args .. ')<CR>'
   end
 end
 
--- a/tests/test_jump.lua
+++ b/tests/test_jump.lua
@@ -183,6 +183,12 @@ T['state']['updates `mode`'] = function()
 
   type_keys('d', 't', 'e')
   eq(get_state().mode, 'nov')
+  type_keys('V', 't', 'e')
+  eq(get_state().mode, 'V')
+  child.ensure_normal_mode()
+  -- Ensure dot-repeat does not update mode after the jump
+  type_keys('.')
+  eq(get_state().mode, 'V')
   child.lua('MiniJump.stop_jumping()')
 end
 
@@ -402,7 +408,7 @@ T['Jumping with f/t/F/T']['works in Operator-pending mode'] = new_set({
     type_keys('d', '2', key, 'e')
     eq(get_lines(), { line_seq[3] })
 
-    -- Just typing `key` shouldn't repeat action
+    -- Just typing `key` shouldn't repeat motion
     local cur_pos = get_cursor()
     type_keys(key)
     eq(get_cursor(), cur_pos)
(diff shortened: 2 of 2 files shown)


abeldekat:
I am not sure if the extra tests in 'works after jump in Operator-pending mode' are actually needed, as the behavior of `;` has not been changed.
It still reuses the jump from dot-repeat. The motion is only different because dot-repeat itself has been changed.

tests/test_jump.lua
```diff
@@ -582,6 +582,36 @@ T['Jumping with f/t/F/T']['can be dot-repeated if did not jump at first'] = func
   validate('dT', 1, 5, 'abcdg')
 end
 
+T['Jumping with f/t/F/T']['only uses jumps from actions in dot-repeat'] = function()
```
echasnovski: The description of the test case looks confusing to me. Maybe something like "dot-repeat is not affected by regular jumping"?

echasnovski (changes requested):
Thanks for the PR!

The idea and tests overall look good! The large amount of comments is only there because it is possible to focus on little things :)

echasnovski:
Also, please adjust commit message to fit the new PR title. Dot-repeat is technically not only about changes (yank can be dot-repeated under some circumstances, for example).

abeldekat:
> Also, please adjust commit message to fit the new PR title. Dot-repeat is technically not only about changes (yank can be dot-repeated under some circumstances, for example).

Indeed. However, as I understand this PR, the title should be:

"ensure dot-repeat jumping is not affected by regular jumping"

echasnovski:
> Indeed. However, as I understand this PR, the title should be:
> 
> "ensure dot-repeat jumping is not affected by regular jumping"

[It goes both ways](https://github.com/nvim-mini/mini.nvim/pull/2284#discussion_r2858600740). And it is also a breaking change. So updated to be "make dot-repeat jumping separate from regular jumping".

lua/mini/jump.lua
```diff
@@ -398,7 +398,11 @@ H.make_expr_jump = function(backward, till)
       if H.cache.has_changed_cursor then return end
       vim.cmd('undo!')
     end)
-    return 'v<Cmd>lua MiniJump.jump()<CR>'
+
+    -- Encode state in expression for dot-repeat
+    local state = MiniJump.state
```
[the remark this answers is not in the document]
echasnovski: > That change suddenly uses `backward=true` and `till=true`, which might not be what the user expects.

But this is exactly how `nvim --clean` behaves. Which is the main source of truth right now.

tests/test_jump.lua
```diff
@@ -878,6 +908,26 @@ T['Repeat jump with ;']['works after jump in Operator-pending mode'] = function(
...
+
+  type_keys('$', 'T', '3') -- antagonist motion, no count
+  eq(get_lines(), { '3e4e5e6e7e8e' })
+  eq(get_cursor(), { 1, 1 })
+
+  type_keys('.') -- dot-repeat action
+  eq(get_lines(), { '36e7e8e' })
+  eq(get_cursor(), { 1, 1 })
+
+  type_keys(';') -- repeat motion from latest dot-repeat
```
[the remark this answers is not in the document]
abeldekat: I think it should work as expected now.

See MiniJump.expr_jump
See test `T['state']['updates mode']`
See test `T['Repeat jump with ;']['works after jump in Operator-pending mode']`

doc/mini-jump.txt
```diff
@@ -189,6 +189,15 @@ Parameters ~
   previous/current line. Note that with backward jump this might lead to
   cursor being on target if can't be put past the line.
 
+------------------------------------------------------------------------------
+                                                          *MiniJump.expr_jump()*
+                          `MiniJump.expr_jump`({...})
```
echasnovski: No new exported functions, please. This is too much for such a small gain.

I still think that it should be possible to adjust `MiniJump.jump()` to work in this case. Like maybe setting temporary private "flag" inside returned expression (which is executed on every dot-repeat).

echasnovski (changes requested):

doc/mini-jump.txt
abeldekat: Done. I realized I can use `MiniJump.state.mode == 'no' and vim.fn.mode(1) == 'nov'` to distinguish the initial call from `make_expr_jump` from the dot-repeat.

abeldekat: I pushed a new commit

lua/mini/jump.lua
```diff
@@ -181,6 +181,12 @@ MiniJump.state = {
 MiniJump.jump = function(target, backward, till, n_times)
   if H.is_disabled() then return end
 
+  -- If dot-repeating the state needs to be restored after the jump
+  local state_snapshot
+  if MiniJump.is_expr and not (MiniJump.state.mode == 'no' and vim.fn.mode(1) == 'nov') then
```
echasnovski: Why extra check for the forced charwise mode is needed?

lua/mini/jump.lua
```diff
@@ -230,6 +236,12 @@ MiniJump.jump = function(target, backward, till, n_times)
   -- Track cursor position to account for movement not caught by `CursorMoved`
   H.cache.latest_cursor = H.get_cursor_data()
   H.cache.has_changed_cursor = not vim.deep_equal(H.cache.latest_cursor, init_cursor_data)
+
+  -- If dot-repeating restore the state to its value from before this jump
+  if state_snapshot then
+    MiniJump.state = state_snapshot
+    MiniJump.state.jumping = true
+  end
```
echasnovski:
```suggestion
  if needs_restore then
    state_snapshot.jumping = true
    MiniJump.state = state_snapshot
    MiniJump._is_expr = nil
  end
```

tests/test_jump.lua
```diff
@@ -878,6 +910,26 @@ T['Repeat jump with ;']['works after jump in Operator-pending mode'] = function(
   type_keys('d', '2f', 'e', ';')
   eq(get_lines(), { '3e4e5e' })
   eq(get_cursor(), { 1, 3 })
+
+  -- Like nvim --clean, it does not use the latest dot repeat
+  set_lines({ '1e2e3e3e3e4e5e6' })
```
echasnovski: Why is there three `3` 😆 The point is to have a number for each `e` to easily check that the action is over the expected range (without having to count letters).

lua/mini/jump.lua
```diff
@@ -181,6 +181,12 @@ MiniJump.state = {
 MiniJump.jump = function(target, backward, till, n_times)
   if H.is_disabled() then return end
 
+  -- If dot-repeating the state needs to be restored after the jump
+  local state_snapshot
+  if MiniJump.is_expr and not (MiniJump.state.mode == 'no' and vim.fn.mode(1) == 'nov') then
+    state_snapshot = vim.deepcopy(MiniJump.state)
+  end
```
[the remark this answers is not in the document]
abeldekat: Very nice!

abeldekat:
I pushed the new commit

abeldekat:
Strange, some of my review comments end up as 'pending'.


About the question regarding forced charwise mode:

> Why extra check for the forced charwise mode is needed?

`   if MiniJump.is_expr and not (MiniJump.state.mode == 'no' and vim.fn.mode(1) == 'nov') then `

Apologies, I misread the question. Indeed, this is an example of doing too much...
Changed into:   

`local needs_restore = MiniJump._is_expr and MiniJump.state.mode ~= 'no'`

echasnovski:
Thanks again for the PR! I've merged into a temporary branch to polish a bit and then merge into `main`.

echasnovski:
The work from this PR should now be part of the latest `main`. Thanks again for the PR!

Couple of notes:
- After checking the PR locally, I noticed that [this comment](https://github.com/nvim-mini/mini.nvim/pull/2284#discussion_r2859875027) was not really addressed. It went unnoticed primarily because there was no dedicated test added for this.

    The reason for this was that created expression [hard coded the initial target](https://github.com/nvim-mini/mini.nvim/blob/79f34ad725745b5e3e42b6aee83bfb07faf3eeb3/lua/mini/jump.lua#L391). So doing something like `fe` -> `d;` (hard coded `e` as target) -> `fx` -> `.` (still used `e` as target although it should use `x`, as in `nvim --clean`). Using `nil` in case of `repeat_jump` (i.e. `;`) solved the issue.

- Adjusted some tests to be more to the point.

- Added small note in documentation to show that the dot-repeat behavior is intentional (i.e. "as in clean Neovim").

lua/mini/jump.lua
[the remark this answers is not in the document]
yahave: Fixed: changed `as never` to `as unknown as CoachingGuidelineStorePort` as suggested.

PR #2291, opened 2026-03-02
fix(jump): consumes a char on dot-repeat when target not in text

### LINKED ISSUE #2290: [mini.jump] Consumes a char on dot-repeat when target not in text
### Contributing guidelines

### Module(s)

mini.jump

### Neovim version

0.12 (!at least latest Nightly build!)

### Description

Related: Jump consumes characters when it cannot find the target #688

Issue #688 was fixed in commit [4f69339](https://github.com/nvim-mini/mini.nvim/commit/4f6933914b0ac8a3a08d9107c4c4348c9546ac59)

Although the fix was applied with dot-repeat in mind, the test scenario from OP still fails for dot-repeat, also on Neovim v0.9.5

The applied [solution](https://github.com/nvim-mini/mini.nvim/blob/4f6933914b0ac8a3a08d9107c4c4348c9546ac59/lua/mini/jump.lua#L368) does not target dot-repeat, as only the initial jump enters `vim.schedule` in `make_expr`

### Reproduction

```lua
vim.pack.add({ "https://github.com/nvim-mini/mini.nvim" })
require("mini.jump").setup()
```

4. Replace this with description of interactive reproduction steps along with the behavior you observe.
   Feel free to include images/videos/etc, this helps a lot.

https://github.com/user-attachments/assets/ec1790fa-43a2-483b-b352-e4102e9121a2

Text in testfile:

```txt
alice bob mark
alice
```
  
 
- Place cursor on line 1, first column
- Type: `dfm`
- Type: `.`
- Observe that the "a" from "mark" has been consumed

Alternatively:
- Place cursor on line 1, on the "a" in "mark"
- Type: `dfm`
- Type: `.`
- Observe that the "a" from "mark" has been consumed

</details>

abeldekat:
Resolve #2290

### DIFF
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -181,6 +181,14 @@ MiniJump.state = {
 MiniJump.jump = function(target, backward, till, n_times)
   if H.is_disabled() then return end
 
+  -- Ensure to undo "consuming a character" effect if there is no target found
+  -- Do it here to act on dot-repeat
+  local has_changed_cursor = false
+  local undo_no_move = function()
+    if not has_changed_cursor then vim.cmd('undo!') end
+  end
+  if MiniJump._is_expr then vim.schedule(undo_no_move) end
+
   -- Dot-repeat should not change the state, so save it to later restore
   local is_dot_repeat = MiniJump._is_expr and not MiniJump._is_expr_init
   MiniJump._is_expr, MiniJump._is_expr_init = nil, nil
@@ -234,7 +242,7 @@ MiniJump.jump = function(target, backward, till, n_times)
 
   -- Track cursor position to account for movement not caught by `CursorMoved`
   H.cache.latest_cursor = H.get_cursor_data()
-  H.cache.has_changed_cursor = not vim.deep_equal(H.cache.latest_cursor, init_cursor_data)
+  has_changed_cursor = not vim.deep_equal(H.cache.latest_cursor, init_cursor_data)
 
   -- Restore the state if needed
   if is_dot_repeat then
@@ -403,11 +411,6 @@ H.make_expr_jump = function(backward, till)
     local target = isnt_repeat_jump and H.get_target() or nil
     if isnt_repeat_jump and target == nil then return '<Esc>' end
     H.update_state(target)
-
-    vim.schedule(function()
-      if H.cache.has_changed_cursor then return end
-      vim.cmd('undo!')
-    end)
 
     -- Set a flag to distinguish first call from dot-repeat
     MiniJump._is_expr_init = true
--- a/tests/test_jump.lua
+++ b/tests/test_jump.lua
@@ -531,15 +531,21 @@ end
 
 T['Jumping with f/t/F/T']['does nothing if there is no place to jump'] = function()
   -- Normal mode
-  local validate_normal = function(keys, start_col, ref_mode)
-    set_lines({ 'abcdefg' })
+  local validate_normal = function(keys, start_col, ref_mode, line)
+    set_lines({ line })
     set_cursor(1, start_col)
 
     type_keys(keys, 'd')
 
     -- It shouldn't move anywhere and should not modify text
     eq(get_cursor(), { 1, start_col })
-    eq(get_lines(), { 'abcdefg' })
+    eq(get_lines(), { line })
+    eq(child.fn.mode(), ref_mode)
+
+    -- The above applies to subsequent dot-repeats as well
+    type_keys('.')
+    eq(get_cursor(), { 1, start_col })
+    eq(get_lines(), { line })
     eq(child.fn.mode(), ref_mode)
 
     -- Ensure there is no jumping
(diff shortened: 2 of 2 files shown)


abeldekat:
I wonder if the vim.schedule for undo could also be achieved with a one-time mode change event.

echasnovski:
Thanks for the PR!

My initial impression is that the change does more than is needed to fix the described issue. **Or** it fixes more than described in the commit or issue.

Because the part of "works on first try but doesn't work on dot-repeat" is probably meant to be fixed by moving the logic from `make_expr_jump` into `jump()` itself. Especially since there is a `MiniJump._is_exp` private indicator.

I think the following should fix the issue:

```diff
diff --git a/lua/mini/jump.lua b/lua/mini/jump.lua
index 3ac813d8..7165a1bf 100644
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -181,6 +181,14 @@ MiniJump.state = {
 MiniJump.jump = function(target, backward, till, n_times)
   if H.is_disabled() then return end
 
+  -- Ensure to undo "consuming a character" effect if there is no target found
+  -- Do it here to act on dot-repeat
+  local has_changed_cursor = false
+  local undo_no_move = function()
+    if not has_changed_cursor then vim.cmd('undo!') end
+  end
+  if MiniJump._is_expr then vim.schedule(undo_no_move) end
+
   -- Dot-repeat should not change the state, so save it to later restore
   local is_dot_repeat = MiniJump._is_expr and not MiniJump._is_expr_init
   MiniJump._is_expr, MiniJump._is_expr_init = nil, nil
@@ -234,7 +242,7 @@ MiniJump.jump = function(target, backward, till, n_times)
 
   -- Track cursor position to account for movement not caught by `CursorMoved`
   H.cache.latest_cursor = H.get_cursor_data()
-  H.cache.has_changed_cursor = not vim.deep_equal(H.cache.latest_cursor, init_cursor_data)
+  has_changed_cursor = not vim.deep_equal(H.cache.latest_cursor, init_cursor_data)
 
   -- Restore the state if needed
   if is_dot_repeat then
@@ -404,11 +412,6 @@ H.make_expr_jump = function(backward, till)
     if isnt_repeat_jump and target == nil then return '<Esc>' end
     H.update_state(target)
 
-    vim.schedule(function()
-      if H.cache.has_changed_cursor then return end
-      vim.cmd('undo!')
-    end)
-
     -- Set a flag to distinguish first call from dot-repeat
     MiniJump._is_expr_init = true
 
```

abeldekat:
> My initial impression is...

Yes, you are right. I even started with another version of your approach.

Whilst working on the issue, I encountered behavior which I thought to be a bug. Only on the last moment I saw that it was to be [expected](https://github.com/nvim-mini/mini.nvim/blob/0098de999048af0539183d625c52d733318a441b/tests/test_jump.lua#L717), also understanding [this](https://github.com/nvim-mini/mini.nvim/blob/0098de999048af0539183d625c52d733318a441b/tests/test_jump.lua#L520) test.

Because of that misperception, I thought I needed to get rid of [this](https://github.com/nvim-mini/mini.nvim/blob/0098de999048af0539183d625c52d733318a441b/lua/mini/jump.lua#L201) early return.

I submitted the current version because it worked and I think it has two main benefits:

- Only check if [target](https://github.com/nvim-mini/mini.nvim/blob/0098de999048af0539183d625c52d733318a441b/lua/mini/jump.lua#L200) is present when absolutely necessary
- No need to check if the cursor has changed in order to undo, as that check is a derivative of knowing that a successful jump has occurred. I can schedule a direct undo. 

Would you like me to resubmit with your version?

echasnovski:
> I submitted the current version because it worked and I think it has two main benefits:
> 
>     * Only check if [target](https://github.com/nvim-mini/mini.nvim/blob/0098de999048af0539183d625c52d733318a441b/lua/mini/jump.lua#L200) is present when absolutely necessary
> 
>     * No need to check if the cursor has changed in order to undo, as that check is a derivative of knowing that a successful jump has occurred. I can schedule a direct undo.

Both of these are nice improvements, but they are not about fixing the issue. It is hard to see if the approach is good if it does several things at once (i.e. fixes the issue and refactors the code to address two points above). I am not opposed to refactor but it needs to be a separate commit *that doesn't touch tests and still passes them*.

So I think following "as less diff as possible" while fixing the issue is a better approach here. If my above diff works, let's use it at least for now.

abeldekat:
> If my above diff works, let's use it at least for now.

Done.

Would you accept a refactoring pr?

echasnovski:
> Would you accept a refactoring pr?

Maybe after this is merged. But there are some other things in `jump()` code that caught my eye as worth addressing (it is a fairly old code). So there will be a lot of nit picking.

echasnovski:
Since tests fail on Nightly (due to upstream neovim/neovim#38137), I've merged into temporary branch to polish and push manually.

Thanks again for noticing the issue and doing the PR!

**Edit**: Should now be part of the `main` with a bit adjusted test code and commit message.

PR #2295, opened 2026-03-04
test(ALL): on Nightly `nvim_win_get_config ()` always returns "style"

abeldekat:
See PR 38122 in neovim/neovim

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -367,7 +367,6 @@ Callback for each file action event will receive `data` field
 - <to> - full path of entry after action (`nil` for permanent "delete" action).
 
 ------------------------------------------------------------------------------
-                                                            *MiniFiles-examples*
 # Toggle explorer ~
 
 Use a combination of |MiniFiles.open()| and |MiniFiles.close()|: >lua
@@ -404,6 +403,11 @@ updated internally. Use `MiniFilesWindowUpdate` event for them: >lua
 
       -- Ensure fixed height
       config.height = 10
+
+------------------------------------------------------------------------------
+                                                            *MiniFiles-examples*
+      -- Ensure style
+      config.style = nil
 
       -- Ensure no title padding
       local n = #config.title
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -398,6 +398,9 @@
 ---
 ---       -- Ensure fixed height
 ---       config.height = 10
+
+---       -- Ensure style
+---       config.style = nil
 ---
 ---       -- Ensure no title padding
 ---       local n = #config.title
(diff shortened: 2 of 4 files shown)


echasnovski:
Thanks for the PR!

Couple of notes about the commit message (I should probably document that in "MAINTAINING.md"...):
- Please use "Nightly" in commit message and not "nightly". This makes it consistent across commits and might help when searching through them.
- All function names should be in `` ` `` quotes. Ideally with `()` suffix, but sometimes okay to omit if there is no other good way to make first the title message shorter.
- Using `#38122` notation means that it links to issue/PR/discussion in this repository. Linking to other repository is done like `neovim/neovim#38122`, but please don't link like this in commit message because it will result in several mentions in that issue/PR (like after rebasing and similar). So in this case I usually prefer something like "See PR 38112 in neovim/neovim".
- Please preface any extra details in the commit body with "Details" line. This will potentially make it easier to parse commits for details.

---

As there is another failing test on Nightly, could you please take a look at it? I have a feeling it might be an actual Neovim regression, though, but let's hope it is not.

abeldekat:
I resubmitted and will look at the other failing test

echasnovski:
> So in this case I usually prefer something like "See PR 38112 in neovim/neovim".

Wait, it is 38122, not 38112. Sorry.

abeldekat:
New push....)

abeldekat:
@echasnovski, the following test fails for mini.files:

```txt
FAIL in tests/test_files.lua | Events | `MiniFilesWindowUpdate` can customize internally set window config parts:
  Failed expectation for screenshot equality to reference at "tests/screenshots/tests-test_files.lua---Events---`MiniFilesWindowUpdate`-can-customize-internally-set-window-config-parts"
  Cause: different `attr` cell at line 2 column 56, reference = "6", observed = "5"
```

This is also caused by PR 38122 in neovim/neovim: "fix(api): return "style" in nvim_win_get_config()"

The preceding commit(84d84) from PR 38138 "fix(window): style=minimal window loses local options when switching buffers#38138" passes

echasnovski:
> This is also caused by PR 38122 in neovim/neovim: "fix(api): return "style" in nvim_win_get_config()"

I see. Still needs to be adjusted, as the test is about for the officially documented example in `MiniFiles-examples`. Probably explicitly setting `config.style = nil` should be enough. Otherwise it will not show cursorline in explorer windows, it seems.

abeldekat:
> Still needs to be adjusted...

Done

echasnovski:
> > Still needs to be adjusted...
> 
> Done

Thanks! Not a big problem to adjust in 'mini.files', but I wonder if [this is better be reverted upstream](https://github.com/neovim/neovim/pull/38122#issuecomment-3996994189).

echasnovski:
Let's wait for the resolution of neovim/neovim#38152. The 'mini.files' fix (which is a bit flawed in its current form, by the way) won't be needed if that is merged.

echasnovski:
Thanks again for the PR!

Part of this (the `test(map)` case) should now be in the `main` branch. I was *this close* to merging the adjusted 'mini.files' fix, but it looks like neovim/neovim#38152 will get merged and it won't be needed.

PR #2302, opened 2026-03-06
feat(completion): Allow to set min window width

gx089:
This allows the user to set a minimum width for signature and info windows. If width_min isn't explicitly set, the default behavior is retained.

```lua
  window = {
    info = {... width_min = 6 },
    signature = { ... width_min = 10 },
  },
```

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -303,6 +303,7 @@ MiniCompletion.config = {
 
   -- Configuration for action windows:
   -- - `height` and `width` are maximum dimensions.
+  -- - `width_min` is the minimum width dimension.
   -- - `border` defines border (as in `nvim_open_win()`; default "single").
   window = {
     info = { height = 25, width = 80, border = nil, width_min = nil },


echasnovski:
Thanks for the PR!

I am afraid, this is a bit too much for 'mini.completion'. Otherwise it opens a pass to a slippery slope of various dedicated small scoped options.

Currently 'mini.completion' provides dedicated [events](https://nvim-mini.org/mini.nvim/doc/mini-completion.html#minicompletion-events) which should be enough to customize information and signature windows in any way users prefer. Including setting minimum width.

Closing as not planned.

gx089:
Okay, I tried to figure out how to achieve this via builtin methods like `events` you mentioned.
Let's assume I want to set the min width for the info window to 6 for example. 
Can you please provide some hint to me how this can be done. I couldn't figured it out how to do this without changing the the underling core lua code.

Thanks in advance @echasnovski  :-)

Screenshot_20260306_123951-1

echasnovski:
The general idea is to create an autocommand for the `User` event with documented patterns. Inside that autocommand: get current window width and possibly adjust it to be at least a certain value. Here is an example:

```lua
local set_min_width = function(ev)
  -- Ignore adjusting signature window
  if ev.data.kind == 'signature' then return end

  -- Get current width and possibly adjust to be at least 20
  local win_id = ev.data.win_id
  local config = vim.api.nvim_win_get_config(win_id)
  -- Also adjust title since it might have been truncated
  local config_updates = { width = math.max(config.width, 30), title = 'Info' }
  vim.api.nvim_win_set_config(win_id, config_updates)
end
local pattern = { 'MiniCompletionWindowUpdate', 'MiniCompletionWindowOpen' }
vim.api.nvim_create_autocmd('User', { pattern = pattern, callback = set_min_width })
```

One possible downside of this approach is that it might interfere with computation of which side of popup menu to use to show the info/signature window. I.e. if there is small room to the right of it and the width computed by 'mini.completion' fits while the desired minimum one does not. So I'd recommend using not too big value for minimum width.

gx089:
Thank you for your effort writing this snippet. 

It solves the problem but partially. If the width of the (float)window exceeds the rest of the space on the screen, the window is gets positioned on the left side (instead of right side) of the completion menu. This is a normal and expected behavior.
But the window doesn't align well - one part of the window is covered under the completion menu.

Nevertheless, thank you for clarification and providing this hint. :-)
On the weekend I will spend some time digging more into it.
At the moment I go the easy way and use a local copy of mini.completion with those changes. Works perfectly for my needs without any interference, overlapping etc. 


Btw.: The hyperlink of "User" is referencing to a 404 Page
https://nvim-mini.org/mini.nvim/doc/mini-completion.html#minicompletion-events

echasnovski:
> It solves the problem but partially. If the width of the (float)window exceeds the rest of the space on the screen, the window is gets positioned on the left side (instead of right side) of the completion menu. This is a normal and expected behavior.
> But the window doesn't align well - one part of the window is covered under the completion menu.

As I said - not perfect. But can be adjusted and probably relatively concisely.

> Btw.: The hyperlink of "User" is referencing to a 404 Page
> https://nvim-mini.org/mini.nvim/doc/mini-completion.html#minicompletion-events

Thanks. This is a known neovim.io problem which broke the expected way of linking to a tag. There are solutions in pending PRs, not merged yet.

Issue #688, opened 2024-02-10
Jump consumes characters when it cannot find the target


ghostbuster91:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated 'mini.nvim' to latest version

### Module(s)

Jump

### Description

Hi, for the context, I am the author of https://github.com/ghostbuster91/nvim-next

When trying to implement dot-repeat for my custom movements (`f`/`t`/etc) and failing miserably many times I finally came across your repository where it seems to work. Hats off to your vim-fu. 

However I noticed one bug - MiniJump consumes one character when using custom `f` motion with operator-pending and when the target does not exist. I tried fixing it myself but without luck.
I even tried deleting the whole body of the function that is being passed to the expr but the character was still being consumed.

### Neovim version

0.9.5

### Steps to reproduce

I changed `jump.lua` file so that it would invoke the setup when sourced:
```diff
diff --git a/jump.lua b/jump.lua
index d5b0c9c..d7a1959 100644
--- a/jump.lua
+++ b/jump.lua
@@ -488,4 +488,6 @@ function H.map(mode, key, rhs, opts)
     vim.api.nvim_set_keymap(mode, key, rhs, opts)
 end
 
+MiniJump.setup()
+
 return MiniJump
```

1. `nvim -nu jump.lua`
2. open any file with some text inside
3. type `dfk` on a line where there is no `k` 

### Expected behavior

Nothing should happen when there is no target in the current line.

### Actual behavior

One character is consumed. The same happens when we repeat action with `.`

echasnovski:
Thanks for the issue!

I am not quite sure what is the essence of the issue.

'mini.jump' defines its own mappings for `f`, `F`, `t`, `T`, `;` making them work across multiple lines. Pressing `dfk` on a line where there is no "k" should delete text up until and including the next "k" in the lines below.

Would you mind clarifying what exactly you mean by "cosumes characters"? And, please, provide reproducible example which does not alters module's code.

ghostbuster91:
Sure, here is the minimal config to reproduce the issue:
```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
[...]

PR #2303, opened 2026-03-06
refactor(jump): only do extra search for target if there was no jump

abeldekat:
Details:

Instead of always checking if target exists in text, only do so if there was no jump.
Additional benefit: The code using the cursor to check if a jump occurred can be replaced by a more direct check

### DIFF
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -181,16 +181,8 @@ MiniJump.state = {
 MiniJump.jump = function(target, backward, till, n_times)
   if H.is_disabled() then return end
 
-  -- Ensure to undo "consuming a character" effect if there is no target found
-  -- Do it here to act on dot-repeat
-  local has_changed_cursor = false
-  local undo_no_move = function()
-    if not has_changed_cursor then vim.cmd('undo!') end
-  end
-  if MiniJump._is_expr then vim.schedule(undo_no_move) end
-
   -- Dot-repeat should not change the state, so save it to later restore
-  local is_dot_repeat = MiniJump._is_expr and not MiniJump._is_expr_init
+  local is_expr, is_dot_repeat = MiniJump._is_expr, MiniJump._is_expr and not MiniJump._is_expr_init
   MiniJump._is_expr, MiniJump._is_expr_init = nil, nil
   local state_snapshot = is_dot_repeat and vim.deepcopy(MiniJump.state) or nil
 
@@ -201,12 +193,6 @@ MiniJump.jump = function(target, backward, till, n_times)
     H.message('Can not jump because there is no recent `target`.')
     return
   end
-
-  -- Determine if target is present anywhere in order to correctly enter
-  -- jumping mode. If not, jumping mode is not possible.
-  local search_pattern = [[\V]] .. vim.fn.escape(MiniJump.state.target, [[\]])
-  local target_is_present = vim.fn.search(search_pattern, 'wn') ~= 0
-  if not target_is_present then return end
 
   -- Construct search and highlight pattern data
   local pattern, hl_pattern, flags = H.make_search_data()
@@ -227,28 +213,33 @@ MiniJump.jump = function(target, backward, till, n_times)
 
   -- Make jump(s)
   H.cache.n_cursor_moved = 0
-  local init_cursor_data = H.get_cursor_data()
   local was_jumping = MiniJump.state.jumping
   MiniJump.state.jumping = true
   if not was_jumping then H.trigger_event('MiniJumpStart') end
   H.trigger_event('MiniJumpJump')
 
+  local has_jumped = false
   for _ = 1, MiniJump.state.n_times do
-    vim.fn.search(pattern, flags)
+    local row = vim.fn.search(pattern, flags)
+    has_jumped = has_jumped or row > 0
   end
 
   -- Open enough folds to show jump
-  vim.cmd('normal! zv')
+  if has_jumped then vim.cmd('normal! zv') end
 
   -- Track cursor position to account for movement not caught by `CursorMoved`
   H.cache.latest_cursor = H.get_cursor_data()
-  has_changed_cursor = not vim.deep_equal(H.cache.latest_cursor, init_cursor_data)
 
   -- Restore the state if needed
-  if is_dot_repeat then
-    state_snapshot.jumping = true
-    MiniJump.state = state_snapshot
-  end
+  MiniJump.state = is_dot_repeat and state_snapshot or MiniJump.state
+  -- Final jumping should only be true if jump was performed or target is present
+  local search_pattern = [[\V]] .. vim.fn.escape(MiniJump.state.target, [[\]])
+  MiniJump.state.jumping = has_jumped or vim.fn.search(search_pattern, 'wn') ~= 0
+
+  -- Ensure to undo "consuming a character" effect if there is no target found
+  -- Do it here to also act on dot-repeat
+  if has_jumped or not is_expr then return end
+  vim.schedule(function() vim.cmd('undo!') end)
 end
 
 --- Make smart jump


abeldekat:
@echasnovski,

The solution applied to solve issue #688 in February 2024 had a side-effect regarding highlighting.

Relevant part of diff:

```diff

@@ -364,6 +365,10 @@ H.make_expr_jump = function(backward, till)
     if target == nil then return '<Esc>' end
     H.update_state(target, backward, till, vim.v.count1)

+    vim.schedule(function()
+      if H.cache.has_changed_cursor then return end
+      vim.cmd('undo' .. (vim.fn.has('nvim-0.8') == 1 and '!' or ''))
+    end)
     return 'v<Cmd>lua MiniJump.jump()<CR>'

```

Whilst the scheduled `undo` command restored the character, it also prevented the highlighting.

> mark

Using its previous commit '1d49300d50a2'

- Open file, position cursor on the 'a'
- Type `dfm`
- Notice that the 'a' disappears, cursor is on 'r'
- Notice that 'm' is highlighted

Using the fix, commit '4f69339'

- Open file, position cursor on the 'a'
- Type `dfm`
- Notice that text is not modified, as expected
- Notice that m is not highlighted
- Type `.`
- Notice that 'm' is highlighted, as dot-repeat has no scheduled `undo`

To restore the original highlighting:

```diff
diff --git a/lua/mini/jump.lua b/lua/mini/jump.lua
index 234510b9..7bf50dd6 100644
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -239,7 +239,8 @@ MiniJump.jump = function(target, backward, till, n_times)
   -- Ensure to undo "consuming a character" effect if there is no target found
   -- Do it here to also act on dot-repeat
   if has_jumped or not is_expr then return end
-  vim.schedule(function() vim.cmd('undo!') end)
+  -- vim.schedule(function() vim.cmd('undo!') end)
+  vim.api.nvim_create_autocmd('ModeChanged', { once = true, callback = function() vim.cmd('un
do!') end })
 end

 --- Make smart jump
```

I think it also is more precise to use the `ModeChanged` event.

echasnovski:
Thanks for the PR!

I'll take a closer look a bit later (probably next week), as I can not quickly see if the new approach is optimal. Should be, as tests pass and it is negative diff :muscle:

> To restore the original highlighting:
> ...
> I think it also is more precise to use the `ModeChanged` event.

I don't really understand why `ModeChanged` event is better or even why it works. Initial reaction here is that executing the `undo!` command without `vim.schedule` might be dangerous due to how Neovim works (`:h textlock`, `:h :map-expression`, and similar things).

abeldekat:
> I don't really understand why ModeChanged event is better or even why it works. Initial reaction here is that executing the undo! command without vim.schedule might be dangerous due to how Neovim works (:h textlock, :h :map-expression, and similar things).

Me neither. I do think that the 'undo!' is the culprit here. If the command is deferred instead of scheduled, the highlighting is removed when the defer activates.

In addition to 'ModeChanged', 'CursorMoved' also works.

I also tried with 'nvim_feedkeys' and the escape character and did not succeed. In flash.nvim I found a [snippet](https://github.com/folke/flash.nvim/blob/fcea7ff883235d9024dc41e638f164a450c14ca2/lua/flash/util.lua#L18) that does work. The diff with that code applied:

```diff
diff --git a/lua/mini/jump.lua b/lua/mini/jump.lua
index 234510b9..8fc8cb0e 100644
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -239,7 +239,11 @@ MiniJump.jump = function(target, backward, till, n_times)
   -- Ensure to undo "consuming a character" effect if there is no target found
   -- Do it here to also act on dot-repeat
   if has_jumped or not is_expr then return end
-  vim.schedule(function() vim.cmd('undo!') end)
+
+  -- vim.schedule(function() vim.cmd('undo!') end)
+  local t = function(str) return vim.api.nvim_replace_termcodes(str, true, true, true) end
+  vim.api.nvim_feedkeys(t('<C-\\><C-n>'), 'nx', false)
+  vim.api.nvim_feedkeys(t('<esc>'), 'n', false)
 end

 --- Make smart _jump_
```

The disadvantage of that code is that test "can be dot-repeated if did not jump at first" breaks. I think one can argue that dot-repeating an operation that did not occur lacks intuitiveness. It's not supported by builtin dot-repeat. Surprisingly, flash is able to dot-repeat when initially the jump was not possible. Flash is [not](https://github.com/folke/flash.nvim/blob/fcea7ff883235d9024dc41e638f164a450c14ca2/doc/flash.nvim.txt#L71) using 'expr' mappings though...

echasnovski:
> I also tried with 'nvim_feedkeys' and the escape character and did not succeed. In flash.nvim I found a [snippet](https://github.com/folke/flash.nvim/blob/fcea7ff883235d9024dc41e638f164a450c14ca2/lua/flash/util.lua#L18) that does work. The diff with that code applied:

I would rather avoid using `feedkeys()` whenever possible. It proved to cause problems over time.

> The disadvantage of that code is that test "can be dot-repeated if did not jump at first" breaks. I think one can argue that dot-repeating an operation that did not occur lacks intuitiveness. It's not supported by builtin dot-repeat.

Yeah, indeed built-in dot-repeat does not work that way. The current behavior still makes more sense to me as it allows a smoother repeated actions like: `dtx` -> switch buffer -> `.` -> etc.

echasnovski:
Thanks again for the PR! It should now be a part of `main` after I made couple of small refactors that were bothering me for quite a while. A total diff of -16 for the  core function while passing all tests is a very good result!

> The solution applied to solve issue #688 in February 2024 had a side-effect regarding highlighting.

Regarding this issue, I am not able to reproduce the steps you describe (maybe anymore or maybe I wasn't able to in the first place, don't remember). With the latest change in how Operator-pending and dot-repeat works, the best approach would be to keep highlighting *only in the direction that dot-repeat will repeat*, but as far as I can tell there is no existing code for that while adding one would be not concise.

I was thinking about not highlighting when jumping in Operator-pending mode (since it is relatively easy to implement now), but indeed highlighting places where dot-repeat would take the cursor sounds useful. So not yet sure what to do here. One thing I am certain: I'd rather have any highlighting mismatch than add `feedkeys()` or `ModeChanged` autocommand workarounds.

abeldekat:
Thanks!

```lua

  if is_expr and not has_jumped then vim.schedule(function() vim.cmd('undo!') end) end

```

I had this originally and thought you'd like the undo statement outside the if better. But indeed, this is more expressive.

echasnovski:
> I had this originally and thought you'd like the undo statement outside the if better. But indeed, this is more expressive.

Usually - yeah, early return in this kind of situations is nice. But in this case it fits a single line (although pretty convoluted) and reads almost to the point of not needing the comment. Two lines here is more simple though, so the this decision can depend on time of day or day of the week :)

abeldekat:
https://github.com/user-attachments/assets/8997ed18-3474-47a4-bc94-927850cd9d4a

>> The solution applied to solve issue https://github.com/nvim-mini/mini.nvim/issues/688 in February 2024 had a side-effect regarding highlighting.

> Regarding this issue, I am not able to reproduce the steps you describe 

That's strange. See the video.

> With the latest change in how Operator-pending and dot-repeat works, the best approach would be to keep highlighting only in the direction that dot-repeat will repeat, but as far as I can tell there is no existing code for that while adding one would be not concise.

I agree

> So not yet sure what to do here. One thing I am certain: I'd rather have any highlighting mismatch than add feedkeys() or ModeChanged autocommand workarounds.

If the highlighting is a requirement, than perhaps the onetime CursorMoved event, as shown in the  video, can be considered. I think it might be an appropriate workaround, as indeed the cursor is moved when a character is consumed.

Shall I create a separate issue?

echasnovski:
> That's strange. See the video.

No highlighting after the first `dfm` I did reproduce. I could not reproduce the dot-repeat part (I saw again no highlighting, but the steps describe yes highlighting), which I thought was the issue.

> * Type `.`
> 
> * Notice that 'm' is highlighted, as dot-repeat has no scheduled `undo`

---

> If the highlighting is a requirement, than perhaps the onetime CursorMoved event, as shown in the video, can be considered. I think it might be an appropriate workaround, as indeed the cursor is moved when a character is consumed.

That is such a niche and undocumented behavior (`CursorMoved` trigger after "consuming" a character in Operator-pending mode if cursor did not move) that I am not fully confident there is a guarantee it will be preserved forever.

> Shall I create a separate issue?

Maybe that's a good idea, yeah. At least to (concisely) describe all this behavior. Please make a feature request that is more in terms of "adjust highlighting after jumping in Operator-pending mode", because this is the core problem here. Which my guess would require a fundamental solution: either by not highlighting, or rethinking the whole way of doing expression mapping jumping (and the `undo!` that it requires).

PR #2310, opened 2026-03-11
test(ALL): fix tests that fail when using git worktree

abeldekat:

Details(editted):
- When using git worktree, the mini.nvim repository is also checked-out under a different name. That causes some tests to fail.

- test_tabline: The tabline in test 'deduplicates named labels' changes when the cwd is not 'mini.nvim'. Solution: increase child.o.columns

- test_snippets: The screenshot in 'Session persists after :edit' assumes that the name of the repro is mini.nvim. Solution: use a dummy statusline.

### DIFF
--- a/tests/screenshots/tests-test_snippets.lua---Session---persists-after-`-edit`
+++ b/tests/screenshots/tests-test_snippets.lua---Session---persists-after-`-edit`
@@ -5,7 +5,7 @@
 4|~                                       
 5|~                                       
 6|~                                       
-7|</mini.nvim/tests/dir-snippets/tmp 1,4-5
+7|Dummy statusline                        
 8|-- INSERT --                            
 
 -|---------|---------|---------|---------|
--- a/tests/test_snippets.lua
+++ b/tests/test_snippets.lua
@@ -3550,6 +3550,9 @@ T['Session']['persists after `:edit`'] = function()
   child.cmd('edit')
   sleep(small_time)
 
+  -- Ensure expect_screenshot succeeds regardless of actual path
+  child.o.statusline = 'Dummy statusline'
+
   -- Should preserve both highlighting and data
   validate_active_session()
   child.expect_screenshot()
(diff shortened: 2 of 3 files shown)


echasnovski:
Could you please elaborate on what is the problem here? How can I reproduce the test fails?

All tests (implicitly) assume that they are run from inside the project's root directory. Usually it is called 'mini.nvim', but with Git worktrees it can have another name and that is the problem, right? If yes, then no repro steps needed. Just please make sure that after this PR all tests pass in this scenario (preferably on all supported Neovim versions, but that should be okay if works on latest release).

---

Please:
- Use capital letter at the start of the comment.
- Prefer using `child.fn.chdir('xxx')` instead of `child.cmd('cd xxx')`.
- Avoid using placeholder words like "foo", "bar", etc.

Also, I am not convinced that changing a directory to something that doesn't exist is a good idea. Could you please find another (concise) way to make 'mini.tabline' test pass in this case?

abeldekat:
> All tests (implicitly) assume that they are run from inside the project's root directory. Usually it is called 'mini.nvim', but with Git worktrees it can have another name and that is the problem, right?

Yes, that's the problem. I have 'mini.nvim' in the 'projects/plugin' folder, alongside 'mini_fix_x' and 'mini_test_fix_x'

abeldekat:
I pushed a new version

> Also, I am not convinced that changing a directory to something that doesn't exist is a good idea. Could you please find another (concise) way to make 'mini.tabline' test pass in this case?

I don't understand unfortunately. The directory in `tests/dir_tabline/` does exist, in order to have a fixed `cwd`, containing as many characters as the string 'mini.nvim'.
I renamed to `dir_other`, only containing a dummy file.

echasnovski:
> I don't understand unfortunately. The directory in `tests/dir_tabline/` does exist, in order to have a fixed `cwd`, containing as many characters as the string 'mini.nvim'.

Sorry, I missed that this PR adds the new directory. I was confused that there is a `child.fn.chdir()` into a non-existent directory.

I think this is better be solved by increasing `child.o.columns` to a very big number (with a comment about why this is done). No need for a new directory with the same number of letters.

abeldekat:
> I think this is better be solved by increasing child.o.columns 

Much better indeed. Done!

abeldekat:
> @abeldekat, did you make sure that this is...

Yes. I also grepped the screenshots searching for `mini.nvim` in the statusline...)

PR #2312, opened 2026-03-12
feat(files): Add line number display configuration

Aristosis:

Provide option to show configurable line numbers for the
active/focused file window in `windows.line_numbers`
- `windows.line_numbers.enabled` boolean
- `windows.line_numbers.relative` boolean, whether line numbers are in the style of the `relativenumber` option

I tried to assess where this would best fit in, as in some places I tried there was an ugly one draw delay before the numbers appeared. Hope this is good

https://github.com/user-attachments/assets/ecd2b62c-b36a-4e60-b08b-6ea782ae34b4

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -603,6 +603,12 @@ Defaults ~
 
     -- Customization of explorer windows
     windows = {
+      -- Enable/customize line numbers
+      line_numbers = {
+        enabled = false,
+        -- Whether to use relative numbers
+        relative = false,
+      },
       -- Maximum number of windows to show side by side
       max_number = math.huge,
       -- Whether to show preview of file/directory under cursor


abeldekat:
Hi Aristosis, 

I think it is already possible to set those options in the 'MiniFilesWindowUpdate' event.

In my config the linenumber options are set as follows:

```lua

 local restrict_linenumbers = function(args) -- add linenumbers only to active window
    local win_id = vim.api.nvim_get_current_win()
    if win_id and win_id == args.data.win_id then
      vim.wo[win_id].relativenumber = true
      vim.api.nvim_create_autocmd('WinLeave', {
        once = true,
        callback = function() vim.wo[win_id].relativenumber = false end,
      })
    end
  end
  Config.new_autocmd('User', 'MiniFilesWindowUpdate', restrict_linenumbers, 'Restrict linenumbers')

```

echasnovski:
Thanks for the PR!

This is a bit too much added configuration. This is already possible with available `MiniFilesWindowUpdate` `User` event. Like this:

```lua
local add_focused_linenumber = function(ev)
  local win_id = ev.data.win_id
  vim.wo[win_id].number = win_id == vim.api.nvim_get_current_win()
end
vim.api.nvim_create_autocmd('User', { pattern = 'MiniFilesWindowUpdate', callback = add_focused_linenumber })
```

This approach is similar to the one described in ["Customizing windows"](https://nvim-mini.org/mini.nvim/doc/mini-files.html#minifiles-examples-customizewindows) example.

Closing as not planned.

Aristosis:
> This is a bit too much added configuration. This is already possible with available `MiniFilesWindowUpdate` `User` event. Like this:

Hi! That approach results in the aforementioned one frame delay which can look really ugly... I understand the philosophy behind mini but I think this would be very fitting as it is very similar to a vim buffer
I'm sure you've given this enough thought to provide a recipe, but I think its still probably valuable considering one of the people that replied uses it in their config. Thanks a lot

Aristosis:
Here is a video of what that issue looks like. I agree that things like this can open up a floodgate of allowing several tiny tweaks but this is extremely simple 🫪

https://github.com/user-attachments/assets/f8628cdd-8da9-4ced-9d34-6560a0e9edc4

Edit: forgot to attach the video

echasnovski:
I do occasionally see the flicker, but it is by far not frequent. Here is the demo in which I don't think I've seen a flicker even once:

https://github.com/user-attachments/assets/4d9cbad6-2575-4738-bf33-193559dbc6d8

The presence of flicker probably has to do with machine being either fast or slow (not sure what is the case here).

But in any case, I am afraid the config from this PR is still too much to add.

---

There is a very vague plan to refactor 'mini.files' to allow any layout, in which case this kind of change should be more robust to add. But the plan hasn't crystallized yet, unfortunately.

PR #2324, opened 2026-03-20
fix(operators): remove new default `grx` LSP mapping

abeldekat:
Details

- New builtin lsp mapping "grx" executes `vim.lsp.codelens.run()`

  See PR 37689 in neovim/neovim

  This causes test 'Replace works with `cmdheight=0`' to fail:

  Can not use `child.get_screenshot` because child process is blocked.

  In mini.operators, default `gr*` LSP mappings are removed

### DIFF
--- a/lua/mini/operators.lua
+++ b/lua/mini/operators.lua
@@ -729,6 +729,7 @@ H.apply_config = function(config)
       remove_lsp_mapping('n', 'grn')
       remove_lsp_mapping('n', 'grr')
       remove_lsp_mapping('n', 'grt')
+      remove_lsp_mapping('n', 'grx')
     end
 
     if prefix == 'gx' and vim.fn.has('nvim-0.10') == 1 then
--- a/tests/test_operators.lua
+++ b/tests/test_operators.lua
@@ -135,6 +135,7 @@ T['setup()']['removes built-in LSP mappings'] = function()
   eq(child.fn.maparg('gri'), '')
   eq(child.fn.maparg('grn'), '')
   eq(child.fn.maparg('grt'), '')
+  eq(child.fn.maparg('grx'), '')
 end
 
 T['setup()']['remaps built-in `gx` mappings'] = function()


abeldekat:
Shall I add the new mapping to MiniMax as `<leader>lx`?

echasnovski:
Thanks for the PR!

Could you please more closely follow the 6a6a37ec54f833e2a733f5f04959f506520701ba? I.e. almost the same set of changes and the almost the same commit message.

---

> Shall I add the new mapping to MiniMax as `<leader>lx`?

This is a bit early for MiniMax. Let's wait until the Neovim 0.12 release.

Besides, I am not sure that `<Leader>lx` is a good LHS. I would expect something like `<Leader>lc` (*C*odelens) or `<Leader>ll` (*L*ens; similar to how `<Leader>la` is "*A*ctions").

abeldekat:
> Could you please more closely follow the https://github.com/nvim-mini/mini.nvim/commit/6a6a37ec54f833e2a733f5f04959f506520701ba? I.e. almost the same set of changes and the almost the same commit message.

Done

echasnovski:
> > Shall I add the new mapping to MiniMax as `<leader>lx`?
> 
> This is a bit early for MiniMax. Let's wait until the Neovim 0.12 release.
> 
> Besides, I am not sure that `<Leader>lx` is a good LHS. I would expect something like `<Leader>lc` (_C_odelens) or `<Leader>ll` (_L_ens; similar to how `<Leader>la` is "_A_ctions").

I've only now realized that `vim.lsp.codelens.run()` is not new and is present at least since Neovim>=0.10. So adding it to MiniMax can be done now. And I think the `<Leader>ll` fits more into the current language mapping convention.

But at the same time, I've never really used codelens, so there is no mapping for it already. Do you know any particularly good use for it? Enough to warrant the extra `<Leader>ll` mapping?

abeldekat:
> But at the same time, I've never really used codelens, so there is no mapping for it already. Do you know any particularly good use for it? Enough to warrant the extra <Leader>ll mapping?

No, I have never used it as well. I do think it makes sense for MiniMax to provide alternatives for builtin `lsp` mappings that are deleted.

PR #2332, opened 2026-03-28
fix(jump): do not force charwise Operator-pending jump

### LINKED ISSUE #2323: [mini.jump] Improve highlighting after jumping in Operator-pending mode
### Contributing guidelines

### Module(s)

mini.jump

### Description

This feature request is based on [this](https://github.com/nvim-mini/mini.nvim/pull/2303#issuecomment-4078015984) comment

-- edited --

> mark

- Open file, position cursor on the 'a'
- Type `dfm`
- Notice that text is not modified, as expected
- Notice that m is *not* highlighted

Notably, highlighting does work as expected using the same scenario with different file contents: Empty first line, mark is on the second with cursor on its a

abeldekat:
Details:
- ~~Always~~ highlight matches even when the Operator-pending jump did not succeed. This can happen if the jump is not possible, but there are matches in the other direction.

- The internal `undo` workaround is not needed anymore.

- ~~This change is breaking for users that test `MiniJump.state.mode` to be equal to 'nov'. In that situation, the mode is now 'no'.~~

Resolve #2323

### DIFF
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -205,6 +205,9 @@ MiniJump.jump = function(target, backward, till, n_times)
   H.timers.idle_stop:stop()
   H.timers.idle_stop:start(config.delay.idle_stop, 0, vim.schedule_wrap(function() MiniJump.stop_jumping() end))
 
+  -- Force charwise-visual in Operator-pending expression mapping
+  if is_expr then vim.cmd('normal! v') end
+
   -- Make jump(s)
   H.cache.n_cursor_moved = 0
   local was_jumping = MiniJump.state.jumping
@@ -224,15 +227,16 @@ MiniJump.jump = function(target, backward, till, n_times)
   -- Track cursor position to account for movement not caught by `CursorMoved`
   H.cache.latest_cursor = H.get_cursor_data()
 
-  -- Restore the state if needed. It should a jumping state if there was jump
+  -- Restore the state if needed. The jumping state is applied if there was jump
   -- or if it is possible to jump in other direction (i.e. target is present).
   MiniJump.state = is_dot_repeat and state_snapshot or MiniJump.state
   local search_pattern = '\\V' .. vim.fn.escape(MiniJump.state.target, '\\')
   MiniJump.state.jumping = has_jumped or vim.fn.search(search_pattern, 'wn') ~= 0
 
-  -- Ensure to undo "consume a character" effect in Operator-pending expression
-  -- mapping if there is no target found. Do it here to also act on dot-repeat.
-  if is_expr and not has_jumped then vim.schedule(function() vim.cmd('undo!') end) end
+  -- If target not found in Operator-pending expression mapping, charwise-visual
+  -- is reverted, preventing a character from being consumed.
+  -- Do it here to also act on dot-repeat.
+  if is_expr and not has_jumped then vim.cmd('normal! v') end
 end
 
 --- Make smart jump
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

As this stands, I don't think it is a `feat(jump)!`, but a regular `fix(jump)`:

- Resolves some edge cases that were there due to `:undo` (like in a new test for empty line).
- The `no` mode is a more correct mode anyway, since `nov` is a *forced* charwise mode (which should be the case for something like `dv`-`f`) which was a byproduct of me not finding another way to implement the intended jumping behavior (since it requires forced charwise movement to match the found patterns).
- It doesn't seem to break any explicitly documented behavior.

Also:

- I did notice that #2323 does not fully describe the behavior: there are problems if `mark` is the first line, but everything seems to work as expected (and as after this PR's change) if: the first line is empty, `mark` is on the second with cursor on its `a`.
- The "highlight all matches after Operator-pending mode jump" is not something that can be considered a feature. Ideally it would be highlighting only the matches that can be used to dot-repeat. Adding this functionality would be a `feat()`, but I don't think it can be done concisely. Highlighting all matches doesn't have much utility since it is not possible to change direction or `till` "on the fly" for dot-repeat.

So describing this as a `fix(jump): do not force charwise Operator-pending jump` is closer to what this achieves.

---

I noticed one problem with this change: highlighting disappears only after *second* unrelated cursor move if Operator-pending jump did not actually jump. I.e. with `mark` as a line and cursor on `a`: `dfm` - wait for highlight - `l` and it *still* highlights - `l`and it stops highlighting.

This might be not a problem if there is no highlighting after an Operator-pending jump, but currently there is. I am still on the fence about whether removing all highlighting is a good thing or a technical necessity.

abeldekat:
> I noticed one problem with this change: highlighting disappears only after second unrelated cursor move if Operator-pending jump did not actually jump. I.e. with mark as a line and cursor on a: dfm - wait for highlight - l and it still highlights - land it stops highlighting.

Indeed, that's a side effect of removing the 'undo'. On the first 'l', H.cache.n_cursor_moved is still 0....

I would like to fix that, not sure how at the moment.

abeldekat:
> I would like to fix that, not sure how at the moment.

Would the code below be acceptable?

```lua
  -- If target not found in Operator-pending expression mapping, charwise-visual
  -- is reverted, preventing a character from being consumed.
  -- Do it here to also act on dot-repeat.
  if is_expr and not has_jumped then
    H.on_cursormoved()
    vim.cmd('normal! v')
  end
```

echasnovski:
> Would the code below be acceptable?

Judging simply by the code - not quite. It basically reads as "If inside expression jump and *not* jumped act as on cursor *moved*".

abeldekat:
>  noticed one problem with this change: highlighting disappears only after second unrelated cursor move if Operator-pending jump did not actually jump. I.e. with mark as a line and cursor on a: dfm - wait for highlight - l and it still highlights - l and it stops highlighting.

In normal mode, a similar scenario also happens on latest main: With mark as a line and cursor on a: `fm` - wait for highlight - `l` and it still highlights - `l` and it stops highlighting.

Same cause: State is in jumping mode, but there was no jump, and thus no `CursorMoved`

I have an idea to fix this in the same manner for all modes. It's probably best to create a new PR after this one is merged.

abeldekat:
I changed the text of corresponding issue #2323 to be less verbose. Please let me know if anything else needs to be done before merging.

echasnovski:
This now should be part of the latest `main`.

The code looked good, I only tweaked a new "cursor on empty line" testing.

PR #2335, opened 2026-03-30
fix(jump): ensure one non-jump movement is enough to stop jumping

abeldekat:

Details:
- When the state is jumping, but there was no jump, two non-jump
  movements are needed to stop jumping.
  This can happen when the jump is not possible and there are targets
  in the other direction.

> mark

With cursor on a: 

- type fm 
- wait for highlight 
- type l and it still highlights 
- l and it stops highlighting.

Cause: State is in jumping mode, but there was no jump, and thus no CursorMoved

This PR builds upon the commit proposed in #2332: fix(jump): do not force charwise Operator-pending jump#2332

### DIFF
--- a/lua/mini/jump.lua
+++ b/lua/mini/jump.lua
@@ -205,6 +205,9 @@ MiniJump.jump = function(target, backward, till, n_times)
   H.timers.idle_stop:stop()
   H.timers.idle_stop:start(config.delay.idle_stop, 0, vim.schedule_wrap(function() MiniJump.stop_jumping() end))
 
+  -- Force charwise-visual in Operator-pending expression mapping
+  if is_expr then vim.cmd('normal! v') end
+
   -- Make jump(s)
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

The change looks reasonable. I've applied the relevant commit manually and tweaked a test case a bit (see 90c402eaf6c923b0c169b7325d7143a8766e8905). What I've come to realize (some time after writing all 'mini.jump' tests) is that validating conditions inside a `for` loop for different cases is not really convenient in the long term. If something fails, it might be hard to figure out which case broke since traceback will point to a line inside a loop. So (at least for now) writing a single `validate()` helper and then calling it on separate lines for different inputs is the preferred approach.

Another approach would be to have parametrized tests, of course, but it will inflate the number of test cases without much profit. It should be preferred if resetting to a clean state is either very hard or impossible.

abeldekat:
Great, thank you!

Issue #2215, opened 2026-01-07
Make 'mini.files' file system actions LSP aware


yadav-deepakk:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/nvim-mini/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/nvim-mini/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.files

### Description

Currently when we do editing in a directory structure through `mini.files` , it is treated like directory structure change only, but if it is a project(like java-based/ spring-boot/react) then code refactoring is also needed. Manual change of import statements, class names are hectic task for maintaining project. So idea is to have a lsp renaming whenever there is change in directory/file names, so that refactoring burden for programmer gets reduced. 

or Just correct me if mini.files has this functionality and I have skill issue at my end.

echasnovski:
Thanks for the suggestion!

It was discussed in #400, #1586, and #2025. See [this comment](https://github.com/nvim-mini/mini.nvim/discussions/2025#discussioncomment-14515019) for the most up-to-date info.

As it is planned (but with a low-ish priority), let's leave this one open.

TheLeoP:
After a small research I've found that:

## Basedpyright - willRenameFiles

- `willRenameFiles` -> fail -> `willRenameFiles` doesn't work to revert the text edits from the first `willRenameFiles`.
- after the first `willRenameFiles`, even when the rename succeeded, the language server seems to be left in a broken state (it seems the modified import statement as an error). ~~This doesn't happen if the renamed file was previously opened in Neovim (my guess is that Basedpyright behaves like this because the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename) mentions that renaming a file is signaled to a language server by a document close notification (with the old file name) followed by a document open notification (with the new file name).) This seems to be a bug on the Basedpyright side, but I think that the LSP spec may not be specific enough on how to handle renaming files.~~ Making this work seems to be more complicated that what I originally thought, now I can't reliably reproduce the state where the language server is not left in a broken state.

### Filter

```lua
{ {
  pattern = {
    glob = "**/*",
  },
} }
```
## lua_ls - didRenameFiles

- `didRenameFiles` -> fail -> `didRenameFiles` works to revert the text edits from the first `didRenameFiles`.
[...]

PR #2340, opened 2026-04-02
feat(mini.files): make file system actions LSP aware

TheLeoP:

Addresses #2215

This is a proof of concept to test the idea of including this kind of LSP support inside of mini.files. It may not be merged at all or it may help to add new autocmds to allow users/other plugins to implement this themselves.

A few things to take into account

- the control flow of `H.fs_do.delete` had to be changed to allow checking `success` a single time before triggering `didDeleteFile` notifications (actually, it seems like I broke something)
- a big part of the code simply filters the files before sending the requests/notifications to a language server as specified by the LSP spec (and, honestly, that logic should probably be extracted into a single function)
- for now, the `timeout` for `client:request_sync` is harcoded
- the code should make some considerations to support older Neovim versions (`vim.glob.to_lpeg` [has a bug prior to 0.11](https://github.com/neovim/neovim/issues/28931) that requires manually sorting items inside brackets as a workaround, `client:request_sync`/`client:notify` are not methods prior to 0.11, etc)
- I though about checking the user permissions on a given directory to further test if the file operations will succeed, but the file permissions semantics for these kind of operations seem to be different in Linux and Windows (and, as far as I can tell, `libuv` does not offer an abstraction on top of it. Maybe `fs_stat` related function would work on Windows with the same semantics, I haven't tested it yet)


TheLeoP:
To test the PR using `lua_ls` you can have a project with the following structure

```
test_project/
|- main.lua
|- something.lua
|- .git

.git is an empty file, I simply use it so `lua_ls` has a `root_marker`
```

with the following content

```lua
-- main.lua
local something = require("something").something

something()

-- something.lua
local M = {}

M.something = function()
	print("something")
end

return M
```
you can then `nvim main.lua`, `: lua MiniFiles.open()`m rename `something.lua` to `something_else.lua` and accept the changes with `=`. Neovim should prompt you for confirmation before updating the import statement on `main.lua` to `local something = require("something_else").something`

TheLeoP:
The tests errors seem to go away if I comment out

https://github.com/nvim-mini/mini.nvim/blob/134f9ddcd0b7ea85dbd73c6a4d91db989b06e4d3/lua/mini/files.lua#L2898

and 

https://github.com/nvim-mini/mini.nvim/blob/134f9ddcd0b7ea85dbd73c6a4d91db989b06e4d3/lua/mini/files.lua#L2911

But I'm not sure why.

In particular, the problematic lines seem to be

https://github.com/nvim-mini/mini.nvim/blob/134f9ddcd0b7ea85dbd73c6a4d91db989b06e4d3/lua/mini/files.lua#L2754

and https://github.com/nvim-mini/mini.nvim/blob/134f9ddcd0b7ea85dbd73c6a4d91db989b06e4d3/lua/mini/files.lua#L2802

changing them to

```lua
local clients = {}
```

seems to also make the test pass.

Are these tests specially time sensitive? I can't think of another reason of why they would be failing

lua/mini/files.lua
```diff
@@ -2742,19 +2742,125 @@ H.fs_actions_apply = function(fs_actions)
...
+  -- TODO(TheLeoP): should we explicitly check permissions before willXxx
+  -- requests? It seems like Windows has different permissinon rules for
+  -- creating/deleting/moving files and `libuv` doesn't provide an absraction
+  -- for them
+  local lsp_params = {
+    files = { {
+      uri = vim.uri_from_fname(path),
+    } },
+  }
+  H.lsp_will_fs_do('create', lsp_params)
```
echasnovski: It seems that `willCreateFiles` is for files only, right? This will also send a request for a directory.

lua/mini/files.lua
```diff
@@ -2742,19 +2742,125 @@ H.fs_actions_apply = function(fs_actions)
...
+
+  -- TODO(TheLeoP): configurable timeout
+  local timeout = 1000
+  for _, client in ipairs(clients) do
+    local filters = client.server_capabilities.workspace.fileOperations[method].filters --[=[@as lsp.FileOperationFilter[]]=]
+    local matching_functions = {}
+    for _, filter in ipairs(filters) do
+      local ignore_case = filter.pattern.options and filter.pattern.options.ignoreCase
+      local glob = filter.pattern.glob
+      if ignore_case then glob = glob:lower() end
```
echasnovski: `string.lower()` works only for Latin characters. Using `vim.fn.tolower()` is more robust.

lua/mini/files.lua
```diff
@@ -2742,19 +2742,125 @@ H.fs_actions_apply = function(fs_actions)
   end
 end
 
+H.lsp_will_fs_do_methods = {
+  create = 'willCreate',
+  delete = 'willDelete',
+  rename = 'willRename',
+}
+
+H.lsp_will_fs_do = function(action, params)
```
echasnovski: I have so many questions (mostly due to LSP complexity, not only due to the PR itself) ...

1. It seems that it is possible to make requests/notifications in bulk for every three operation kinds. I'd say it is worth taking advantage of that: it is less talking back-and-forth between the client and the server, less computation based on the client capabilities, etc.
2. There is a lot of code duplication between two functions. Either moving that to a separate function is better. Or even probably just using a single function that has something like `if method:sub(1, 3) == 'did' then ... else ... end`.
3. I think `H.lsp_{will,did}_fs_do_method` is a bit too much. Maybe supplying a method directly when calling a function is more explicit.

So all in all, my first instinct is that adding all LSP related stuff (done in bulk) in the `H.fs_actions_apply()` (so that it is "`willXxx` requests", "action", "`didXxx` notifications") should be less complex and more concise. Yes, it will probably require an duplicating check about if the actions is predicted to be successful, but can be worked around later. This way it can also reuse computed parameters.

lua/mini/files.lua
```diff
@@ -2742,19 +2742,125 @@ H.fs_actions_apply = function(fs_actions)
...
+  local method = H.lsp_will_fs_do_methods[action]
+  local full_method = 'workspace/' .. method .. 'Files'
+  local clients = vim.lsp.get_clients({ method = full_method })
+
+  -- TODO(TheLeoP): configurable timeout
```
echasnovski: I don't think this is huge deal. The plan is to add something like boolean `config.options.lsp`, so that it can be disabled if there is a slow server. But not sure, maybe a table `config.lsp = { enable = true, ... }` is better. Just don't like another table-like option.

echasnovski (commented):
Thanks for the PR!

There are at least two good things: the LSP complexity seems to be understood and managed (the worst part for me :) ) and there is a room for making it more concise.

echasnovski:
> Are these tests specially time sensitive? I can't think of another reason of why they would be failing

I don't think so. Probably, some logic has been broken. Didn't read too much into the code.

> * the code should make some considerations to support older Neovim versions (`vim.glob.to_lpeg` [has a bug prior to 0.11](https://github.com/neovim/neovim/issues/28931) that requires manually sorting items inside brackets as a workaround, `client:request_sync`/`client:notify` are not methods prior to 0.11, etc)

Restricting this functionality to only Neovim<0.11 is fine.

> * I though about checking the user permissions on a given directory to further test if the file operations will succeed, but the file permissions semantics for these kind of operations seem to be different in Linux and Windows (and, as far as I can tell, `libuv` does not offer an abstraction on top of it. Maybe `fs_stat` related function would work on Windows with the same semantics, I haven't tested it yet)

I am okay with these checks for `willXxx` requests be on the "best effort" basis. It is the fault of the LSP specifications to have designed such a questionable (i.e. I don't understand why it is like this) approach :)

---

I'll also take a look into how to better handle the "close on lost focus" in 'mini.files'. It will probably have to be "don't close if inside non-normal buffer in a floating window" type of exception.

echasnovski:
> I'll also take a look into how to better handle the "close on lost focus" in 'mini.files'. It will probably have to be "don't close if inside non-normal buffer in a floating window" type of exception.

Okay, I've looked into this. Going the "don't close explorer if in floating window" goes against some of the use cases that originally prompted the tracking of lost focus. Like "I want to `nvim .` and open file picker" and in general opening a file picker when inside an explorer.

I think the best compromise here would be to temporarily adjust `vim.ui.select()` to set a temporary flag to ignore tracking lost cursor.

Something like this (as of the current state of this PR) seems to work:

```diff
diff --git a/lua/mini/files.lua b/lua/mini/files.lua
index 304e895f..fce1bbb7 100644
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -1540,7 +1540,7 @@ end
 H.explorer_track_lost_focus = function()
   local track = vim.schedule_wrap(function()
     local ft = vim.bo.filetype
-    if ft == 'minifiles' or ft == 'minifiles-help' then return end
+    if H.skip_track_lost_focus or ft == 'minifiles' or ft == 'minifiles-help' then return end
     local cur_win_id = vim.api.nvim_get_current_win()
     MiniFiles.close()
     pcall(vim.api.nvim_set_current_win, cur_win_id)
@@ -2753,6 +2753,15 @@ H.lsp_will_fs_do = function(action, params)
   local full_method = 'workspace/' .. method .. 'Files'
   local clients = vim.lsp.get_clients({ method = full_method })
 
+  local ui_select_orig = vim.ui.select
+  vim.ui.select = function(items, opts, on_choice)
+    H.skip_track_lost_focus = MiniFiles.get_explorer_state() ~= nil
+    ui_select_orig(items, opts, function(...)
+      H.skip_track_lost_focus, vim.ui.select = nil, ui_select_orig
+      on_choice(...)
+    end)
+  end
+
   -- TODO(TheLeoP): configurable timeout
   local timeout = 1000
   for _, client in ipairs(clients) do
```

One thing to extra consider is whether there will be any side effects if `vim.ui.select()` is not called as a result of the LSP request.

---

Thinking about it more, maybe it is a good idea to have this kind of mock in general for the whole duration of the explorer. Will also work nicely with something like 'mini.snippets' (which can use `vim.ui.select` to show matched snippets). Hmm... Need to think a bit more.

echasnovski:
> Thinking about it more, maybe it is a good idea to have this kind of mock in general for the whole duration of the explorer. Will also work nicely with something like 'mini.snippets' (which can use `vim.ui.select` to show matched snippets). Hmm... Need to think a bit more.

Yeah, ignoring tracking lost focus while `vim.ui.select()` (or `vim.ui.input()`) is active is probably a good idea anyway. Here is my [current WIP solution](https://github.com/nvim-mini/mini.nvim/compare/main...backlog).

I tried it with this PR and:
- With using 'mini.pick' for `vim.ui.select` renaming a single file works (with `lua_ls`), renaming two - does not. Because it tries to spawn two 'mini.pick' windows in parallel, which is not possible and there are cryptic errors. I think using `vim.ui.select()` for this kind of stuff in core is a mistake (especially with `coroutine` kind of magic). But at least this is a one more argument for bulk processing LSP requests.
- With 'snacks.nvim' for `vim.ui.select` renaming even a single just doesn't work. No LSP related adjustments are made.

TheLeoP:
Thanks for your feedback! I'll be a bit busy today and tomorrow, so I'll address your comments and I'll keep working on this PR on Sunday

echasnovski:
> * With using 'mini.pick' for `vim.ui.select` renaming a single file works (with `lua_ls`), renaming two - does not. Because it tries to spawn two 'mini.pick' windows in parallel, which is not possible and there are cryptic errors. I think using `vim.ui.select()` for this kind of stuff in core is a mistake (especially with `coroutine` kind of magic). But at least this is a one more argument for bulk processing LSP requests.

Soo... I spent some time investigating in the hopes of opening an issue in neovim/neovim to drop the `vim.ui.select()` + coroutine solution. Key takeaways:
- The reason for this approach was neovim/neovim#21026.
- Many consecutive calls that make `window/showMessageRequest` (which is for `LuaLS` is done after `workspace/didRenameFiles`) work with built-in `vim.ui.select()` **and** some simple async one. But it doesn't work with all currently popular `vim.ui.select()` implementations: 'mini.pick', 'snacks.picker', 'fzf-lua'. My guess is that all of them use coroutines themselves to have optimized behavior.
- I don't think Neovim core is going to change anything since it was a result of a wider change of running with coroutines. Hence I am not keen on spending countless hours arguing that for core.
- The best way forward looks to me like:
    - Make only a single LSP request/notify per method+sync pair. I.e. aggregate `create`, `rename`, `delete` actions and advertise them to the LSP servers once per synchronization. This will minimize issues with all `vim.ui.select` implementations.
    - ~~Figure out a way for 'mini.pick' to have two and more pickers "in the queue". Maybe only for `vim.ui.select()`. Right now it throws an error, which is not nice.~~ This should now work on the latest `main`.

---

Repro steps for what I did:

- Create even more minimal Lua project:

    <details><summary>Test Lua project</summary>

    'main.lua':

    ```lua
    local uuu = require('uuu')
    local vvv = require('vvv')
    ```

    'uuu.lua':

    ```lua
    return 'uuu'
    ```

    'vvv.lua':

    ```lua
    return 'vvv'
    ```

    </details>

- Create an 'init.lua' with a separate `NVIM_APPNAME`:

    <details><summary>~/.config/nvim-repro/init.lua</summary>

    ```lua
    -- vim.ui.select() implementation =============================================

    -- -- Custom implementation
    -- local n = 0
    -- vim.ui.select = function(items, opts, on_choice)
    --   local win_id
    --   local width = math.ceil(0.25 * vim.o.columns)
    --   local height = math.min(vim.o.lines, #items)
    --
    --   local buf_id = vim.api.nvim_create_buf(false, true)
    --   local lines = vim.tbl_map(opts.format_item or tostring, items)
    --   vim.api.nvim_buf_set_lines(buf_id, 0, -1, false, lines)
    --
    --   n = n + 1
    --   local win_opts = { relative = 'editor', row = n, col = n, height = height, width = width, border = 'single' }
    --   win_id = vim.api.nvim_open_win(buf_id, true, win_opts)
    --   vim.api.nvim_win_set_cursor(win_id, { 1, 0 })
    --
    --   -- Choose and close
    --   local close = function(idx)
    --     vim.api.nvim_win_close(win_id, true)
    --     vim.api.nvim_buf_delete(buf_id, { force = true })
    --     on_choice(items[idx], idx)
    --   end
    --
    --   local choose = function() close(vim.api.nvim_win_get_cursor(win_id)[1]) end
    --   vim.keymap.set('n', '<CR>', choose, { buffer = buf_id })
    --
    --   local quit = function() close() end
    --   vim.keymap.set('n', '<Esc>', quit, { buffer = buf_id })
    -- end

    -- From 'mini.pick'
    vim.pack.add({ 'https://github.com/nvim-mini/mini.nvim' })
    require('mini.pick').setup()

    -- -- From 'fzf-lua'
    -- vim.pack.add({ 'https://github.com/ibhagwan/fzf-lua' })
    -- require('fzf-lua').setup()
    -- require('fzf-lua').register_ui_select()

    -- -- From 'snacks.nvim'
    -- vim.pack.add({ 'https://github.com/folke/snacks.nvim' })
    -- require('snacks').setup({ picker = { enable = true } })

    -- Helpers for `workspace/didRenameFiles` notifications =======================
    _G.start_luals = function()
      local config = { name = 'lua-ls', cmd = { 'lua-language-server' }, root_dir = vim.fn.getcwd() }
      _G.client_id = vim.lsp.start(config)
    end

    local make_didrename_params = function(from_name, to_name)
      local cwd = vim.fs.abspath(vim.fn.getcwd())
      local old = 'file://' .. vim.fs.joinpath(cwd, from_name)
      local new = 'file://' .. vim.fs.joinpath(cwd, to_name)
      return { files = { { oldUri = old, newUri = new } } }
    end

    _G.make_one_notify = function()
      local client = vim.lsp.get_client_by_id(_G.client_id)
      client:notify('workspace/didRenameFiles', make_didrename_params('uuu.lua', 'uuu2.lua'))
    end

    _G.make_two_notify = function()
      local client = vim.lsp.get_client_by_id(_G.client_id)
      client:notify('workspace/didRenameFiles', make_didrename_params('uuu.lua', 'uuu2.lua'))
      client:notify('workspace/didRenameFiles', make_didrename_params('vvv.lua', 'vvv2.lua'))
    end

    _G.make_two_notify_simul = function()
      local cwd = vim.fs.abspath(vim.fn.getcwd())
      local old_1 = 'file://' .. vim.fs.joinpath(cwd, 'uuu.lua')
      local new_1 = 'file://' .. vim.fs.joinpath(cwd, 'uuu2.lua')
      local old_2 = 'file://' .. vim.fs.joinpath(cwd, 'vvv.lua')
      local new_2 = 'file://' .. vim.fs.joinpath(cwd, 'vvv2.lua')
      local params = { files = { { oldUri = old_1, newUri = new_1 }, { oldUri = old_2, newUri = new_2 } } }

      local client = vim.lsp.get_client_by_id(_G.client_id)
      client:notify('workspace/didRenameFiles', params)
    end
    ```

    </details>

- `NVIM_APPNAME=nvim-repro nvim -- main.lua uuu.lua vvv.lua` from the root of the test Lua project. Assumes `lua-language-server` is installed systemwide.
- `:lua start_luals()` and wait for it to load (diagnostic hints will be shown in sign column and statusline).
- `:lua make_one_notify()` and accept. It works. Revert its changes.
- `:lua make_two_notify_simul()` and accept. It works. Revert its changes.
- `:lua make_two_notify()`. It works only for built-in and custom minimal `vim.ui.select()`, but not for 'mini.pick' (throws error), 'fzf-lua' (only shows one selection window and acts only on one notification), 'snacks.picker' (just does nothing).

lua/mini/files.lua
TheLeoP: I pushed a couple of commits that should address all of these comments. The only difference is that instead of reusing the parameters for the LSP operations, I'm checking which operations succeeded for the `didXxx` notifications

lua/mini/files.lua
TheLeoP: It seems like [language servers should exclude file/directories if they are not interested in those](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#fileOperationPatternKind).

Edit: while looking this up, I realized that my current implementation of filters assumes that **all** filters should match a file for it to be considered, but it seems like **any** filter must match instead. This makes the implementation a bit more complex, let me know what you think

TheLeoP:
The bulk LSP request/notifications should be working now. I left them on separated commits to make reviewing a bit easier.

Let me know if you have any more feedback regarding the implementation. After that is done, I'll focus on creating some tests. What would be your preference for testing this? My guess is that it would be to mock a language server.

I can take a look at `mini.pick` later to see if I can do something to allow it to queue `vim.ui.select` calls.

lua/mini/files.lua
```diff
@@ -2734,10 +2734,37 @@ H.fs_actions_to_lines = function(fs_actions)
...
+  H.lsp_fs_hook('willCreate', {
+    files = vim.tbl_map(H.get_lsp_file_create_from_action, filtered_create),
+  })
+  H.lsp_fs_hook('willDelete', {
+    files = vim.tbl_map(H.get_lsp_file_delete_from_action, filtered_delete),
+  })
+  H.lsp_fs_hook('willRename', {
+    files = vim.tbl_map(H.get_lsp_file_rename_from_action, filtered_rename),
+  })
+
```
echasnovski: All these three and the `didXxx` notifications have the exact same logic, just different field names. I am fairly certain that there is no need for any `H.get_lsp_file_xxx_form_action` helpers. All their logic can be done at the top of `lsp_fs_hook(method, diffs)` with something like this:

```lua
H.lsp_fs_hook = function(method, diffs)
  -- Compute LSP params
  local files, to_uri = {}, vim.uri_from_fname
  for _, d in ipairs(diffs) do
    local file = {}
    file[d.to ~= nil and 'oldUri' or 'uri'] = d.from ~= nil and to_uri(d.from) or nil
    file[d.from ~= nil and 'newUri' or 'uri'] = d.to ~= nil and to_uri(d.to) or nil
    table.insert(files, file)
  end

  local params = { files = files }

  -- ...
```

lua/mini/files.lua
```diff
@@ -2734,10 +2734,37 @@ H.fs_actions_to_lines = function(fs_actions)
...
+  local successfull_create = {}
+  local successfull_delete = {}
+  local successfull_rename = {}
```
echasnovski: With shorter names, fitting them on the same line will be very readable.

lua/mini/files.lua
```diff
@@ -2734,10 +2734,37 @@ H.fs_actions_to_lines = function(fs_actions)
 end
 
 H.fs_actions_apply = function(fs_actions)
+  local filtered_create = vim.tbl_filter(
+    function(diff) return diff.from and diff.action == 'create' and not H.fs_is_present_path(diff.from) end,
```
echasnovski: This will not filter any create actions. The logic of `fs_actions` is like this:

- "Create" action [only has `to`](https://github.com/nvim-mini/mini.nvim/blob/402ee6c6ec8ea44b22330446c8fb4e615fd3953e/lua/mini/files.lua#L1697).
- "Delete" action [has meaningful `from`](https://github.com/nvim-mini/mini.nvim/blob/402ee6c6ec8ea44b22330446c8fb4e615fd3953e/lua/mini/files.lua#L1730) (the possible `to` when moving to trash is not relevant).
- "Rename" action has both `from` and `to` (as it is [copy](https://github.com/nvim-mini/mini.nvim/blob/402ee6c6ec8ea44b22330446c8fb4e615fd3953e/lua/mini/files.lua#L1701) plus delete).

So I don't think there should be any checks for `from` and `to` when there is a check for `action`. Otherwise, it should be fixed in diff computation.

lua/mini/files.lua
```diff
@@ -2749,6 +2776,94 @@ H.fs_actions_apply = function(fs_actions)
...
+  })
+end
+
+H.lsp_fs_hook = function(method, params)
+  local full_method = 'workspace/' .. method .. 'Files'
+  local clients = vim.lsp.get_clients({ method = full_method })
+
+  -- TODO(TheLeoP): configurable timeout
+  local timeout = 1000
+  for _, client in ipairs(clients) do
```
echasnovski, line 2852: I am sorry I don't have a more actionable direction for now, but could you please make it less nested (it is currently *way* too nested) and easier to read?

lua/mini/files.lua
```diff
@@ -2749,6 +2776,94 @@ H.fs_actions_apply = function(fs_actions)
...
+          local type = H.fs_get_type(path)
+          if filter.pattern.matches == 'folder' then return type == 'directory' end
+          if filter.pattern.matches == 'file' then return type == 'file' end
+        end)
+      end
+      table.insert(grouped_matching_functions, matching_functions)
+    end
+
+    local filtered_params = {
+      files = {},
```
echasnovski: Please don't create multiline tables if they can fit in a single line. `StyLua` won't reformat these into a single line if there are trailing commas, so these have to be done manually.

echasnovski (commented):
This is visibly better now, thanks! Added the next round of review.

lua/mini/files.lua
TheLeoP: Got it, thanks for the explanation! That extra check was because of a misunderstanding on my side when I got some test failures

lua/mini/files.lua
TheLeoP: Sorry, I'm used to always using multiline tables and I always forget this when contributing to mini.nvim. I'll take it into account

lua/mini/files.lua
TheLeoP: Would you be ok with me removing the extra `not H.fs_is_present_path(d.to)` checks for `willXxx` operations? Currently, my code looks something like

```lua
  local files, to_uri = {}, vim.uri_from_fname
  for _, d in ipairs(diffs) do
    local will_create = method:match('Create$')
      and d.action == 'create'
      and (not H.fs_is_present_path(d.to) or method:sub(1, 3) == 'did')
    local will_delete = method:match('Delete$') and d.action == 'delete'
    local will_rename = method:match('Rename$')
      and d.action == 'rename'
      and (not H.fs_is_present_path(d.to) or method:sub(1, 3) == 'did')

    if will_create then table.insert(files, { uri = to_uri(d.to) }) end
    if will_delete then table.insert(files, { uri = to_uri(d.from) }) end
    if will_rename then table.insert(files, { oldUri = to_uri(d.from), newUri = to_uri(d.to) }) end
  end
```

which seems not that easy to follow. Removing that extra check it could be

```lua
  local files, to_uri = {}, vim.uri_from_fname
  for _, d in ipairs(diffs) do
    if method:match('Create$') and d.action == 'create' then table.insert(files, { uri = to_uri(d.to) }) end
    if method:match('Delete$') and d.action == 'delete' then table.insert(files, { uri = to_uri(d.from) }) end
    if method:match('Rename$') and d.action == 'rename' then
      table.insert(files, { oldUri = to_uri(d.from), newUri = to_uri(d.to) })
    end
  end
```

echasnovski: Completely removing the check is not fine. Reading spec makes be believe that all `willXxx` requests should be done before actually performing an action. Right now it can make a request but not perform an action later. Reducing this type of bad behavior as much as reasonably possible is very much needed.

lua/mini/files.lua
```diff
@@ -2749,6 +2756,83 @@ H.fs_actions_apply = function(fs_actions)
...
+
+  H.lsp_fs_hook('didCreate', ok_actions)
+  H.lsp_fs_hook('didDelete', ok_actions)
+  H.lsp_fs_hook('didRename', ok_actions)
+end
+
+H.lsp_fs_hook = function(method, diffs)
+  local files, to_uri = {}, vim.uri_from_fname
+  for _, d in ipairs(diffs) do
+    if method:match('Create$') and d.action == 'create' then table.insert(files, { uri = to_uri(d.to) }) end
```
echasnovski: This indeed makes the usage of this function much cleaner, but at the cost of a more complex cycle and refusing to do the future success check for `willXXX` methods (which is very much needed).

This needs both simplification and more behavior. Is there a problem with going [this route](https://github.com/nvim-mini/mini.nvim/pull/2340#discussion_r3043924882) but to do the actual `table.insert()` step behind a single `if ... then ... end`?

echasnovski:
> * Figure out a way for 'mini.pick' to have two and more pickers "in the queue". Maybe only for `vim.ui.select()`. Right now it throws an error, which is not nice.

This should now work on the latest `main`. As well as calling `vim.ui.select()` or `vim.ui.input()` when inside 'mini.files' explorer.

lua/mini/files.lua
TheLeoP: Yeah, I was expecting that answer. I just wanted to be sure before continuing with the PR. Thanks!

lua/mini/files.lua
TheLeoP: > Is there a problem with going this route...

My only gripe with that approach is that I couldn't think of a way of distinguishing between `delete` (which may have both `from` and `to`) and rename by only relying on `from` and `to`. So, I tried a similar, but different approach:

```lua
    local file = {
      [d.action == 'rename' and 'oldUri' or 'uri'] = d.from and to_uri(d.from) or nil,
      [d.action == 'rename' and 'newUri' or 'uri'] = d.action ~= 'delete' and to_uri(d.to) or nil,
    }
```

> ... but to do the actual table.insert() step behind a single if ... then ... end?

If we want to use a single if statement, the condition will be a bit complex. I ended up with

```lua
  local files, to_uri = {}, vim.uri_from_fname
  for _, d in ipairs(diffs) do
    local file = {
      [d.action == 'rename' and 'oldUri' or 'uri'] = d.from and to_uri(d.from) or nil,
      [d.action == 'rename' and 'newUri' or 'uri'] = d.action ~= 'delete' and to_uri(d.to) or nil,
    }
    if
      (method == 'willCreate' and not H.fs_is_present_path(d.to))
      or (method == 'willRename' and not H.fs_is_present_path(d.to))
      or (method ~= 'willCreate' and method ~= 'willRename')
    then
      table.insert(files, file)
    end
  end
```

TheLeoP:
I just tested the latest changes on `main` and they work perfectly. Thanks!

lua/mini/files.lua
TheLeoP: I tried to simplify the code as much as possible without actually pulling out some of it into a helper function on the latest commit, let me know what you think. If the nesting is still annoying, I could try to use a helper function to make it more readable

lua/mini/files.lua
echasnovski: As is, this doesn't filter according to the input `method`. I'll push my refactor shortly.

echasnovski:
To spare both of us the back and forth about mine vague "make less nesting" suggestion, I've pushed what I think is a reasonable improvement (4b300bc6f4c402050d4ca26f3ee4bae30c8cb976). It did increase line count by 12 lines, but this is probably due to comments, new robustness checks, and performance improvements (mostly by avoiding doing things when they are not needed). This should work okay now.

Please double check the new code. If you notice an incorrect behavior, please point that out. After that - squash all commits before it.

---

Now the less fun part - tests. Yes, I'd like this to be done by an explicit new mocked LSP server (there are 'tests/mock-lsp' examples for reference). It essentially should test:
- All LSP actions are done in full when they are expected to be done. With appropriate parameters at appropriate time.
- No LSP actions are done if they are expected to be not done (like not fit any kind of filter, etc.).
- Works with several active LSP servers.

lua/mini/files.lua
echasnovski: It was better indeed. But spare us both more back and forth, I've spent some time simplifying to what I think is a reasonable simplicity (what that means exactly might change tomorrow, but oh well :) ). Pushed the commit directly to the PR branch.

lua/mini/files.lua
```diff
@@ -2782,6 +2789,86 @@ H.fs_actions_apply = function(fs_actions)
...
+  if #clients == 0 then return end
+
+  -- Transform 'mini.files' diffs into LSP file actions for the input method
+  local files, to_uri = {}, vim.uri_from_fname
+  local needs_check = method == 'willCreate' or method == 'willRename'
+  local is_create, is_delete, is_rename =
+    vim.endswith(method, 'Create'), vim.endswith(method, 'Delete'), vim.endswith(method, 'Rename')
+  for _, d in ipairs(diffs) do
+    -- Precompute LSP file type for filters (path can be not yet on disk)
+    local file = { fs_type = (d.from or d.to):find('/$') ~= nil and 'folder' or 'file' }
```
TheLeoP: This will always get overridden by the `file = {...}` inside of the if statements

echasnovski: Ah, yeah, indeed. Pushed a fixup.

TheLeoP:
For the `didXxx` notifications, would you prefer the mock LSP to react to them by:
- triggering a `window/showMessageRequest` request to the client and then sending a `workspace/applyEdit` request to the client (like `lua_ls` does)
- directly send a `workspaced/applyEdit` request
- some other option, like doing the changes inside of the mock LSP server by using the Neovim API directly

Also, for the tests with multiple LSP servers attached, you want them to be different servers that react to the `wilXxx`/`didXxx` LSP hooks, right?

TheLeoP:
I still have to implement the tests for the `didXxx` notifications and multiple LSP clients, but let me know what you think about the other two tests so far.

For the failing tests, I need to regenerate the snapshot and update the expected highlight groups, right? Also, is the `mock` directory the correct location for the `will.lua` file? I'm not confident in my understanding of the directory structure.

---

I also caught an error with my previous assumptions regarding `ignoreCase` while creating the tests and fixed it in the last commit :p

lua/mini/files.lua
```diff
@@ -2782,6 +2789,88 @@ H.fs_actions_apply = function(fs_actions)
...
+  local make_filter = function(scheme, ref_fs_type, glob, ignore_case)
+    if ignore_case then glob = vim.fn.tolower(glob) end
+    local glob_lpeg = vim.glob.to_lpeg(glob or '**')
+    return function(lsp_file)
+      local uri = lsp_file.uri or lsp_file.oldUri
+      local fname = vim.uri_to_fname(uri)
+      if ignore_case then fname = vim.fn.tolower(fname) end
+      return is_scheme(uri, scheme) and is_fs_type(lsp_file, ref_fs_type) and glob_lpeg:match(fname) ~= nil
+    end
+  end
```
echasnovski, line 2874:
```suggestion
  local make_filter = function(scheme, ref_fs_type, glob, ignore_case)
    local adjust_case = ignore_case and vim.fn.tolower or function(x) return x end
    local glob_lpeg = vim.glob.to_lpeg(adjust_case(glob) or '**')
    return function(lsp_file)
      local uri = lsp_file.uri or lsp_file.oldUri
      local fname = adjust_case(vim.uri_to_fname(uri))
      return is_scheme(uri, scheme) and is_fs_type(lsp_file, ref_fs_type) and glob_lpeg:match(fname) ~= nil
    end
  end
```

Yeah, nice find! Can be made a bit more concise, though.

tests/test_files.lua
```diff
@@ -5887,6 +5887,110 @@ T['Events']['`MiniFilesActionMove` triggers'] = function()
...
+
+  -- Mock server
+  child.cmd('luafile tests/mock-lsp/files.lua')
+
+  return file_path
+end
+
+T['LSP'] = new_set()
+
+T['LSP']['works with willCreate, willRename and willDelete'] = function()
```
echasnovski: Please split into one LSP method per case. If it requires a common validator, then extract a single `local validate_xxx = function() ... end` after `T['LSP']`.

lua/mini/files.lua
```diff
@@ -2782,6 +2789,88 @@ H.fs_actions_apply = function(fs_actions)
...
     end
   end
+
+  H.lsp_fs_hook('didCreate', ok_actions)
+  H.lsp_fs_hook('didDelete', ok_actions)
+  H.lsp_fs_hook('didRename', ok_actions)
+end
+
+H.lsp_fs_hook = function(method, diffs)
+  local full_method = 'workspace/' .. method .. 'Files'
```
echasnovski, line 2824: Could you please add a link to the relevant part of LSP spec? That describes both the request and filter structures.

lua/mini/files.lua
```diff
@@ -2782,6 +2789,88 @@ H.fs_actions_apply = function(fs_actions)
...
+
+    table.insert(filters, make_filter(scheme, matches, glob, ignore_case))
+  end
+
+  local params = { files = {} }
+  for _, f in ipairs(lsp_files) do
+    -- Should pass if there are no filters
+    local ok = #filters == 0
+    for _, filt in ipairs(filters) do
+      ok = ok or filt(f)
```
echasnovski, line 2894: After a brief reading of the LSP spec, I don't remember seeing any clarification that filters needs to be combined with `or` and not `and`. Could you please link (in the GitHub comment) where this is clarified?

tests/mock-lsp/files.lua
```diff
@@ -0,0 +1,82 @@
...
+      didCreate = _G.filter_configs,
+      didDelete = _G.filter_configs,
+      didRename = _G.filter_configs,
+    },
+  },
+}
+
+local make_will_request = function(method, changes)
+  return function(params)
+    table.insert(_G.lsp_requests, method)
```
echasnovski: Should this also log `params` instead of `_G.params` for a more reliable test?

tests/mock-lsp/files.lua
```diff
@@ -0,0 +1,82 @@
...
+  end
+end
+
+local requests = {
+  initialize = function(_) return { capabilities = capabilities } end,
+  shutdown = function(_) return nil end,
+
+  ['workspace/willCreateFiles'] = make_will_request('workspace/willCreateFiles', {
+    {
+      range = { start = { line = 0, character = 0 }, ['end'] = { line = 0, character = 0 } },
```
echasnovski: Maybe make them also configurable via global variable, like `filter_configs`?

echasnovski (changes requested):
That's a good start for the tests! 👍

echasnovski:
> For the `didXxx` notifications, would you prefer the mock LSP to react to them by:

Whatever you think will cover real world usage better without accounting for rare cases that are not required by the spec. My guess would be that `window/showMessageRequest` is one of those things. Sending an `workspaced/applyEdit` may as well also be one of those things. As far as I understand the spec, the notification just needs to be recorded by the server and later it needs to decide to do whatever it wants to do. So maybe even only testing that notification was received with proper data is enough.

> For the failing tests, I need to regenerate the snapshot and update the expected highlight groups, right?

If they are the result of the adding new entries to 'tests/dir-files' (which they indeed look like), then yeah.

> Also, is the `mock` directory the correct location for the `will.lua` file? I'm not confident in my understanding of the directory structure.

As this is a file that is used within 'mini.files' tests, then it should be inside a 'tests/dir-files' directory. Beyond that it is on "whatever feels right" basis. I'd put this in the 'tests/dir-files/lsp-files' directory.

> Also, for the tests with multiple LSP servers attached, you want them to be different servers that react to the `wilXxx`/`didXxx` LSP hooks, right?

Whatever reasonably covers the "there are multiple LSP clients attached to the buffer" cases. Maybe a separate test case that has several servers attached, each with different capabilities? Like to test that if the server doesn't support a method, then there is no request done for it.

lua/mini/files.lua
TheLeoP: The LSP spec, sadly, does not clarify this. That's why initially I though that filters had to be joined with an `and`. The [VSCode implementation](https://github.com/microsoft/vscode-languageserver-node/blob/85a9202a7778912161adc3b2915747096843cc01/client/src/common/fileOperations.ts#L146-L147) assumes that only one of the filters needs to match and early returns `true` if any of the filters matches. But, it would probably be worth opening an issue into the LSP spec github repo asking for clarification.

There is a similar issue already on the repo https://github.com/microsoft/language-server-protocol/issues/2203 , that clarifies some behavior that's not defined on the spec (it seems like an empty `filters` array should result in no files being matched, for example). But, it doesn't touch on the subject of if all filters should match or only one filter should match to consider that a file matches.

echasnovski: > The [VSCode implementation](https://github.com/microsoft/vscode-languageserver-node/blob/85a9202a7778912161adc3b2915747096843cc01/client/src/common/fileOperations.ts#L146-L147) assumes that only one of the filters needs to match and early returns `true` if any of the filters matches. But, it would probably be worth opening an issue into the LSP spec github repo asking for clarification.

Then at least for now using `or` is more useful. Since LSP specification often is adjusted based on the actual VSCode behavior. It is also reasonable based on [your early discovery](https://github.com/nvim-mini/mini.nvim/issues/2215#issuecomment-4174354212). For example, `typescript-tools` provides two filters: one for files and one for folders. It is not possible to match both of them.

> There is a similar issue already on the repo [microsoft/language-server-protocol#2203](https://github.com/microsoft/language-server-protocol/issues/2203) , that clarifies some behavior that's not defined on the spec (it seems like an empty `filters` array should result in no files being matched, for example).

Although "nothing matches if there is no filter" does sound cleaner, I doubt that it will be useful with the LSP servers. Although linked above earlier discovery does show that LSP servers actually provide filters, my guess is that not all of them do. Suggesting users to manually adjust server capabilities is fine, but I think it can be made useful out of the box.

Could you also add a link to that issue along with a concise description of why 'mini.files' doesn't do it that way? So that in the future it might be reverted to follow LSP spec (if there is an explicit clarification).

TheLeoP:
It took me longer than I expected, but the tests cover a bunch more use cases now. So, I marked the PR as ready for review.

---

I even ended up opening https://github.com/neovim/neovim/pull/39099 while looking into how to send server to client requests from an in-memory LSP :p.

Sorry for all the back and forth, btw.

TheLeoP:
I'm not sure why the tests on ubuntu with Neovim `0.10.4` are failling. It seems to be hitting https://github.com/neovim/neovim/issues/35550 , but that issue hasn't been solved as far as I can tell. So, I'm not sure why it would only happen on `0.10.4`.

I'm also not sure why Windows is not triggering the `willCreate` hook for the file `aaa`. Will take a look tomorrow

tests/mock-lsp/file-ops.lua
```diff
@@ -0,0 +1,76 @@
...
+  },
+}
+
+_G.did_callback = _G.did_callback or function(params, dispatchers) end
+
+local make_will_request = function(method)
+  return function(params)
+    table.insert(_G.lsp_requests, method)
+    table.insert(_G.request_params, params)
+    local path = vim.fn.fnamemodify('tests/dir-files/lsp-files/main.lua', ':p')
```
echasnovski: The idea of separate LSP mocks is that they should not depend on module specific test files. Can this be be generalized (like by using `params`)?

tests/mock-lsp/files.lua
```diff
@@ -0,0 +1,76 @@
+_G.lsp_requests = {}
+_G.request_params = {}
+_G.lsp_notifications = {}
+_G.notification_params = {}
```
echasnovski: Four tracking logs is too much. I think only `lsp_requests` and `lsp_notifications` is enough. The logging can then be `table.insert(_G.lsp_requests, { method, params })`.

tests/mock-lsp/only_did_rename.lua
```diff
@@ -0,0 +1,65 @@
+_G.lsp_requests_only_did_rename = {}
```
echasnovski: Creating two separate mocks with a lot of duplicating code only for a single test seems too much. Trying to reuse a single mock is probably better here. And I think it should be possible:

- Create logs once, via `_G.lsp_requests = _G.lsp_requests or {}`.
- Prefer a configurable `capabilities`.
- Prefer a configurable LSP server name: `local server_name = _G.server_name or 'file-methods-lsp'`.
- Check `capabilities` when mocking requests and notifications. Or if it is too complex or not concise or not possible, then checking the server's name might be fine.

tests/mock-lsp/files.lua
```diff
@@ -0,0 +1,76 @@
+_G.lsp_requests = {}
```
echasnovski: That's a nit pick, but could you please rename this mock to something that is not a module's name. It is somewhat confusing to have both 'lua/mini.files.lua' and 'tests/mock-lsp/files.lua'. Maybe 'file-ops.lua' (as a capability is `fileOperations`)?

echasnovski (changes requested):
Overall tests look okay. I've pushed a commit with formatting and small code organization tweaks (trying to keep lines of code as small as reasonably possible).

I am still thinking about the best way to have this configurable. Do you think something like `config.options.lsp_integration = true` would be enough? Or should it be full-blown `config.lsp = { enable = true, timeout = 1000 }` (as in 'oil.nvim')?

---

After addressing this review comments, there is a documentation left. I am thinking:

- A small note at the top in the `Features:` list. This should also be mirrored in the module's README.
- A *slightly* more detailed note in the `MiniFiles-manipulation` section.

echasnovski:
> Sorry for all the back and forth, btw.

No need to apologize. That the nature of the PR review and, if anything, it should be me who needs to apologize :)

> I'm not sure why the tests on ubuntu with Neovim `0.10.4` are failling. It seems to be hitting [neovim/neovim#35550](https://github.com/neovim/neovim/issues/35550) , but that issue hasn't been solved as far as I can tell. So, I'm not sure why it would only happen on `0.10.4`.

Does the feature even work on Neovim 0.10 or is it only test? If not, I am fine with only enabling this on Neovim>=0.11 (i.e. `-- TODO: Remove after compatibility with Neovim=0.10 is dropped`, `if vim.fn.has('nvim-0.11') == 0 then H.lsp_fs_hook = function() end end`.

> I'm also not sure why Windows is not triggering the `willCreate` hook for the file `aaa`. Will take a look tomorrow

Sorry, I have no clue either. Maybe `**/{aaa,bbb}` glob doesn't work there like on Unix?

tests/mock-lsp/file-ops.lua
TheLeoP: A global variable could be used for configuring this. If `params` is used for this, it would mean executing an edit on the created/renamed file. It would still be necessary to use something different for deleted files, though

TheLeoP:
> I am still thinking about the best way to have this configurable. Do you think something like config.options.lsp_integration = true would be enough? Or should it be full-blown config.lsp = { enable = true, timeout = 1000 } (as in 'oil.nvim')?

I prefer the first option (`config.options.lsp_integration = true`), I don't think many people would like to customize the timeout. But, the second option seems more future proof in case other LSP related options need to be added eventually.

TheLeoP:
After that last change the tests at least seem to be running on `0.10.4`. I'll keep investigating tomorrow, I also want to debug the Windows failure on my Windows PC.

tests/mock-lsp/file-ops.lua
echasnovski: Whatever makes it not assume that it runs in a 'mini.files' test in the most concise way.

echasnovski:
I've pushed some tweaks to the latest changes. Including a change in how client is requested on Neovim<0.11. Tests still fail (differently), but my guess is that this needs adjustments in mocks.

I also removed the `TODO` about showing an error to the user. This is not how other LSP aware modules work (like 'mini.completion'), they silently ignore errors. If the behavior is to be changed, then It should be consistent across all modules.

> I prefer the first option (`config.options.lsp_integration = true`), I don't think many people would like to customize the timeout. But, the second option seems more future proof in case other LSP related options need to be added eventually.

Whether it needs to be future proof is the whole point :)

Let's make it more concise then: `config.options.lsp_timeout = 1000` with value `0` meaning no LSP integration. The 'oil.nvim' also has the "autosave" option, but I don't think I like this behavior (requiring an explicit save from the user seems less surprising).

The [relevant section in 'MAINTAINING.md'](https://github.com/nvim-mini/mini.nvim/blob/main/MAINTAINING.md#adding-new-config-settings) will be a helpful checklist here.

TheLeoP:
> Does the feature even work on Neovim 0.10 or is it only test? If not, I am fine with only enabling this on Neovim>=0.11 (i.e. -- TODO: Remove after compatibility with Neovim=0.10 is dropped, if vim.fn.has('nvim-0.11') == 0 then H.lsp_fs_hook = function() end end.

It seems like Neovim `0.10.4` has a bug where `vim.lsp.get_clients({method = method})` can return a client to an LSP server even if `method` is not supported. Using `lua_ls` `vim.lsp.get_clients({method = 'workspace/willCreateFiles'})` returns a list that includes `lua_ls` even though it does not support the `workspace/willCreateFiles` method.

I'll enable the feature only on Neovim>=0.11

tests/test_files.lua
```diff
@@ -5887,6 +5888,281 @@ T['Events']['`MiniFilesActionMove` triggers'] = function()
...
+end
+
+local validate_lsp_will = function(method, files, lines)
+  eq(child.lua_get('_G.lsp_requests["file-methods-lsp"]'), { { method, { files = files } } })
+  eq(get_lines(), lines)
+  child.lua('_G.lsp_requests = {}')
+end
+
+T['LSP']['works with `willCreateFiles`'] = function()
+  if child.fn.has('nvim-0.11') == 0 then MiniTest.skip('LSP integration requires Neovim>=0.11') end
```
echasnovski: This can be done once in test set hook. Like [here](https://github.com/nvim-mini/mini.nvim/blob/d76247d6e3ffa222784be5b156770e37c665f67e/tests/test_cmdline.lua#L391).

echasnovski:
> I'll enable the feature only on Neovim>=0.11

Fine by me. Probably requires an extra "requires Neovim>=0.11" note in a "# LSP integration" help section.

TheLeoP:
I think the PR should be ready for a final review now

lua/mini/files.lua
```diff
@@ -2852,10 +2852,12 @@ H.lsp_fs_hook_client = function(client, full_method, lsp_files)
   local is_fs_type = function(lsp_file, ref_fs_type) return ref_fs_type == nil or ref_fs_type == lsp_file.fs_type end
   local make_filter = function(scheme, ref_fs_type, glob, ignore_case)
     local adjust_case = ignore_case and vim.fn.tolower or function(x) return x end
+    local to_fname = H.is_windows and function(x) return (vim.uri_to_fname(x):gsub('\\', '/')) end
+      or function(x) return vim.uri_to_fname(x) end
```
echasnovski: Are you sure this is a good idea to do in the code itself and not only in tests?

There is nothing in the LSP spec that says that path slashes should always be `/` for glob matching. There is only "(e.g. `**​/*.{ts,js}` matches all TypeScript and JavaScript files)", which I guess can be taken as "assume to use only forward slashes", but still.

And the behavior of `vim.uri_to_fname` to explicitly use `\` on Windows seems to be [intentional](https://github.com/neovim/neovim/blob/32e249dfa6fcf2d079df962473bd257bf04c2a16/test/functional/lua/uri_spec.lua#L102-L109). Still not sure how good this is.

CHANGELOG.md
```diff
@@ -35,6 +35,10 @@ There are following change types:
 
 ## mini.files
 
+### Expand
```
echasnovski, line 48: I've pushed a more detailed 'CHANGELOG.md' entry. Please revert all the other formatting changes and rebase to resolve merge conflicts.

tests/mock-lsp/file-ops.lua
```diff
@@ -0,0 +1,71 @@
...
+
+local capabilities = { workspace = { fileOperations = file_operations_config } }
+
+_G.did_callback = _G.did_callback or function(params, dispatchers) end
+
+local make_will_request = function(method)
+  return function(params)
+    _G.lsp_requests[server_name] = _G.lsp_requests[server_name] or {}
+    table.insert(_G.lsp_requests[server_name], { method, params })
+    local path = vim.fn.fnamemodify('tests/dir-files/lsp-files/main.lua', ':p')
```
echasnovski: @TheLeoP, this is still not resolved.

echasnovski (changes requested):
Overall looks okay. Pushed fixes for some nit picky issues, but there are still couple of unresolved ones.

lua/mini/files.lua
TheLeoP: > Are you sure this is a good idea to do in the code itself and not only in tests?

Yes. As you mentioned, `vim.uri_to_fname` seems to intentionally always use `\` on Windows, which results in globs like `**/something` to never match paths like `C:\something` (while it matches `C:/something` ).

> There is nothing in the LSP spec that says that path slashes should always be / for glob matching. There is only "(e.g. **​/*.{ts,js} matches all TypeScript and JavaScript files)", which I guess can be taken as "assume to use only forward slashes", but still.

There's also no mention of what path separator should `glob`s use in any issue on the LSP spec github repository :/.

Other glob libraries seem to always use `/` as path separator and LSP servers (like `lua_ls`) seem to send globs that use `/` as a path separator, even on Windows. So, I guess that normalizing paths to use `/` in Windows is required to match them against globs.

tests/mock-lsp/file-ops.lua
TheLeoP: Sorry, I forgot resolving this. I'm on it.

tests/mock-lsp/file-ops.lua
```diff
@@ -2,8 +2,7 @@ local server_name = _G.server_name or 'file-methods-lsp'
 _G.lsp_requests = _G.lsp_requests or {}
 _G.lsp_notifications = _G.lsp_notifications or {}
 
-_G.workspace_edits_response = _G.workspace_edits_response
-  or { range = { start = { line = 0, character = 0 }, ['end'] = { line = 0, character = 0 } }, newText = '' }
+_G.workspace_edit_response = _G.workspace_edit_response or nil
```
echasnovski: This does nothing except "declare" configuration possibility. It is not nothing, but I'd still prefer it removed to be more consistent with other LSP mocks.

echasnovski:
> I think the PR should be ready for a final review now

After resolving the latest comment, everything seems to be okay. I'll take some time for a final review with a fresh head in a couple of days.

TheLeoP:
Implementing this took longer than I expected, lol. Thanks for all your feedback.

I'll be waiting for that review in a few days

echasnovski:
> Implementing this took longer than I expected, lol.

Sorry for that :sweat_smile:

TheLeoP:
There's no need to apologize! I just underestimated LSP

echasnovski (approved):
I've given a fresh look and didn't find anything obviously wrong with the code and approach overall. I've squashed and pushed some [cosmetic changes](https://github.com/nvim-mini/mini.nvim/compare/cc3786fb6a48880240febc643ace8036d307e4ff..c710f518bf59b41449393ded87a09f838ae3cbf9).

Let's finally merge this and hope that it won't break too many configs :)

---

@TheLeoP, huge thanks for doing this! Adding this feature was a long time TODO item, but it is mentally and time consuming for me to work on LSP and tree-sitter related stuff. It was a huge help. Hope it wasn't too taxing on your side.

PR #2353, opened 2026-04-07
fix(pick): Handle history editing with keycodes

suy:

### DIFF
--- a/lua/mini/extra.lua
+++ b/lua/mini/extra.lua
@@ -981,7 +981,7 @@ MiniExtra.pickers.history = function(local_opts, opts)
     local cur_match = MiniPick.get_picker_matches().current
     local cur_scope, cur_item = cur_match:match('^(.) (.*)$')
     if not (cur_scope == ':' or cur_scope == '/' or cur_scope == '?') then return end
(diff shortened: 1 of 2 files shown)


suy:
I'm not too confident that I did the test right (but it passes for me locally), so let me know if I should change anything. The issue with the code is that I had in the history stuff like `nmap d<Return> <Cmd>wa<Bar>!busted<Return>`, so it would fail when editing. This fixed it. :)

echasnovski:
Thanks for the PR!

The fix in code seems to be okay, but the test was not quite right: a bit too wide effects (adding new command line history entry affects many other test cases) and not enough coverage. And it was into `stable` and not `main` branch.

So I pushed a fix myself and it should be part of latest `main`. Hope this is okay.

suy:
Definitely! Thanks for everything you do in Neovim. 🙂

I targetted stable because, besides that it's what I am using, I did not see advice on what branch to use on the CONTRIBUTING.md file. But either it's alright for me.

echasnovski:
> ... I did not see advice on what branch to use on the CONTRIBUTING.md file.

To be fair, that's the first time that's ever happened :) Besides, it also lists that there should first be an existing issue, but that is more or less fine in this case.

PR #2356, opened 2026-04-08
feat(git): add `command.cmd` option to configure `:Git` command name

### LINKED ISSUE #1187: Option to rename :Git command to something else like :MiniGit so that vim-fugitive which also uses :Git can be used
### Contributing guidelines

### Module(s)

mini-git

### Description

Hi echasnovski! Is it possible to rename :Git command to something like :MiniGit so that vim-fugitive can be used along with mini-git. I want mini-git show_at_cursor to work as it is while wanting the vim-fugitive's capabilities if possible. Can this be done in any way.

dannydulai:
## Summary
- Add `command.cmd` config option to allow renaming the `:Git` user command
- Useful for avoiding conflicts with other plugins that define `:Git` (e.g. vim-fugitive)
- Defaults to `"Git"` for backwards compatibility
- Updated help docs to document the new option

Example usage:
```lua
require('mini.git').setup({
  command = { cmd = 'MiniGit' },
})
```

Resolve #1187

### DIFF
--- a/doc/mini-git.txt
+++ b/doc/mini-git.txt
@@ -134,8 +134,8 @@ This workflow can be made more interactive when used with mapping, like this: >l
 <
 ------------------------------------------------------------------------------
                                                                           *:Git*
-The `:Git` user command runs `git` CLI call with extra integration for currently
-opened Neovim process:
+The `:Git` user command (renameable via `command.cmd` in |MiniGit.config|) runs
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

As was discussed in #1187, actually implementing this was not the problem and is just out of scope. Configuring user command name crosses the line of what is a good idea to have configurable.

PR #2357, opened 2026-04-09
fix(misc): update `safely` to handle breaking Nightly autocmd changes

abeldekat:

Details:
- PR [38851](https://github.com/neovim/neovim/pull/38851) in neovim/neovim

The test that failed: 'works with "event" without patterns'

### DIFF
--- a/lua/mini/misc.lua
+++ b/lua/mini/misc.lua
@@ -394,6 +394,8 @@ H.execute_later = function()
 end
 
 H.make_defer_autocmd = function(event, pattern, f, trace)
+  pattern = pattern and #pattern > 0 and pattern or nil
+
   local au_id
   local function cb()
     -- Execute exactly once, not once per event or pattern match


lua/mini/misc.lua
```diff
@@ -335,7 +335,7 @@ MiniMisc.safely = function(when, f)
     local ev, patt = events:match('^(.+)~(.+)$')
     local event = vim.split(ev or events, ',', { trimempty = true })
     local pattern = vim.split(patt or '', ',', { trimempty = true })
-    H.make_defer_autocmd(event, pattern, f, trace)
+    H.make_defer_autocmd(event, #pattern > 0 and pattern or nil, f, trace)
```
echasnovski: I think a more universal fix would be to have this kind of adjustment at the top of `H.make_defer_autocmd()` itself.

echasnovski (changes requested):
Thanks for the PR! Look reasonable after adjusting nit picking review comment :)

lua/mini/misc.lua
abeldekat: Yes I thought about that. Decided against because pattern seemed more of a pass-through in that function, unaware of the type of the pattern.

abeldekat: Done

seandewar:
I humbly ask for your forgiveness :innocent:

echasnovski:
> I humbly ask for your forgiveness 😇

We'll think about it. Might require couple of extra months of Neovim Pro+ subscription for @abeldekat and me.

echasnovski:
Should now be part of the latest `main` after a last minute tweak: `pattern and #pattern > 0 and pattern or nil` -> `if type(pattern) == 'table' and #pattern == 0 then pattern = nil end`. Mostly because I fell that Lua's hack for a "ternary operator" works clearly only if it is `aaa and xxx or yyy`. If the condition requires extra subconditions (`aaa and bbb and xxx or yyy`), then it is very confusing to me.

abeldekat:
Perfect. Thanks!

PR #2364, opened 2026-04-11
test(extra): fix test failing conditionally on mini's local path

abeldekat:

Details:
- When mini.nvim is checked out locally to for example '~/projects/fix_mini', test 'oldfiles-respect local_opts.preserve_order' fails because search key 'x' is now present in all items

- Function `validate_spawn_log` is unused and has been removed

___

This PR adds the full test path to the test buffers, in order to display only the relative part in the oldfiles picker.

For example, the display of the first item changes from `~/projects/plugins/mini_test_fix_extra/a...` into `tests/dir-extra/auav`.

The search keys have been changed from 'xy' into 'uv'(letters not in `tests/dir-extra`)

I tested a checkout of mini to directory `abcdefghijklmnopqrstuvwxyz0123456789` to make sure that there are no other similar failing tests....

### DIFF
--- a/tests/test_extra.lua
+++ b/tests/test_extra.lua
@@ -176,12 +176,6 @@ local get_spawn_log = function() return child.lua_get('_G.spawn_log') end
 
 local clear_spawn_log = function() child.lua('_G.spawn_log = {}') end
 
-local validate_spawn_log = function(ref, index)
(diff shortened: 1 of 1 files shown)


echasnovski (approved):
Thanks for the PR! Looks good.

PR #2366, opened 2026-04-13
fix(indentscope): skip drawing indicator on closed fold lines

JanValiska:
## Problem

 When a fold is closed, the indentscope indicator extmark (placed with `virt_text_pos='overlay'`) overlays on top of the foldtext, making the  fold summary unreadable.

## Reproduction

1. Open a file with folds (e.g. a Lua file with treesitter foldexpr)
2. Use a custom `foldtext` that returns virtual text
3. Close a fold (`zc`) that is inside an indentscope
4. The indentscope `│` character overlays on top of the foldtext, covering characters at that column position

## Fix

One-line guard in `H.make_draw_function` that skips lines belonging
to a closed fold before placing the extmark:

```lua
if vim.fn.foldclosed(l) ~= -1 then return true end

Lines hidden inside a fold don't need extmarks, and the first line
of a fold displays foldtext which should not be overwritten.
```

### DIFF
--- a/lua/mini/indentscope.lua
+++ b/lua/mini/indentscope.lua
@@ -987,6 +987,9 @@ H.make_draw_function = function(indicator, opts)
     -- Don't put extmark outside of indicator range
     if not (indicator.top <= l and l <= indicator.bottom) then return true end
 
+    -- Don't draw on lines inside a closed fold (would overlay foldtext)
+    if vim.fn.foldclosed(l) ~= -1 then return true end
+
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR!

This behavior is intentional for at least these reasons:
- It adds a performance overhead for a relatively rare use case (a combination of a closed fold that has non-trivial information at scope's line). 
- It works with `vim.o.foldtext=''`, which disables foldtext and "... the line is displayed normally with highlighting and no line wrapping". Which is what I'd recommend using anyways.
- I don't think interrupted drawing of indent scope looks nice.

PR #2372, opened 2026-04-15
feat(completion): support lsp completion item.command

neo451:

Use `client:exec_cmd` to run `command` field attached to completion items, useful for additional command running, found this is not supported while https://github.com/obsidian-nvim/obsidian.nvim/pull/757 and which use command to create new notes on item accept, I think many LSP servers also use it to add auto imports

Did not find any discussions on it and not sure whether it is in scope, feels pretty natural to add.

### DIFF
--- a/lua/mini/completion.lua
+++ b/lua/mini/completion.lua
@@ -1342,7 +1342,7 @@ H.make_lsp_extra_actions = function(lsp_data)
   -- Prefer resolved item over the one from 'textDocument/completion'
   local item = H.completion.lsp.resolved[lsp_data.item_id] or lsp_data.item
 
-  if item.additionalTextEdits == nil and not lsp_data.needs_snippet_insert then return end
+  if item.additionalTextEdits == nil and not lsp_data.needs_snippet_insert and item.command == nil then return end
   local snippet = lsp_data.needs_snippet_insert and H.get_completion_word(item) or nil
 
   -- Make extra actions not only after an explicit `<C-y>` (accept completed
@@ -1370,7 +1370,11 @@ H.make_lsp_extra_actions = function(lsp_data)
     end
 
     -- Try to only apply additional text edits for non-snippet items
-    if snippet == nil then return H.apply_text_edits(item.client_id, item.additionalTextEdits) end
+    if snippet == nil then
+      H.apply_text_edits(item.client_id, item.additionalTextEdits)
+      H.exec_command(item.client_id, item.command)
+      return
+    end
 
     -- Revert to initial completion state to respect text edit coordinates
     local init_base = H.completion.init_base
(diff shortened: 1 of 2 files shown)


echasnovski:
Thanks for the PR!

I am pretty sure I looked into this at least during the most recent 'mini.completion' set of improvements (LSP related and not). Don't quite remember the reason for this not being added, but vaguely remember there was one. Maybe the lack of an easy and not-that-new (`client:exec_cmd` is Neovim>=0.11) way to execute the command.

Although the code change looks reasonable. I'll take a closer look later.

neo451:
just for additional context

this is what blink.cmp use: https://github.com/saghen/blink.cmp/blob/37c96dc06032b3f128b230b5fb193706535b3811/lua/blink/cmp/sources/lsp/init.lua#L191

from this wrapper I would assume `_exec_cmd` is safe to call for pre-0.11, but need to verify

and I looked at what nvim-cmp does, it has this stalled issue on this  https://github.com/hrsh7th/cmp-nvim-lsp/pull/73

echasnovski:
Thanks again for the PR!

I've merged into a temporary branch to possibly finish this myself. If at all, It'll probably be merged into `main` at least tomorrow.

echasnovski:
@neo451, apart from 'obsidian.nvim', do you happen to know an example of a common LSP server that uses `completionItem.command`? I'd like to test it against the real world use case.

neo451:
@echasnovski I believe rust-analyzer expose auto import as a server command:

https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide-assists/src/handlers/auto_import.rs

But I have not used it myself for a long time, so not 100% sure.

echasnovski:
> @echasnovski I believe rust-analyzer expose auto import as a server command:
> 
> https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide-assists/src/handlers/auto_import.rs
> 
> But I have not used it myself for a long time, so not 100% sure.

Thanks! I've tried `rust-analyzer` with some auto-importing (like for `MAX`) and it uses `additionalTextEdits`. Don't know how to make it use the `command` field.

Is `command` support in 'obsidian.nvim' experimental (i.e. requires the unmerged PR)? If not, could you please list as detailed as possible steps for how I can test `command` support locally?

echasnovski:
I didn't manage to find a real world LSP server that I could easily set up *and* reproduce it returning `completionItem.command`. But let's hope that its support straightforward enough to not introduce any serious issues.

This should now be resolved on the `main` branch. @neo451, could you please test and see if it works as you'd expect?

neo451:
ok, I confirm it indeed works very well on the main now, I was on a small trip yesterday so did not get back to you as soon as possible, and just got back home to found you work so swiftly, thank you for your awesome work!

---

side note, now thinking about it, it does make sense that most common real world LSPs can just use additional text edits for these post actions, their editing footprint should be mostly contained in one file and be declarative, obsidian.nvim is just "abusing" it as a reliable way to run post completion accept hooks, but yeah I guess with more people writing in-process LSP plugins, this support is going to have more real world usecases in the neovim plugin ecosystem.

PR #2382, opened 2026-04-21
test(colors): add more sleeping time to ':Colorscheme' test

abeldekat:

Details:
- Test ':Colorscheme'-'accepts several arguments' fails regularly, especially when repeating 'make test_colors' multiple times. The solution is to also use small_transition_time in the third step to compensate for time spent in the previous steps.

The total sleeping time in the test was exactly 3000 ms. With this PR, the time is 3020 ms.


echasnovski:
Thanks for the PR!

I adjusted the test a bit more to make it more concise and robust. Should be better now on latest `main`.

PR #2389, opened 2026-04-24
fix(files): remember explorer's tabpage when file is opened in new tab

### LINKED ISSUE #2388: [mini.files] Opening mini.files after split to tab breaks future opens
### Contributing guidelines

### Module(s)

mini.files

### Neovim version

0.12.x

### Description

If I open mini.files (from "tab 1") and open the selected file in a new tab (Using ctrl+t binding defined below), then I go back to "tab 1" from the newly opened "tab 2" and open mini.files again, instead of opening mini.files, it jumps back to the newly opened "tab 2" and doesn't open mini.files. However, if I open mini.files from "tab 2", it opens mini.files normally.

I followed the configuration examples that are found in `mini-files.txt`. The configuration looks like this. I'm only including this up front in case the issue is actually something to do with my configuration. There are other mini plugins installed but I assume those will not affect this configuration.

```lua
local MiniFiles = require("mini.files")
MiniFiles.setup()

local minifiles_toggle = function(...)
    if not MiniFiles.close() then
        MiniFiles.open(...)
    end
end

vim.keymap.set("n", "<leader>ew", function()
    minifiles_toggle(nil, false)
end, { desc = "[E]xplore [W]orking Directory" })
vim.keymap.set("n", "<leader>ef", function()
    minifiles_toggle(vim.api.nvim_buf_get_name(0), false)
end, { desc = "[E]xplore [F]iles" })

local mini_files_group = vim.api.nvim_create_augroup("mini_files", { clear = true })

    desc = "Show line numbers File Explorer",
    group = mini_files_group,
    callback = function(args)
        local win_id = args.data.win_id
        vim.wo[win_id].number = true
        vim.wo[win_id].relativenumber = true
    end,
})

-- Define mapping for splitting into files.
local map_split = function(buf_id, lhs, direction)
    local rhs = function()
        -- Make new window and set it as target
        local cur_target = MiniFiles.get_explorer_state().target_window
        local new_target = vim.api.nvim_win_call(cur_target, function()
            vim.cmd(direction .. " split")
            return vim.api.nvim_get_current_win()
        end)

        MiniFiles.set_target_window(new_target)
        MiniFiles.go_in()
    end

    local desc = "Split " .. direction
    vim.keymap.set("n", lhs, rhs, { buffer = buf_id, desc = desc })
end

    desc = "Bindings for splitting on file open",
    pattern = "MiniFilesBufferCreate",
    group = mini_files_group,
    callback = function(args)
        local buf_id = args.data.buf_id
        -- Tweak keys to your liking
        map_split(buf_id, "<C-s>", "belowright horizontal")
        map_split(buf_id, "<C-v>", "belowright vertical")
        map_split(buf_id, "<C-t>", "tab")
    end,
})
```

### Reproduction

    ```lua
    vim.fn.system(clone_cmd)

    vim.cmd('packadd mini.nvim')
    require('mini.deps').setup()

    ```

4.
- Copy the configuration from the above description below the provided content of `nvim-repro/init.lua` and restart neovim.
- Input `<leader>ef` to open mini.files with the provided keybind (since leader key is not set in the provided config, use default leader `\`).
- Select any file and input `<C-t>`. This will open the file in a new tab.
- Input `q`. The new tab should now be visible.
- Input `gt` to go back to the first tab.
- Input `<leader>ef` to open mini.files again.

### The problem
Here I would expect mini.files to open properly, instead it jumps back to the newly opened tab. Interestingly, this only happens if the new tab is opened via mini.files. If you instead open a new tab with any other method (using `<C-t>` in fzf-lua for example) and jump back to the original tab and open mini.files, it opens as expected. This problem only occurs when opening a split from mini.files.

### What I've tried
I thought maybe it had something to do with how mini.files handles the saving of the "previous" state so you can see that I adjusted my keymaps to set `{use_latest}` to false. This unfortunately did not fix the issue.

I am a new dev that's trying to gain additional experience, so I welcome to opportunity to try to fix this on my own, but I would definitely appreciate some pointers on how to get started as well as some patience for any future questions I might have. I also welcome criticism of the results I produce to help me improve. 

</details>

abeldekat:

Resolve #2388


echasnovski:
Thanks for the PR!

This looks like the correct solution indeed. Would you mind please adding two tests:
- One for `close()` returning `nil` and not changing the tabpage in the issue's scenario.
- One for `open()` opening the explorer in the current tabpage in the issue's scenario.

echasnovski:
This should now be part of the latest `main`. I've [adjusted tests](https://github.com/nvim-mini/mini.nvim/compare/07b4d8df1a5b165017272d408423f9e46239b424..be81509b7160aa02264a59789164a2e1c4a6750a): replaced screenshot testing for correct opening with counting floating windows, made better coverage for `use_latest` argument of `open()`, and made both tests a bit more consistent with each other.

Thanks again for the PR!

abeldekat:
Great! Thanks.

PR #2401, opened 2026-05-01
fix(snippets)!: heuristically remove tabstops when text deleted

### LINKED ISSUE #2400: Deleting snippet leaves tabstops and extmarks behind
### Contributing guidelines

### Module(s)

mini.snippets

### Neovim version

0.12.x

### Description

Hello! I'm a big fan of many mini.* plugins. I was trying to integrate mini.snippets into my config when I noticed some really strange behavior: it doesn't seem gracefully handle creating a snippet, and then choosing to delete it.

https://github.com/user-attachments/assets/914a6de5-2940-4cdf-b441-37fd1a52e71d

What I would expect to happen in this situation is for the entire session to be ended, I think the empty extmarks appearing on the next line is not good UX (I'm not familiar with neovim plugin development so it took me while to understand what was happening).

From reading source code and examples, I've come to these assumptions (please correct me if I'm wrong on any):
- Currently the plugin seems to be trying to support exiting from insert mode and coming back, resuming the snippet
- However, it only seems to invalidate the session if the line where a node exists is not present anymore (this only seems to work at the end of the file, as you can see from the example above, I can simply add some text after the snippet and the line still "exists", neovim moves extmarks to the "next" line based on their gravity)

### Reproduction

The config to reproduce this is very simple, I created a simple config with the lazy plugin manager and just mini.snippets + friendly-snippets, no custom or esoteric configuration. If this is not sufficient, please let me know:

https://codeberg.org/ellnix/repro_minisnippets_remaining_extmarks

How to reproduce (with the `foreach` php snippet):
1. Open a php file (or `set ft=php`)
2. Add 3 of lines (either blank or with text)
3. At the top of the file, run the `foreach` snippet
4. Exit to normal mode, delete the snippet
5. Enter insert mode again
6. Observe the extmarks and tabslots on the next line

ellnix:
When the user edits text, check to see if the tabstop nodes have been emptied and pushed to the start of the line, as a heuristic to try to catch cases when the line the node was on was deleted at least up to the end of the node itself.

This heuristic has clear false-positives where it acts, so I did not see fit to fix the tests before getting confirmation that this is a direction that the plugin is willing to go.

Resolve #2400

### DIFF
--- a/lua/mini/snippets.lua
+++ b/lua/mini/snippets.lua
@@ -2084,7 +2084,7 @@ H.var_evaluators = {
 -- Session --------------------------------------------------------------------
 H.get_active_session = function() return H.sessions[#H.sessions] end
 
-H.session_new = function(nodes, snippet, opts)
+H.tabstops_from_nodes = function(nodes)
   -- Compute all present tabstops in session traverse order
   local taborder = H.compute_tabstop_order(nodes)
   local tabstops = {}
@@ -2093,9 +2093,15 @@ H.session_new = function(nodes, snippet, opts)
       { prev = taborder[i - 1] or taborder[#taborder], next = taborder[i + 1] or taborder[1], is_visited = false }
   end
 
+  return taborder[1], tabstops
+end
+
+H.session_new = function(nodes, snippet, opts)
+  local first_tabstop, tabstops = H.tabstops_from_nodes(nodes)
+
   return {
     buf_id = vim.api.nvim_get_current_buf(),
-    cur_tabstop = taborder[1],
+    cur_tabstop = first_tabstop,
     extmark_id = H.extmark_new(0, vim.fn.line('.') - 1, vim.fn.col('.') - 1),
     insert_args = vim.deepcopy({ snippet = snippet, opts = opts }),
     nodes = nodes,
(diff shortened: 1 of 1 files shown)


ellnix:
This branch has changes that I made while experimenting with a way to fix #2400. I realize it diverges from the philosophy of the plugin, I created a PR from it just in case someone finds it interesting, *I don't expect for it to be merged*.

This is my first time writing a non-trivial amount of lua, and my first time writing a contribution to a neovim plugin. If you have the time to provide feedback on how/whether you would solve this, I'd appreciate it :smile:

echasnovski:
Thanks for the PR!

As described [here](https://github.com/nvim-mini/mini.nvim/issues/2400#issuecomment-4359719773) this is a known and documented behavior. Primarily for reasons that there is no robust fix, as any text edit might be intentional from the user. The heuristic approach *might* work, but only if it is concise enough. +78 lines is, unfortunately, not it.

> This is my first time writing a non-trivial amount of lua, and my first time writing a contribution to a neovim plugin. If you have the time to provide feedback on how/whether you would solve this, I'd appreciate it 😄

The Lua side of things looks okay. I didn't read too much into the code logic, but nothing bad catches the eye.

The way I've solved this is by workflow design and not code: always make sure to stop the snippet session if it is not needed. Also by making sure that the bad path does not cause errors and can be quickly fixed (like by pressing `<C-c>`).

Issue #2397, opened 2026-04-29
Search 'markdown_inline' for textobjects when invoked from markdown file


aitvann:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/nvim-mini/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/nvim-mini/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.ai, mini.surround

### Description

I am a long time [treesitter-textobjects](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/) user but recently I decided to give `mini.ai` a try because being able to choose `search_method` is a very powerful feature.

Using regex textobjects for selecting Markdown emphasis is not ideal because it treats everything in between two emphasis as an emphasis in it self (see similar  #1223). That is why I would like to use TreeSitter with custom queries for defining such textobjects.

Code block textobject example

Here is `after/queries/markdown/textobjects.scm`:

``` Scheme
; extends

(fenced_code_block
  (code_fence_content) @cell.inner) @cell.outer
```

And setup:

``` lua
d = spec_treesitter({ a = '@cell.outer', i = '@cell.inner' }),
```

It works as expected. The problem is that `emphasis`, `strong_emphasis` and many other elements are defined inside `markdown_inline` parser instead of `markdown`. 

Here is `after/queries/markdown_inline/textobjects.scm`:

``` Scheme
; extends

(emphasis) @emphasis.outer
((emphasis) @emphasis.inner
    (#offset! @emphasis.inner 0 1 0 -1))
```

And plugin setup:

``` lua
i = spec_treesitter({ a = '@emphasis.outer', i = '@emphasis.inner' }),
```

While code block textobject (`d`) works perfectly fine, emphasis textobject (`i`) does not work at all, it does not even throw errors on invalid query file inside `markdown_inline/`.

Documentation for `MiniAi.gen_spec.treesitter()` [says](https://github.com/nvim-mini/mini.nvim/blob/be81509b7160aa02264a59789164a2e1c4a6750a/doc/mini-ai.txt#L775):
> it uses buffer's |filetype| to determine query language.

What I suppose is happening is that it look for textobjects inside `<vim.bo.filetype>/textobjects.scm` which is `markdown/textobjects.scm` and it does not search `markdown_inline/textobjects.scm` at all.

[...]

PR #2407, opened 2026-05-06
feat(ALL): check children treesitter languages after parents

TheLeoP:

This PR is only a proof of concept of what a fix for #2397 would look like.

Thoughts:

- The easiest way to test this could be to create queries for injecting Lua into some Lua strings
- `mini.surround` doesn't seem to need this change since the cursor always needs to already be on top of the region with a treesitter injection. So, it will be parsed even before this PR when going up the tree.

### DIFF
--- a/lua/mini/ai.lua
+++ b/lua/mini/ai.lua
@@ -993,6 +993,10 @@ end
 ---   Verify with `:=vim.treesitter.query.get('lang', 'textobjects')` and see
 ---   if the target capture is recognized as one.
 --- - It uses buffer's |filetype| to determine query language.
+--- - It first searches the language under cursor for matches. If no matches are
+---   found, it fallbacks to searching parent languages (up to the buffer's root
+---   language). If no matches are found again, it fallbacks to recursively
+---   searching all children languages (from the language under cursor).
 --- - On large files it is slower than pattern-based textobjects. Still very
 ---   fast though (one search should be magnitude of milliseconds or tens of
 ---   milliseconds on really large file).
@@ -1608,21 +1612,28 @@ H.get_matched_ranges_builtin = function(captures)
   -- Get parser (LanguageTree) at cursor (important for injected languages)
   local pos = vim.api.nvim_win_get_cursor(0)
   local lang_tree = parser:language_for_range({ pos[1] - 1, pos[2], pos[1] - 1, pos[2] })
+  local init_lang_tree = lang_tree
 
   local missing_query_langs = {}
   local res = {}
   -- Maybe go up parent trees to work with injected languages
   while vim.tbl_isempty(res) and lang_tree ~= nil do
-    local lang = lang_tree:lang()
-    -- Get query file depending on the local language
-    local query = vim.treesitter.query.get(lang, 'textobjects')
+    H.append_lang_ranges(res, missing_query_langs, buf_id, captures, lang_tree)
 
-    if query ~= nil then H.append_ranges(res, buf_id, query, captures, lang_tree) end
-    if query == nil then missing_query_langs[lang] = true end
-
-    -- `LanguageTree:parent()` was added in Neovim<0.10
+    -- `LanguageTree:parent()` was added in Neovim=0.10
     -- TODO: Drop extra check after compatibility with Neovim=0.9 is dropped
     lang_tree = lang_tree.parent and lang_tree:parent() or nil
+  end
+  -- Fallback to children trees for injected languages
+  if vim.tbl_isempty(res) then
+    local check_children
+    check_children = function(l_tree)
+      for _, child in pairs(l_tree:children()) do
+        H.append_lang_ranges(res, missing_query_langs, buf_id, captures, child)
+        check_children(child)
+      end
+    end
+    check_children(init_lang_tree)
   end
 
   if vim.tbl_isempty(res) and not vim.tbl_isempty(missing_query_langs) then
@@ -1630,6 +1641,14 @@ H.get_matched_ranges_builtin = function(captures)
   end
 
   return res
+end
+
+H.append_lang_ranges = function(res, missing_query_langs, buf_id, captures, lang_tree)
+  local lang = lang_tree:lang()
+  local query = vim.treesitter.query.get(lang, 'textobjects')
+
+  if query ~= nil then H.append_ranges(res, buf_id, query, captures, lang_tree) end
+  if query == nil then missing_query_langs[lang] = true end
 end
 
 H.append_ranges = function(res, buf_id, query, captures, lang_tree)
--- a/lua/mini/surround.lua
+++ b/lua/mini/surround.lua
@@ -1006,6 +1006,10 @@ MiniSurround.gen_spec = { input = {}, output = {} }
 ---   Verify with `:=vim.treesitter.query.get('lang', 'textobjects')` and see
 ---   if the target capture is recognized as one.
 --- - It uses buffer's |filetype| to determine query language.
+--- - It first searches the language under cursor for matches. If no matches are
+---   found, it fallbacks to searching parent languages (up to the buffer's root
+---   language). If no matches are found again, it fallbacks to recursively
+---   searching all children languages (from the language under cursor).
 --- - On large files it is slower than pattern-based textobjects. Still very
 ---   fast though (one search should be magnitude of milliseconds or tens of
 ---   milliseconds on really large file).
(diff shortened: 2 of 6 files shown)


echasnovski:
Thanks for the PR!

> * The easiest way to test this could be to create queries for injecting Lua into some Lua strings

There are already tests for injected languages ([this one](https://github.com/nvim-mini/mini.nvim/blob/48555f1615f08c6318150aacb52bf7d61a5af798/tests/test_ai.lua#L847-L858) for 'mini.ai'). They use built-in Lua -> Vimscript injections.

I think tests for this could use even Lua -> Vimscript -> Lua injections. Like here:

```lua
vim.cmd([[
  set cursorline
  lua local a = 1
]])
```

> * `mini.surround` doesn't seem to need this change since the cursor always needs to already be on top of the region with a treesitter injection. So, it will be parsed even before this PR when going up the tree.

There is less overall complexity if tree-sitter related (and even some other) code in 'mini.ai' and 'mini.surround' is kept as aligned as possible.

---

As per this PR, please reduce code duplication. Both parents and children are meant to be processed exactly the same, right? In both 'mini.ai' and 'mini.surround' this processing is rather complicated, so even with two times duplication it adds a lot.

TheLeoP:
> There are already tests for injected languages

I forgot about that test and I wrote it myself :p. Thanks for the pointer!

Using Lua -> Vimscript -> Lua injections was indeed easier to avoid adding new queries.

> code duplication and `mini.ai`/`mini.surround` alignment on treesitter implementation

I think I improved the alignment/code duplication between `mini.ai` and `mini.surround`, but I don't usually write small functions. So, let me know if you have any more feedback.

lua/mini/ai.lua
```diff
@@ -1612,17 +1612,19 @@ H.get_matched_ranges_builtin = function(captures)
   local missing_query_langs = {}
   local res = {}
   -- Maybe go up parent trees to work with injected languages
-  while vim.tbl_isempty(res) and lang_tree ~= nil do
-    local lang = lang_tree:lang()
-    -- Get query file depending on the local language
-    local query = vim.treesitter.query.get(lang, 'textobjects')
+  local cur_lang_tree = lang_tree
```
echasnovski:
```suggestion
  local init_lang_tree = lang_tree
```

I'd cache initial `lang_tree` to later use when searching through child trees. It makes it more explicit that the step is done for the initial language tree and leads to a slightly smaller diff as a bonus.

lua/mini/surround.lua
```diff
@@ -1545,28 +1545,24 @@ H.get_matched_range_pairs_builtin = function(captures)
   -- Get parser (LanguageTree) at cursor (important for injected languages)
   local pos = vim.api.nvim_win_get_cursor(0)
   local lang_tree = parser:language_for_range({ pos[1] - 1, pos[2], pos[1] - 1, pos[2] })
+  local cur_lang_tree = lang_tree
 
   local missing_query_langs = {}
   -- Compute matched ranges for both outer and inner captures
   -- Maybe go up parent trees to work with injected languages
   local outer_ranges, inner_ranges = {}, {}
```
echasnovski:
```suggestion
  local outer, inner = {}, {}
```

Although this makes it less explicit, together with adjusting names in all the helper functions it feels like easier to understand.

lua/mini/surround.lua
```diff
@@ -1583,6 +1579,22 @@ H.get_matched_range_pairs_builtin = function(captures)
   return res
 end
 
+H.append_lang_ranges = function(outer_ranges, inner_ranges, missing_query_langs, buf_id, captures, lang_tree)
```
echasnovski: Sigh... Six argument helpers is not exactly a nice thing to have, but I don't see any other a better solution.

echasnovski (changes requested):
This indeed looks like a well structured change and improvement!

Apart from variable naming nit picks, could you please also:
- Add a concise note in each function's help about how it searches for the match.
- Add a 'mini.surround' test.
- Split into two commits for for 'mini.ai' and 'mini.surround'. Each with a concise description in `Details:` list about why this is necessary. With also `Resolve #2397` line afterwards.

lua/mini/surround.lua
TheLeoP: We could call `vim.api.nvim_get_current_buf()` inside of the function to avoid passing it as a parameter (? Or maybe create one `outer_ranges`/`inner_ranges`/`missing_query_langs` on each function call and concatenate them all on `H.get_matched_ranges_builtin`

lua/mini/ai.lua
TheLeoP: That seems clearer

TheLeoP:
> Add a concise note **in each function's** help about how it searches for the match.

Are you referring to `H.get_matched_ranges_builtin`? Or to `H.get_nodes_range_builtin` and `H.get_match_ranges_builtin`?

 > Add a concise note in each function's **help** about how it searches for the match.

This is about adding a comment to the private helper functions, right? Or is it about modifying the comments used to generate the documentation for the user-facing `MiniAi.gen_spec.treesitter()` and `MiniSurround.gen_spec.input.treesitter()`?

> Add a 'mini.surround' test.

I don't think a test for this change would be possible inside of `mini.surround`. The cursor needs to always be inside of a surround, which means that the surround would be either on the same `LanguageTree` or in a parent one. I can't think of a scenario where the surrounding would be on a child language tree.

lua/mini/surround.lua
```diff
@@ -1545,35 +1545,31 @@ H.get_matched_range_pairs_builtin = function(captures)
...
+      H.append_lang_ranges(outer, inner, missing_query_langs, buf_id, captures, child)
+    end
+  end
 
   -- Match outer and inner ranges: for each outer range pick the biggest inner
   -- range that lies within outer
   local res = {}
-  for i, outer in ipairs(outer_ranges) do
-    res[i] = { outer = outer, inner = H.get_biggest_nested_range(inner_ranges, outer) }
+  for i, outer in ipairs(outer) do
```
echasnovski: `outer` is redefined here.

lua/mini/surround.lua
echasnovski: That was written like this with performance in mind. But maybe it will not have a big impact to create new tables for each tree as there are usually not so many trees to traverse, right? Maybe it is even a single tree in most cases.

I'll maybe take a look later myself in a separate refactor.

tests/test_ai.lua
```diff
@@ -857,6 +857,16 @@ T['gen_spec']['treesitter()']['works with parent of injected language'] = functi
   validate_find(lines, { 3, 0 }, { 'a', 'F' }, { { 1, 13 }, { 5, 3 } })
 end
 
+T['gen_spec']['treesitter()']['works with direct injected child language'] = function()
+  local lines = {
+    'vim.cmd([[',
+    'set cursorline',
+    'lua local a = function () end',
```
echasnovski:
```suggestion
    'lua local a = function() return true end',
```

This should make it more robust: no space between `function` and `()` is usual (unless it is intentional?) and non-empty body (as empty function might be problematic).

echasnovski:
> This is about adding a comment to the private helper functions, right? Or is it about modifying the comments used to generate the documentation for the user-facing `MiniAi.gen_spec.treesitter()` and `MiniSurround.gen_spec.input.treesitter()`?

Sorry, should have been more clear. Mentioning function's *help* in my mind means the text that is shown when doing `:h <function-name>`. For commenting local functions it is usually "function's comment" or "function's annotation".

And the explanation is very much needed because I just now (possibly) misremembered how this is supposed to work in the 'mini.ai' test. Is it intended and/or reasonable that the search is only done for "one level" children? In particular with this Lua code with cursor at the top, executing `vanF` doesn't work because the target is two levels deep (Lua->Vimscript->Lua):

```lua
vim.cmd([[
set cursorline
lua local a = function() return true end
]])
```

I think I would have expected all children to be used here, no matter however deep. Do you think it is reasonable?

> > Add a 'mini.surround' test.
> 
> I don't think a test for this change would be possible inside of `mini.surround`. The cursor needs to always be inside of a surround, which means that the surround would be either on the same `LanguageTree` or in a parent one. I can't think of a scenario where the surrounding would be on a child language tree.

The 'mini.ai' and 'mini.surround' are very aligned in how they find their targets (textobjects and surroundings). In this particular case, using `next` variant of surrounding action (like `sfn` to "find next surrounding") will work. Like this:

```lua
T['gen_spec']['input']['treesitter()']['works with direct injected child language'] = function()
  local lines = {
    'vim.cmd([[',
    'set cursorline',
    'lua local a = function () end',
    ']])',
  }
  validate_find(lines, { 2, 0 }, { { 3, 14 } }, type_keys, 'sfn', 'F')
end
```

TheLeoP:
>  Is it intended and/or reasonable that the search is only done for "one level" children? 

It was intended, I misunderstood your expectation and though you expected only first level children to be checked. Sorry.

> I think I would have expected all children to be used here, no matter however deep. Do you think it is reasonable?

I would expect the same thing. The only remaining question would be, do you think it would be better to search all the children `LanguageTree` depth first or breadth first? The results would be different on some specific edge-cases, since we are stopping the search as soon as any value is found. I *think* that depth first could be a more intuitive behavior for end users.

tests/test_ai.lua
TheLeoP: > no space between function and () is unusual

Sorry, it was a typo

lua/mini/ai.lua
```diff
@@ -1629,22 +1608,31 @@ H.get_matched_ranges_builtin = function(captures)
...
+    H.append_lang_ranges(res, missing_query_langs, buf_id, captures, lang_tree)
 
-    -- `LanguageTree:parent()` was added in Neovim<0.10
+    -- `LanguageTree:parent()` was added in Neovim=0.10
     -- TODO: Drop extra check after compatibility with Neovim=0.9 is dropped
     lang_tree = lang_tree.parent and lang_tree:parent() or nil
   end
+  -- Fallback to children trees for injected languages
+  local check_children
+  check_children = function(l_tree)
```
TheLeoP: The code to recursively check all children is not as clean as I would like it to be. What do you think, @echasnovski ? Maybe an iterative approach could look cleaner?

lua/mini/ai.lua
```diff
@@ -1629,22 +1608,31 @@ H.get_matched_ranges_builtin = function(captures)
...
+    if not vim.tbl_isempty(res) then return end
+
+    for _, child in pairs(l_tree:children()) do
+      H.append_lang_ranges(res, missing_query_langs, buf_id, captures, child)
+      if not vim.tbl_isempty(res) then return end
+
+      check_children(child)
+    end
+  end
+  check_children(init_lang_tree)
```
echasnovski: 1. I think this code will be rare, so guarding it whole behind the `if vim.tbl_isempty()` is reasonable (although adds one extra indent).
2. The problem with early return inside a cycle is that there is no guarantee about the order of the traversal. So this will have to be "fall back to *all* child trees".

So something like this:

```suggestion
  if vim.tbl_isempty(res) then
    local check_children
    check_children = function(l_tree)
      for _, child in pairs(l_tree:children()) do
        H.append_lang_ranges(res, missing_query_langs, buf_id, captures, child)
        check_children(child)
      end
    end
    check_children(init_lang_tree)
  end
```

lua/mini/ai.lua
echasnovski: Answered [here](https://github.com/nvim-mini/mini.nvim/pull/2407/changes#r3316037307).

lua/mini/surround.lua
```diff
@@ -1545,35 +1545,38 @@ H.get_matched_range_pairs_builtin = function(captures)
...
+    if not vim.tbl_isempty(inner) and not vim.tbl_isempty(outer) then return end
+
+    for _, child in pairs(l_tree:children()) do
+      H.append_lang_ranges(outer, inner, missing_query_langs, buf_id, captures, child)
+      if not vim.tbl_isempty(inner) and not vim.tbl_isempty(outer) then return end
+
+      check_children(child)
+    end
+  end
+  check_children(init_lang_tree)
```
echasnovski: Similar to ['mini.ai' case](https://github.com/nvim-mini/mini.nvim/pull/2407/changes#r3316037307):

```suggestion
  if vim.tbl_isempty(inner) or vim.tbl_isempty(outer) then
    local check_children
    check_children = function(l_tree)
      for _, child in pairs(l_tree:children()) do
        H.append_lang_ranges(outer, inner, missing_query_langs, buf_id, captures, child)
        check_children(child)
      end
    end
    check_children(init_lang_tree)
  end
```

echasnovski (changes requested):
The recursive approach is probably cleaner than iterative one.

Also, please, add the concise note about how traversal is done in respective help annotations.

echasnovski:
> It was intended, I misunderstood your expectation and though you expected only first level children to be checked. Sorry.

No worries. I should have been clearer.

> I would expect the same thing. The only remaining question would be, do you think it would be better to search all the children `LanguageTree` depth first or breadth first? The results would be different on some specific edge-cases, since we are stopping the search as soon as any value is found. I _think_ that depth first could be a more intuitive behavior for end users.

The problem with early return inside a cycle is that there is no guarantee about the order of the traversal. So this will have to be "fall back to all child trees".

echasnovski (approved):
This looks very good as is, but I'll tweak some wording in help. Thank you very much for this work!

Issue #234, opened 2023-02-19
mini.animate: Don't animate scrolling after search


benyn:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.animate

### Description

Thanks for this great plugin! I love `mini.animate` along with some of your other plugins, but thought it'd be great to be able to disable scroll animation after search (`/`).

If the search result is outside my current window, Nvim already moves my window to show the result even before I type `<CR>`. When I type `<CR>`, without `mini.animate` turned on, both the cursor and window just stay there. But with `mini.animate`'s scrolling animation turned on, the window screen moves back up to my original cursor position and comes back down to the search result. That's quite dizzying.

Similar to [LazyVim's config](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/ui/mini-animate.lua), which disables animation on mouse scrolling, I set up the following keymap and it seems to work. But I'm not sure if this is the best way to do this.
```lua
vim.keymap.set("c", "<CR>", function()
  disable_animation = true
  return "<CR>"
end, { expr = true })
```

I'd totally understand if this isn't something you'd like to add officially to the plugin to keep it minimal. But I thought this is highly aligned with the goal of this plugin which is to help our brains cope with context switching better.

echasnovski:
Thanks for your kind feedback!

Yes, this sometimes bothers me too. Not to the point of becoming dizzy, but still.

Main problem here is that `WinScrolled` event is not triggered during incremental search (neovim/neovim/issues/21207) but only after it. Which means that 'mini.animate' is notified that scroll should be done after search has started, at which point it knows only about window view prior whole search and window view after user hits `<CR>`.

At the moment I am extremely hesitant adding custom mappings in order to "patch" any animation behavior.

I'll try to tackle this problem again. If not successful, I'll add a workaround to documentation; similar to as you suggested.

benyn:
Thanks for the quick response! What you suggested makes a lot of sense given the limitation imposed by Neovim. Hope that Neovim issue gets resolved so we can solve this gracefully rather than through a bandaid.

echasnovski:
This should be implemented in latest `main`.

[...]

PR #2408, opened 2026-05-06
fix(animate): react to Nightly triggering `WinScrolled` on `incsearch`

abeldekat:

Details:
- See PR [39308](https://github.com/neovim/neovim/pull/39308) in neovim/neovim

Related to #234


echasnovski:
Thanks for the PR!

If there is any logic change in the code, the commit is a `fix`, not a `test(...)`.

Skipping an existing test on "future facing" Neovim version check is very unusual. It means that certain behavior was intended earlier but not it is not. The `does not automatically animate result of 'incsearch'` describes a behavior that is intended, so there should be a deliberate fix and not just `MiniTest.skip()`.

abeldekat:
As this is a blocking issue, I don't think I can solve it in a timely fashion. Closing in favor of a better solution...

PR #2411, opened 2026-05-08
feat(base16): add support for blink.cmp

kristijanhusak:
Add support for [blink.cmp](https://github.com/saghen/blink.cmp) to base16. 
Used same colors as for nvim-cmp, just adjusted the hl group names.


echasnovski:
Thanks for the PR!

This is a bit too big of an addition. Looking at 'nvim-cmp', I'd say that adding explicit override for every `Kind` is a bit too much also.

As I don't use 'blink.cmp', could you please post a screenshot of it using these highlights before and after changes from this PR?

kristijanhusak:
> This is a bit too big of an addition. Looking at 'nvim-cmp', I'd say that adding explicit override for every Kind is a bit too much also.

I'm not sure what you mean. nvim-cmp is already in the code. I just followed the same structure while adjusting the hl group names.

> As I don't use 'blink.cmp', could you please post a screenshot of it using these highlights before and after changes from this PR?

Here are a couple of screenshots. It doesn't show all the colors, but gives the idea how it looks.

screenshot_2026_05_13_11_48_14
screenshot_2026_05_13_11_46_54

echasnovski:
> I'm not sure what you mean. nvim-cmp is already in the code. I just followed the same structure while adjusting the hl group names.

I meant that if I were to add the 'nvim-cmp' integration now, it would probably not have all the `XxxKind` groups.

> > As I don't use 'blink.cmp', could you please post a screenshot of it using these highlights before and after changes from this PR?
> 
> Here are a couple of screenshots. It doesn't show all the colors, but gives the idea how it looks.

They don't look much different. Those are both "after", right?

kristijanhusak:
> I meant that if I were to add the 'nvim-cmp' integration now, it would probably not have all the XxxKind groups.

I can move things around if you have some suggestions. I think this is fine since users can enable/disable specific integrations.


> They don't look much different. Those are both "after", right?

Ah yeah, right. Here's the comparison:

<table>
<tr>
 <td>before</td>
 <td>after</td>
<tr>
 <td>screenshot_2026_05_13_12_09_04</td>
 <td>screenshot_2026_05_13_11_48_14</td>
<tr>
 <td>screenshot_2026_05_13_12_08_42</td>
 <td>screenshot_2026_05_13_11_46_54</td>
</table>

echasnovski:
> I can move things around if you have some suggestions. I think this is fine since users can enable/disable specific integrations.

No need, thanks. It is mostly me thinking if adding all these `XxxKind` groups is worth it for 'blink.cmp'.

> Ah yeah, right. Here's the comparison:

Thanks, that's useful! The obvious usability issue is not showing matched characters. The not colored kinds are kind of okay.

I'll need a bit more time to think about this.

kristijanhusak:
> if adding all these XxxKind groups is worth it for 'blink.cmp'

I'd say that blink.cmp is more popular nowadays than nvim-cmp, so I think it's worth having both of these.

-----

Note that I'm using a custom layout for the completion popup. This is how it looks with default layout:

<table>
<tr>
 <td>before</td>
 <td>after</td>
<tr>
 <td>
screenshot_2026_05_13_12_51_28</td>


 <td>screenshot_2026_05_13_12_49_18</td>
</table>

echasnovski:
This should now [be fixed on latest `main`](https://github.com/nvim-mini/mini.nvim/blob/6a8aab4b07c6caa1d0643c1c9dccd5b2d5f5e094/lua/mini/base16.lua#L1342-L1369) (along with parallel integration in 'mini.hues'). I manually tested the final groups and not all of them needed to be set. And setting all `XxxKind` groups indeed is what a good color scheme should be doing (although 'blink.cmp' itself can be a bit more flexible here by reusing highlights from 'mini.icons').

Thanks again for the PR! I've added you as a co-author of the 'mini.base16' integration commit.

PR #2422, opened 2026-05-17
test(pick): react to Nightly improving `bufdelete` from floating windows

abeldekat:

Details:
- See PR [39800](https://github.com/neovim/neovim/pull/39800) in neovim/neovim

I think this is a complicated one...

Somehow, in `tests/test_pick.lua | default_choose() | works for directory path`, with mini.files, the deletion of `buf_id_init` fails: Failed to unload buffer.

I replayed the test steps manually and did not see anything unusual. Deleting the initial buffer when mini.files is open seems to work just fine.

### DIFF
--- a/tests/test_pick.lua
+++ b/tests/test_pick.lua
@@ -2125,10 +2125,9 @@ T['default_choose()']['works for directory path'] = function()
     validate_buf_name(buf_id_init, path)
 
     -- Cleanup
-    child.api.nvim_buf_delete(buf_id_init, { force = true })
-    child.api.nvim_buf_delete(buf_id_cur, { force = true })
-
     child.lua('if _G.MiniFiles ~= nil then _G.MiniFiles.close() end')
+    pcall(child.api.nvim_buf_delete, buf_id_init, { force = true })
+    pcall(child.api.nvim_buf_delete, buf_id_cur, { force = true })
   end
 
   validate(test_dir, test_dir, 'netrw')


echasnovski:
> I think this is a complicated one...
> 
> Somehow, in `tests/test_pick.lua | default_choose() | works for directory path`, with mini.files, the deletion of `buf_id_init` fails: Failed to unload buffer.
> 
> I replayed the test steps manually and did not see anything unusual. Deleting the initial buffer when mini.files is open seems to work just fine.

That's strange indeed. My guess is that this *might* be an upstream issue that is worth trying to find a minimal plugin-free reproduction for.

echasnovski:
Well, there is neovim/neovim#39827, but I am not sure it is even related (as I can not reproduce the error from the test fail).

**Edit**: Yes, it is an upstream issue, technically speaking. In this case there is an extra [`BufEnter` event triggered](https://github.com/neovim/neovim/issues/39827#issuecomment-4470869254), which arguably should not happen. This triggers [directory tracking callback](https://github.com/nvim-mini/mini.nvim/blob/18797ed18f347d0d0da27fec28a67979b1207f70/lua/mini/files.lua#L1383) which [deletes a buffer before the "original" `nvim_buf_delete()` can](https://github.com/nvim-mini/mini.nvim/blob/18797ed18f347d0d0da27fec28a67979b1207f70/lua/mini/files.lua#L1434). Hence the error.

Let's wait a bit for this to be resolved upstream. Otherwise, moving those cleanup commands after `MiniFiles.close()` looks fine, but I'd use `pcall(child.api.nvim_buf_delete, ...)` instead of an extra `if` check.

abeldekat:
```lua
require('mini.files').setup()
require('mini.pick').setup()
```

1. nvim
2. `:lua MiniPick.default_choose('/home/user/somedir')`
3. `:lua vim.api.nvim_buf_delete(1, { force = true })`
4. Observe: 

```lua
E5108: Lua: [string ":lua"]:1: Failed to unload buffer.                                                                                                                                        
stack traceback:                                                                                                                                                                               
        [C]: in function 'nvim_buf_delete'                                                                                                                                                     
```

Note:
Succeeds when step one takes an argument: `nvim somefile.txt`
Also succeeds when replacing step 2 with: `lua MiniFiles.open('/home/user/somedir')`

echasnovski (approved):
Well, the upstream problem is reported and tracked upstream. After another look, I think this order of operations makes sense in its own rights. So lets merge this.

Thanks again for the PR!

echasnovski:
For posterity: this failing test would have been now fixed upstream (in the next Nightly). But it already did good by detecting an upstream issue early on.

PR #2428, opened 2026-05-23
test(keymap): react to Nightly changes regarding autoindent

abeldekat:

Details:
- See PR [39961](https://github.com/neovim/neovim/pull/39961) in neovim/neovim

I wasn't able to fix this properly. The test now passes because I disabled `autoindent`, allowing the third line to be empty.

I wonder why the `nvim-autopairs` mock is more specific than the other mocks, which mostly add a line to the `_G.log`. Is the comment stating `-- Mock exactly how 'nvim-autopairs' works here` still applicable? Perhaps a simple `_G.log` addition is enough.

The failing test:

```lua
FAIL in tests/test_keymap.lua | map_multistep() | built-in steps | nvimautopairs_cr:
  Failed expectation for equality
  Cause: different values at key 3, left = ")", right = ""
  Left:  { "", "(", ")", "" }
  Right: { "", "(", "", ")" }
  Traceback:
    tests/test_keymap.lua:1088
```

EDIT:
I think [this](https://github.com/neovim/neovim/issues/37780) issue can also be closed...

EDIT2:
It's strange that the third empty line is now subject to truncation. Maybe the vim-patch is not selective enough?

### DIFF
--- a/tests/test_keymap.lua
+++ b/tests/test_keymap.lua
@@ -1080,6 +1080,7 @@ T['map_multistep()']['built-in steps']['nvimautopairs_cr'] = function()
   eq(get_lines(), { '', '' })
 
   mock_plugin('nvim-autopairs')
+  child.bo.autoindent = false
 
   -- Should respect pairs
   type_keys('()', '<Left>')


echasnovski:
Thanks for the PR!

> I wonder why the `nvim-autopairs` mock is more specific than the other mocks, which mostly add a line to the `_G.log`. Is the comment stating `-- Mock exactly how 'nvim-autopairs' works here` still applicable?

Mostly because [it requires extra key translation](https://github.com/nvim-mini/mini.nvim/blob/a405061478d8027b4e21214d897ce1eb33115792/lua/mini/keymap.lua#L823) to be useful. And that sequence of keys is [exactly what 'nvim-autopairs' uses](https://github.com/windwp/nvim-autopairs/blob/7b9923abad60b903ece7c52940e1321d39eccc79/lua/nvim-autopairs.lua#L613-L616).

As the matter of fact, the 'nvim-autopairs' plugin itself now acts on Nightly exactly as is detected by 'mini.keymap' test. It might be the problem in either vim-patch or 'nvim-autopairs'. Don't really have the bandwidth to investigate this (although it would probably be a net positive for Neovim ecosystem :) ).

tests/test_keymap.lua
```diff
@@ -1081,6 +1081,9 @@ T['map_multistep()']['built-in steps']['nvimautopairs_cr'] = function()
 
   mock_plugin('nvim-autopairs')
 
+  -- Avoid trunctation of the empty line
+  child.bo.autoindent = false
+
```
echasnovski:
```suggestion
  child.bo.autoindent = false
```

This comment is useful now in context of this PR, but it is 1) only relevant for Neovim>=0.13 (which is not a stable release); 2) might become outdated if the behavior is fixed on Nightly. Besides, the code is mostly self-explanatory.

abeldekat: done

Issue #2164, opened 2025-12-10
mini.files: Combine directories that contain exactly one directory into one entry


carbon-steel:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/nvim-mini/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/nvim-mini/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.files

### Description

In cases like working with a Java codebase, there are directory that contain another directory and nothing else. For example,

* `com/` only contains `google/`
* which only contains `search/`
* which only contains `team_name/`
* which only contains `application_name/` which contains all the source code for an app

Using mini.files, I often have to enter each directory individually. I'd like the option to configure the plugin to combine such cases of nesting into a single entry in the mini.files UI. In the case above, I'd see `com/google/search/team_name/application_name`. Pressing enter on that entry would take me directly to the `application_name` directory.

A slight drawback to this feature: this feature would mean that we'd have to have a way to also enter the `com/` without skipping all the way to `application_name` since there is the (minority) case where we don't want to skip.

There is a bit of thought needed on the exact design/implementation of this feature, but I think this idea would make mini.files more ergonomic for the common use case.

echasnovski:
Thanks for the suggestion!

This was already suggested in #1217... I think. As the example is also from Java :)

The [conclusion](https://github.com/nvim-mini/mini.nvim/issues/1217#issuecomment-2351085697) still stands. The solution from #1184 might be a good enough alternative here.

Issue #1217, opened 2024-09-14
support merge emty directory


mycf:
### Contributing guidelines

- [X] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [X] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.files

### Description

less press `l`
<img width="301" alt="image" src="https://github.com/user-attachments/assets/bd7de221-5f42-4603-9589-e5a3727fdd8d">

echasnovski:
Thanks for the suggestion!

This is not planned as I think this behavior is "too magical". See #1184 for a possible solution that makes "less press `l`" possible.

PR #2432, opened 2026-05-26
feat(files): add visual directory collapse

2KAbhishek:

### Problem
Java/Kotlin/Android developers navigating deeply nested, single-child directory structures face a "navigation tax," having to repeatedly press `l` or `h` through empty intermediate directories.

### Proposed Solution
This PR adds visual directory collapsing to `mini.files` when `content.collapse_single_child = true` is set (disabled by default).

- **Forward Collapse**: Merges single-child directories visually (e.g. `a/b/c/`). Path index is registered to the leaf, allowing preview and selection (`l`) to jump straight to the nested files.
- **Symmetric Navigation**: Pressing `h` climbs up all single-child folders in one step. Updated cursor decoding so `h` lands exactly on the collapsed line instead of reverting to line 1.
- **Orphan Directory Cleanup**: Deletes/renames of collapsed leaves automatically clean up empty parent directories up to the buffer root path on synchronization, checking that parent buffers are not actively preserving them.
- **Flicker Fix**: Replaced `children_path_ids` caching with `cached_children` to preserve custom collapsed names across window updates.

### Verification
- All existing tests pass.
- Added 3 new unit test cases covering collapse rendering, reverse navigation, and directory cleanup.

### Screenshots

When config is false:
image

When config is true:
image

### DIFF
--- a/doc/mini-files.txt
+++ b/doc/mini-files.txt
@@ -582,6 +582,8 @@ Defaults ~
   MiniFiles.config = {
     -- Customization of shown content
     content = {
+      -- Whether to collapse single-child directory chains
+      collapse_single_child = false,
       -- Predicate for which file system entries to show
       filter = nil,
       -- Highlight group to use for a file system entry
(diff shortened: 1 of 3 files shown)


2KAbhishek:
@echasnovski can you please take a look at this PR 🙏🏽 
It will be very helpful to have this in

echasnovski:
Thanks for the PR!

This was already suggested and discussed in #2164. The [conclusion](https://github.com/nvim-mini/mini.nvim/issues/1217#issuecomment-2351085697) still stands, so this is not planned.

PR #2438, opened 2026-05-28
fix(files): remove invalid window from `explorer.windows`

abeldekat:

Details:
- When MiniFilesWindowUpdate uses `get_explorer_state()` with enabled preview, an error occurs after the user presses `o`, `<Esc>`, `u`.

The error:

```txt 
Error in User Autocommands for "MiniFilesWindowUpdate":                                                                                                                                      
Lua callback: ...m-repro/site/pack/vnr/start/mini.nvim/lua/mini/files.lua:1105: Invalid window id: 1004                                                                                      
stack traceback:                                                                                                                                                                             
        [C]: in function 'nvim_win_get_buf'                                                                                                                                                  
        ...m-repro/site/pack/vnr/start/mini.nvim/lua/mini/files.lua:1105: in function 'get_explorer_state'                                                                                   
        /home/user/.config/nvim-repro/init.lua:12: in function </home/user/.config/nvim-repro/init.lua:12>                                                                                 
        [C]: in function 'nvim_exec_autocmds'                                                                                                                                                
        ...m-repro/site/pack/vnr/start/mini.nvim/lua/mini/files.lua:3096: in function 'trigger_event'                                                                                        
        ...m-repro/site/pack/vnr/start/mini.nvim/lua/mini/files.lua:1873: in function 'explorer_refresh_depth_window'                                                                        
        ...m-repro/site/pack/vnr/start/mini.nvim/lua/mini/files.lua:1566: in function 'explorer_refresh'                                                                                     
        ...m-repro/site/pack/vnr/start/mini.nvim/lua/mini/files.lua:2147: in function 'fn'                                                                                                   
        [string "vim/_core/editor"]:408: in function <[string "vim/_core/editor"]:407>    
```

Repro:

```lua
require('mini.files').setup({ windows = {  preview = true } })
  pattern = 'MiniFilesWindowUpdate',
  callback = function() MiniFiles.get_explorer_state() end,
})
```

1. Open nvim
2. Type `:lua MiniFiles.open()`
3. Type `o` and press `<Esc>`.
4. Type `u`

This happens because the preview buffer is [deleted](https://github.com/nvim-mini/mini.nvim/blob/15abccbe23525ee0502d2efe620c77db00bfb93c/lua/mini/files.lua#L1522). However, `the win_id` is still present in `explorer.windows`.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -1544,14 +1544,8 @@ H.explorer_refresh = function(explorer, opts)
     explorer = H.explorer_sync_cursor_and_branch(explorer, depth)
   end
 
-  -- Unregister windows from showed buffers, as they might get outdated
-  for _, win_id in ipairs(explorer.windows) do
-    -- NOTE: window can be invalid if it was showing buffer that was deleted
-    if H.is_valid_win(win_id) then
-      local buf_id = vim.api.nvim_win_get_buf(win_id)
-      H.opened_buffers[buf_id].win_id = nil
-    end
-  end
+  -- Unregister windows from shown buffers, as they might get outdated
+  explorer = H.explorer_unregister_windows(explorer)
 
   -- Compute depth range which is possible to show in current window
   local depth_range = H.compute_visible_depth_range(explorer, explorer.opts)
@@ -1681,6 +1675,20 @@ H.explorer_sync_cursor_and_branch = function(explorer, depth)
   local is_cur_buf = explorer.depth_focus == depth
   if show_preview and is_cur_buf then table.insert(explorer.branch, cursor_path) end
 
+  return explorer
+end
+
+H.explorer_unregister_windows = function(explorer)
+  local valid_windows = {}
+  for i, win_id in ipairs(explorer.windows) do
+    -- NOTE: window can be invalid if it was showing buffer that was deleted
+    if H.is_valid_win(win_id) then
+      local buf_id = vim.api.nvim_win_get_buf(win_id)
+      H.opened_buffers[buf_id].win_id = nil
+      table.insert(valid_windows, win_id)
+    end
+  end
+  explorer.windows = valid_windows
   return explorer
 end
 
--- a/tests/test_files.lua
+++ b/tests/test_files.lua
@@ -5733,6 +5733,19 @@ T['Events']['`MiniFilesWindowUpdate` can customize internally set window config 
   expect_screenshot()
 end
 
+T['Events']['MiniFilesWindowUpdate with preview works with `get_explorer_state()` after `undo`'] = function()
+  child.lua('MiniFiles.config.windows.preview = true')
+  child.lua([[
+    vim.api.nvim_create_autocmd('User', {
+      pattern = 'MiniFilesWindowUpdate',
+      callback = function() MiniFiles.get_explorer_state() end
+    })
+  ]])
+  open()
+  type_keys('o', '<Esc>', 'u')
+  eq(#get_explorer_state().windows, 2)
+end
+
 T['Events']['`MiniFilesActionCreate` triggers'] = function()
   track_event('MiniFilesActionCreate')
 


lua/mini/files.lua
```diff
@@ -1545,11 +1545,14 @@ H.explorer_refresh = function(explorer, opts)
...
 
   -- Unregister windows from showed buffers, as they might get outdated
-  for _, win_id in ipairs(explorer.windows) do
+  for i, win_id in ipairs(explorer.windows) do
     -- NOTE: window can be invalid if it was showing buffer that was deleted
     if H.is_valid_win(win_id) then
       local buf_id = vim.api.nvim_win_get_buf(win_id)
       H.opened_buffers[buf_id].win_id = nil
+    else
+      H.window_close()
```
echasnovski: I am pretty sure that this line without an argument [does nothing](https://github.com/nvim-mini/mini.nvim/blob/15abccbe23525ee0502d2efe620c77db00bfb93c/lua/mini/files.lua#L2566). So either this line is not needed or the test is not accurate enough.

lua/mini/files.lua
```diff
@@ -1545,11 +1545,14 @@ H.explorer_refresh = function(explorer, opts)
...
+      explorer.windows[i] = nil
```
echasnovski: In theory, this is not the cleanest assignment as it can introduce "holes" in the array (i.e. `explorer.windows[2] = nil` while `explorer.windows[3]` is not `nil`).

My first instinct tells me that the culprit is probably somewhere else.

echasnovski:
Thanks for the PR!

I can reproduce. And not only after `u` (which might indicate on some `undo` related issues), but also after moving cursor to the file that is present on disk.

lua/mini/files.lua
abeldekat: Indeed, the line is not needed. However, I do think the window should be closed, otherwise it might never will? I can add the `win_id` to the call....

lua/mini/files.lua
abeldekat: > My first instinct tells me that the culprit is probably somewhere else.

Interesting.
When I comment [this](https://github.com/nvim-mini/mini.nvim/blob/15abccbe23525ee0502d2efe620c77db00bfb93c/lua/mini/files.lua#L1522) line, the error does not happen.

lua/mini/files.lua
echasnovski: > ... otherwise it might never will?

The `H.is_valid_win(win_id) == false` means one of the two things: `win_id` is `nil` or `win_id` is already closed.

abeldekat: Ok. I will remove that line

lua/mini/files.lua
abeldekat: This behavior could potentially happen starting from [this](https://github.com/nvim-mini/mini.nvim/commit/1b32a3f61f1f) commit.
I think it wasn't noticed because the examples don't use `get_explorer_state` in `MiniFilesWindowUpdate`. I found it when playing with the code example from the [custom side scrolling](https://github.com/nvim-mini/mini.nvim/discussions/2173) discussion

abeldekat: > In theory, this is not the cleanest assignment as it can introduce "holes" in the array (i.e. explorer.windows[2] = nil while explorer.windows[3] is not nil).

I fixed that.

echasnovski:
After a more research, it looks like the culprit is that the events are triggered too early during the "explorer refresh" cycle. This makes code a bit cleaner, but has a downside of making `get_explorer_state()` act on not yet fully refreshed explorer state. The error also happens in `MiniFilesWindowOpen` event.

As usually triggering events when everything is up to date (be it "already up to date" for `open`/`changed` kind of events or "yet up to date" for `close`/`changedPre` kind of events) is a good idea, triggering both problematic events a bit later seems okay. This also shows that all the bookkeeping about the explorer state was correct.

Thanks again for finding this and making a PR!

PR #2444, opened 2026-06-02
fix(files): ensure `reveal_cwd` also works when `/` is `cwd`

### LINKED ISSUE #2443: Error: `branch` contains not a parent-child pair: "/" and "/home"`on reveal cwd when parent is top-level directory is /
### Contributing guidelines

### Module(s)

mini.files

### Neovim version

0.12.x

### Description

Error: `'branch' contains not a parent-child pair: "/" and "/home"`on `reveal cwd` when parent is top-level directory is /.

Expected behavior: do nothing.

### Reproduction

1. `:lua MiniFiles.open(vim.api.nvim_buf_get_name(0))`
2. Navigate leftmost to reach `/`.
3. `Reveal cwd`.

abeldekat:

Resolve #2443

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -1036,7 +1036,7 @@ MiniFiles.reveal_cwd = function()
   local branch, depth_focus = state.branch, state.depth_focus
 
   local cwd = H.fs_full_path(vim.fn.getcwd())
-  local cwd_ancestor_pattern = string.format('^%s/.', vim.pesc(cwd))
+  local cwd_ancestor_pattern = string.format('^%s%s.', vim.pesc(cwd), cwd == '/' and '' or '/')
   while branch[1]:find(cwd_ancestor_pattern) ~= nil do
     table.insert(branch, 1, H.fs_get_parent(branch[1]))
     depth_focus = depth_focus + 1
@@ -3054,7 +3054,7 @@ H.validate_branch = function(x)
   end
   for i = 2, #res do
     local parent, child = res[i - 1], res[i]
-    if (parent .. '/' .. child:match('[^/]+$')) ~= res[i] then
+    if H.fs_child_path(parent, child:match('[^/]+$')) ~= child then
       H.error('`branch` contains not a parent-child pair: ' .. vim.inspect(parent) .. ' and ' .. vim.inspect(child))
     end
   end
--- a/tests/test_files.lua
+++ b/tests/test_files.lua
@@ -1861,6 +1861,20 @@ T['reveal_cwd()']['works when root is already cwd'] = function()
   child.expect_screenshot()
 end
 
+T['reveal_cwd()']['works when cwd is `/`'] = function()
+  if not helpers.is_linux() then MiniTest.skip('Test is only for Linux.') end
+
+  child.fn.chdir('/')
+  open('/usr')
+  eq(get_explorer_state().branch, { '/usr' })
+  eq(get_explorer_state().depth_focus, 1)
+
+  reveal_cwd()
+  eq(get_explorer_state().branch, { '/', '/usr' })
+  eq(#get_explorer_state().windows, 2)
+  eq(get_explorer_state().depth_focus, 2)
+end
+
 T['reveal_cwd()']['properly places cursors'] = function()
   child.lua('MiniFiles.config.windows.width_focus = 20')
   local temp_dir =
@@ -2305,6 +2319,14 @@ T['set_branch()']['works with not absolute paths'] = function()
     set_branch({ '..' })
     eq(get_explorer_state().branch, { full_path(test_dir) })
   end
+end
+
+T['set_branch()']['works with `/` and a subdirectory'] = function()
+  if not helpers.is_linux() then MiniTest.skip('Test is only for Linux.') end
+
+  open('/')
+  set_branch({ '/', '/usr' })
+  expect.no_error(reveal_cwd)
 end
 
 T['set_branch()']['sets cursors on child entries'] = function()


lua/mini/files.lua
```diff
@@ -3054,7 +3054,7 @@ H.validate_branch = function(x)
   end
   for i = 2, #res do
     local parent, child = res[i - 1], res[i]
-    if (parent .. '/' .. child:match('[^/]+$')) ~= res[i] then
+    if (parent .. (parent == '/' and '' or '/') .. child:match('[^/]+$')) ~= child then
```
echasnovski: Comparing to the `child` instead of `res[i]` is a nice touch! However:

- This is more complicated that it should be (even the original code). I think something like `if H.fs_child_path(parent, child:match('[^/]+$')) ~= child` is better. The `res` already contains normalized paths, so it should not be a problem.
- This `reveal_cwd` not error, but it doesn't make it work to actually reveal the `/` as a root if it is not visible.

tests/test_files.lua
```diff
@@ -1861,6 +1861,14 @@ T['reveal_cwd()']['works when root is already cwd'] = function()
   child.expect_screenshot()
 end
 
+T['reveal_cwd()']['works when `/` and a subdirectory are open'] = function()
+  open(full_path('/'))
```
echasnovski: Although CI on Windows didn't fail, I think it needs to be guarded to only run on Linux (`if not helpers.is_linux() then MiniTest.skip('...') end`).

This also doesn't test that the current directory is revealed (which it doesn't and requires extra fix).

echasnovski (changes requested):
Thanks for the PR!

This does fix the error, but to fix the behavior there needs to be another adjustment in `reveal_cwd` itself.

@abeldekat, I am okay with you fixing this for educational purposes for a day or two. If you are short on time or motivation, I already have almost full fix with tests (started on it before seeing your comment in the issue).

lua/mini/files.lua
abeldekat: Done

tests/test_files.lua
abeldekat: Done

abeldekat:
> @abeldekat, I am okay with you fixing this for educational purposes for a day or two. If you are short on time or motivation, I already have almost full fix with tests (started on it before seeing your comment in the issue).

I indeed missed that the reveal does not work when `cwd` is `/`.
Having both time and motivation. Thanks!

lua/mini/files.lua
```diff
@@ -1036,7 +1036,7 @@ MiniFiles.reveal_cwd = function()
   local branch, depth_focus = state.branch, state.depth_focus
 
   local cwd = H.fs_full_path(vim.fn.getcwd())
-  local cwd_ancestor_pattern = string.format('^%s/.', vim.pesc(cwd))
+  local cwd_ancestor_pattern = string.format('^%s%s.', vim.pesc(cwd), (cwd == '/' and '' or '/'))
```
echasnovski:
```suggestion
  local cwd_ancestor_pattern = string.format('^%s%s.', vim.pesc(cwd), cwd == '/' and '' or '/')
```

tests/test_files.lua
```diff
@@ -1861,6 +1861,29 @@ T['reveal_cwd()']['works when root is already cwd'] = function()
   child.expect_screenshot()
 end
 
+T['reveal_cwd()']['works when root is `/`'] = function()
```
echasnovski:
```suggestion
T['reveal_cwd()']['works when cwd is `/`'] = function()
```

In 'mini.files' implementation "root" is the path at the branch start. The test is about when `/` is both current directory and *not* root calling `reveal_cwd` will make it root.

tests/test_files.lua
```diff
@@ -1861,6 +1861,29 @@ T['reveal_cwd()']['works when root is already cwd'] = function()
...
+  if not helpers.is_linux() then MiniTest.skip('Test is only for Linux.') end
+
+  child.fn.chdir('/')
+  open(full_path('/', 'usr'))
```
echasnovski:
```suggestion
  open('/usr')
```

tests/test_files.lua
```diff
@@ -1861,6 +1861,29 @@ T['reveal_cwd()']['works when root is already cwd'] = function()
...
+end
+
+T['reveal_cwd()']['works when `/` and a subdirectory are open'] = function()
+  if not helpers.is_linux() then MiniTest.skip('Test is only for Linux.') end
+
+  open(full_path('/'))
+  go_in()
+  -- Ensure no error: `branch` contains not a parent-child pair: "/" and "/<first_subdir>"
+  expect.no_error(reveal_cwd)
+end
```
echasnovski: I don't think this tests the error reported by the issue and fixed in the PR.

It is better to repurpose for the `set_branch()`. Like that using `{ '/', '/usr' }` as a branch on Linux works as expected.

echasnovski (changes requested):
Yeah, this looks much better now!

lua/mini/files.lua
abeldekat: Done

tests/test_files.lua
abeldekat: Done

tests/test_files.lua
abeldekat: Done

tests/test_files.lua
abeldekat: Done

echasnovski:
Thanks!

I've adjusted the `set_branch()` test to actually test that it successfully sets the expected branch. It is also a test for the error since the test would not have been successful if there was an error.

PR #2467, opened 2026-06-17
fix(files): ensure `BufWinEnter` after `:edit <dir>`

### LINKED ISSUE #2466: When closing mini.files after `:edit <dir>` clues sometimes don't work
### Contributing guidelines

### Module(s)

mini.clue, mini.files

### Neovim version

0.13 (!at least latest Nightly build!)

### Description

When closing mini.files after `:edit <dir>` clues don't work when there is no alternate file.

Expected: if the target buffer is listed, `mini.clues` always attaches after closing the default explorer.

EDIT: The same behavior can be observed when 'mini.starter' is included.

### Reproduction

```lua

vim.pack.add({ 'https://github.com/nvim-mini/mini.nvim' })

local minimisc = require('mini.misc')
local later = function(f) minimisc.safely('later', f) end

vim.g.mapleader = ' '
require('mini.files').setup()
later(
  function()
    require('mini.clue').setup({
      clues = {
        { mode = 'n', keys = '<Leader>e', desc = '+Explore/Edit' },
      },
      triggers = {
        { mode = { 'n', 'x' }, keys = '<Leader>' },
      },
    })
  end
)

```

4. Steps

  - Open nvim
  - Press leader and note that the clue activates
  - Type `:edit .`
  - Type `q`
  - Press leader and note that the clue does not activate
   
</details>

abeldekat:

Details:
- By default, 'mini.files' replaces 'netrw'. When a directory is opened with `:edit <dir>`, and the explorer is closed without opening a file, 'mini.files' shows the alternate buffer if present. If there is no alternate buffer, a new buffer is created. Event `BufWinEnter` needs to fire in order for `mini.clue` to attach.

Resolve #2466

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -1381,7 +1381,9 @@ H.create_autocommands = function(config)
     vim.cmd('silent! autocmd! FileExplorer *')
     vim.cmd('autocmd VimEnter * ++once silent! autocmd! FileExplorer *')
 
-    au('BufEnter', '*', H.track_dir_edit, 'Track directory edit')
+    -- - Use `nested` to allow other events (`BufWinEnter` for 'mini.clue')
+    local opts = { nested = true, group = gr, callback = H.track_dir_edit, desc = 'Track directory edit' }
+    vim.api.nvim_create_autocmd('BufEnter', opts)
   end
 
   au('VimResized', '*', MiniFiles.refresh, 'Refresh on resize')
--- a/tests/test_files.lua
+++ b/tests/test_files.lua
@@ -6324,6 +6324,13 @@ T['Default explorer']['works in `:tabfind .`'] = function()
 end
 
 T['Default explorer']['handles close without opening file'] = function()
+  child.lua([[
+    _G.log = {}
+    vim.api.nvim_create_autocmd('BufWinEnter', {
+      callback = function(data) table.insert(_G.log, data.buf) end,
+    })
+  ]])
+
   local validate = function()
     local buf_name = child.api.nvim_buf_get_name(0)
     child.cmd('edit ' .. test_dir_path)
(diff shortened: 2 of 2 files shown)


echasnovski:
Thanks for the PR!

I can indeed reproduce the issue, however the solution is not quite right. Usually if there is a problem with some event not triggering, two most probable causes are: 1) it is executed inside an autocommand without `nested = true`; 2) there is an explicit/intentional 'eventignore'. The first one is the culprit in this case, as `nvim_buf_delete()` is executed inside a callback (`H.track_dir_edit`) for `BufEnter` function. Modifying the autocommand creation to include `nested = true` fixes the issue here.

Also note, that merging this will have to wait until 0.18.0 release.

abeldekat:
> The first one is the culprit in this case, as nvim_buf_delete() is executed inside a callback (H.track_dir_edit) for BufEnter function.

Initially I worried that `nested` would be propagated to the code in `MiniFiles.open` causing side-effects. I missed that in `H.track_dir_edit` it is `MiniFiles.open` itself that is scheduled. Thanks!

lua/mini/files.lua
```diff
@@ -1371,19 +1371,20 @@ H.apply_config = function(config) MiniFiles.config = config end
...
     vim.cmd('silent! autocmd! FileExplorer *')
     vim.cmd('autocmd VimEnter * ++once silent! autocmd! FileExplorer *')
 
-    au('BufEnter', '*', H.track_dir_edit, 'Track directory edit')
+    -- - Use `nested` to allow other events (`BufWinEnter` for 'mini.clue')
+    local opts = { nested = true, group = gr, callback = H.track_dir_edit, desc = 'Track directory edit' }
+    vim.api.nvim_create_autocmd('BufEnter', opts)
   end
 
+  local au = function(event, pattern, callback, desc)
```
echasnovski: Please, don't move this function definition without explicit reason. The `H.create_autocommands` is a common function in modules which aims to have as unified code inside of it as possible. In all modules if there is a need for the autocomman creation helper, it is created directly below `local gr = ...` using exactly the same code lines.

abeldekat: done

echasnovski (approved):
I'll adjust the test to be a bit more targeted. I don't think it matters how many times `BufWinEnter` was triggered, as long as it was triggered for the newly created buffer. **Edit**: or rather that it is triggered specifically after closing an explorer without opening a file.

echasnovski:
This should now be part of the latest `main`. The adjusted test looks [like](https://github.com/nvim-mini/mini.nvim/blob/c594be5b095204e99473825ec2dee872e7409a4d/tests/test_files.lua#L6338) [this](https://github.com/nvim-mini/mini.nvim/blob/c594be5b095204e99473825ec2dee872e7409a4d/tests/test_files.lua#L6344-L6346).

PR #2469, opened 2026-06-20
fix(notify): make window opaque if $DISPLAY is unset

yilisharcs:


echasnovski:
Thanks for the PR!

This looks like something that is possible for the user to configure on their own: `require('mini.notify').setup({ window = { winblend = vim.env.DISPLAY and 25 or 0 } })`. Together with it not being quite as ubiquitous as needed for this change of default value, I'd say this is not planned.

PR #2470, opened 2026-06-20
feat(splitjoin): support lisp-like indentation

yilisharcs:
i made this specifically for fennel. i don't know enough to make a generic way for all lisps, so for the time being (or forever), the user should be expected to use a hook, like so:

```lua
local function kv_split(positions)
        local n = #positions
        if n < 3 then
                return positions
        end

        local first = positions[1]
        local bracket = vim.api.nvim_buf_get_lines(
                0,
                first.line - 1,
                first.line,
                false
        )[1]:sub(first.col, first.col)

        if bracket == "{" then
                local res = { positions[1] }
                for i = 2, n - 1 do
                        if i % 2 == 1 then
                                table.insert(res, positions[i])
                        end
                end
                table.insert(res, positions[n])
                return res
        end

        return positions
end

vim.b.minisplitjoin_config = {
        detect = {
                brackets = { "%b[]", "%b{}" },
                separator = "%s+",
        },
        split = {
                indent = "lisp",
                hooks_pre = { kv_split },
        },
}
```

### DIFF
--- a/doc/mini-splitjoin.txt
+++ b/doc/mini-splitjoin.txt
@@ -140,6 +140,11 @@ Defaults ~
     split = {
       hooks_pre = {},
       hooks_post = {},
+
+      -- Indentation style for split inner lines
+      --   'shiftwidth' - use `shiftwidth()` for each level.
+      --   'lisp' - align to column of the first element after opening bracket.
+      indent = 'shiftwidth',
     },
 
     -- Join options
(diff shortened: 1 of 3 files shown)


echasnovski:
Thanks for the PR!

I am not inherently opposed to adding customization of how split content is indented, but this feels like both too much (adds a new config value) and not enough (still requires a lot of code on the user side).

For the time being, I believe this should be possible to achieve via custom post-split hooks. If you'll have the solution along these lines, please open a ["Show and tell"](https://github.com/nvim-mini/mini.nvim/discussions/categories/show-and-tell) discussion. If not, then please create a feature request with detailed explanation of how the new lisp-like indent would look like (very much preferred with examples).

PR #2477, opened 2026-06-23
feat(test): add option in `gen_reporter.stdout` to control ANSI output

tunaflsh:
Addresses #2473


echasnovski:
Thanks for the PR!

In the #2473 [I said](https://github.com/nvim-mini/mini.nvim/discussions/2473#discussioncomment-17408129) that I'll think about it. It means that some time has to pass before making the decision about adding this feature. Maybe there might be a better way to handle this.

Plus it requires a test and a bit more documentation. So I'll close this for now.

PR #2478, opened 2026-06-24
fix(files): avoid closing directory buffer when #windows decreases

abeldekat:

https://github.com/user-attachments/assets/ade11fd4-8e7e-427a-a1e4-351f0a39b34d

Whilst working on [this](https://github.com/nvim-mini/mini.nvim/discussions/2448#discussioncomment-17362786) discussion, I saw inconsistencies with regard to the closing of the underlying directory buffer.

When decreasing the number of windows, there are circumstances where the directory buffer is closed. 

```lua

-- Only added for visual feedback:
require('mini.tabline').setup()
require('mini.statusline').setup()

require('mini.files').setup()

```

It happens when 2 or more windows are closed.

Scenario:

- Open nvim
- Type `:edit .`
- Type `ll`
- Notice that the directory buffer is still open
- Type `<`
- Notice that the directory buffer is closed

The video demonstrates "depth_focus = 3", using both `MiniFiles.refresh()` and `MiniFiles.trim_left()`. The second part shows the same scenario with the solution applied.

In the source code, there are two locations where the behavior is triggered:
- [ refresh, explorer normalize](https://github.com/nvim-mini/mini.nvim/blob/c594be5b095204e99473825ec2dee872e7409a4d/lua/mini/files.lua#L1645)
- [ refresh, after the loop](https://github.com/nvim-mini/mini.nvim/blob/c594be5b095204e99473825ec2dee872e7409a4d/lua/mini/files.lua#L1574)

Unfortunately, I don't understand why [BufEnter](https://github.com/nvim-mini/mini.nvim/blob/c594be5b095204e99473825ec2dee872e7409a4d/lua/mini/files.lua#L1425) is triggered prematurely on the directory buffer. I assume it has to do with alternate files.

I fixed it by ensuring that in `BufEnter` the directory buffer is only explicitly closed if the explorer is closed.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -922,11 +922,6 @@ MiniFiles.close = function()
   -- Trigger appropriate event
   H.trigger_event('MiniFilesExplorerClose')
 
-  -- Focus on target window
-  explorer = H.explorer_ensure_target_window(explorer)
-  -- - Use `pcall()` because window might still be invalid
-  pcall(vim.api.nvim_set_current_win, explorer.target_window)
-
   -- Update currently shown cursors
   explorer = H.explorer_update_cursors(explorer)
 
@@ -950,6 +945,11 @@ MiniFiles.close = function()
   -- Update histories and unmark as opened
   H.explorer_path_history[explorer.anchor] = explorer
   H.opened_explorers[explorer.tabpage_id] = nil
+
+  -- Focus on target window
+  explorer = H.explorer_ensure_target_window(explorer)
+  -- - Use `pcall()` because window might still be invalid
+  pcall(vim.api.nvim_set_current_win, explorer.target_window)
 
   -- Return `true` indicating success in closing
   return true
@@ -1427,6 +1427,9 @@ H.track_dir_edit = function(data)
   if vim.api.nvim_get_current_buf() ~= data.buf then return end
 
   if vim.b.minifiles_processed_dir then
+    -- Only delete directory buffer explicitly when explorer is closed
+    if H.explorer_get() ~= nil then return end
+
     -- Smartly delete directory buffer if already visited
     local alt_buf = vim.fn.bufnr('#')
     -- - Setting alternative buffer is enough for the "directory buffer" to be
--- a/tests/test_files.lua
+++ b/tests/test_files.lua
@@ -6369,6 +6369,32 @@ T['Default explorer']['handles forcing other window as current'] = function()
   if child.fn.has('nvim-0.10') == 1 then eq(child.api.nvim_get_current_win(), init_win_id) end
 end
 
+T['Default explorer']['keeps directory buffer open when number of windows decreases'] = function()
+  local validate = function(fn_str)
+    child.cmd('edit ' .. make_test_path('nested'))
+
+    local buffers = child.api.nvim_list_bufs()
+    local dir_buffer_id = buffers[1]
+    eq(child.fn.isdirectory(child.api.nvim_buf_get_name(dir_buffer_id)), 1)
+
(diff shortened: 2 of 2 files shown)


echasnovski:
Hmm... The behavior indeed seems strange, although I don't think it is harmful. Are there any actual problems that come from deleting the "directory buffer" early? Or is it only "doesn't feel right" kind of behavior?

I, of course, don't mind it fixing, only trying to understand the severity. Especially since it requires some extra checks that are not obviously needed.

abeldekat:
There are no actual problems that come from closing the directory buffer early. It is surprising and a bit disorienting for the user, once he sees it happening.

In [this](https://github.com/nvim-mini/mini.nvim/discussions/2448#discussioncomment-17362786) discussion I added a full screen toggle, often triggering the behavior.

abeldekat:
There is an alternative solution:

```diff
diff --git a/lua/mini/files.lua b/lua/mini/files.lua
index b92a2567..b2469d78 100644
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -1572,6 +1572,7 @@ H.explorer_refresh = function(explorer, opts)
   -- Close possibly opened window that don't fit (like after `VimResized`)
   for depth = cur_win_count + 1, #explorer.windows do
     H.window_close(explorer.windows[depth])
+    vim.api.nvim_set_current_win(explorer.windows[cur_win_count])
     explorer.windows[depth] = nil
   end
 
@@ -1643,6 +1644,7 @@ H.explorer_normalize = function(explorer)
   -- still get outdated later if branch is too deep to fit into Neovim's width.
   for i = cur_max_depth + 1, #explorer.windows do
     H.window_close(explorer.windows[i])
+    vim.api.nvim_set_current_win(explorer.windows[cur_max_depth])
     explorer.windows[i] = nil
   end

```

echasnovski:
> There is an alternative solution:

Only by reading the diff, this seems very much too much: 1) it executes the same operation possibly several times; 2) setting current window might trigger relevant events, so should be done only when truly needed; 3) it looks like it might set current window to the last present window even if it is not the "true" current window.

echasnovski:
I've taken a look at this and the issue is a bit more nuanced than the fix in this PR. A general rule of thumb is that current window should be set before closing any window. Otherwise closing the current window will temporarily switch focus to some other window (probably the first "normal" window in the tabpage).

Adjusting `close()` to postpone setting the current window after all explorer ones are closed somewhat follows the same pattern that lead to the current issue: it will guarantee to switch focus to some other window before setting the desired current window. In some cases it will be the same window, in some - not.

The more fundamental solution is to ensure that closing of a window in explorer never leaves the focus from explorer. I think it is achievable even with negative code diff without major consequences (although I might regret saying this; at least all tests pass). So I merged this in the temporary branch to adjust, as the test is mostly good.

echasnovski:
Here is the commit that should fix this: f73ec147427fe998823a5e6b31b16bbfc3bc8ee0

---

I adjusted the test to have a bit less comments (they usually show the intent of codeblock/subtest or clarify some *really* non-obvious behavior) and use `child.fn.bufnr(test_path)` to find the target buffer (instead of relying that it is always `vim.api.nvim_list_bufs()[1]`).

abeldekat:
Thanks! The solution is much better indeed and makes sense.

I did try to set the focus before closing any windows, but it did not work. Skill issues...)

echasnovski:
> I did try to set the focus before closing any windows, but it did not work. Skill issues...)

Well, there was another not obvious place where windows got closed. Making sure that that is not needed is not trivial. I am still not 100% confident it will not backfire.

Issue #1818, opened 2025-05-22
Improve 'mini.jump2d' default config and jumping experience


echasnovski:
### Contributing guidelines

- [x] I have read [CONTRIBUTING.md](https://github.com/echasnovski/mini.nvim/blob/main/CONTRIBUTING.md)
- [x] I have read [CODE_OF_CONDUCT.md](https://github.com/echasnovski/mini.nvim/blob/main/CODE_OF_CONDUCT.md)

### Module(s)

mini.jump2d

### Description

- Update default `config.view` to have `dim = true` and `n_steps_ahead = 2` (or 1). These were not default because they were not present in the initial release and their values preserved the current at the time behavior.
- Consider having default `config.labels` as `'asdfghjkl;'`, i.e. 10 "home row" keys. After test driving them it indeed feels more natural and faster to type, but average label length of targets does go up (less labels -> longer sequences).
- Consider handling `<BS>` during sequence enter to undo previously typed label.
- Add something like a "smart jump":
    - May or may not have a separate `config.mappings.smart_jumping` entry. I currently very much like `sj`: it is in "home row" and works well with 'mini.surround'. It feels better than `<CR>` (because hands stay on the home row which makes further typing faster) and is *almost* feels the same as a single `s` key (which doesn't work with 'mini.surround' and is a bit wasteful).
    - Requires pressing a single character after initiating. That character represents a "jumper identifier" in a concept similar to "textobject/surrounding identifier" from 'mini.ai' and 'mini.surround'. A single jumper is essentially a `start()` table input or function returning it. This allows customizing *everything* for any jumper separately: `view`, `labels`, `allowed_lines`, etc.
    - A default jumper reused config values and sets `spotter` to jump to a cell containing identifier, like `builtin_opts.single_character`. Should also respect 'ignorecase' and 'smartcase'.
    - Something like `<BS>` can be used as "return to previous position", which is a more flexible (yet verbose) way of doing "remote operations": jump (like `sj`-`<char>`-`<seq>`) - do operation - `sj<BS>`. May require exported `MiniJump2d.jump_back()`.
    - Maybe `.` identifier can be used as "jump to any punctuation group".
    - Maybe `<Space>` identifier can be used as "jump to any whitespace or line start".
    - Add `config.custom_jumpers` to customize jumpers for this "smart jumping".

References: #1033, #1755

cc @pkazmier and @abeldekat about possible ideas/feedback here.

echasnovski:
[...]

PR #2482, opened 2026-06-26
fix(jump2d): ensure that `CR` works in the new cmdwin on Nightly

abeldekat:

Details:
- See PR [40325](https://github.com/neovim/neovim/pull/40325) in neovim/neovim

### DIFF
--- a/lua/mini/jump2d.lua
+++ b/lua/mini/jump2d.lua
@@ -790,7 +790,11 @@ H.create_autocommands = function(config)
 
   -- Corrections for default `<CR>` mapping to not interfere with popular usages
   if config.mappings.start_jumping == '<CR>' then
-    local revert_cr = function() vim.keymap.set('n', '<CR>', '<CR>', { buffer = true }) end
+    local revert_cr = function()
+      -- Don't revert when there is an existing buffer local mapping
+      if vim.fn.maparg('<CR>', 'n', false, true).buffer == 1 then return end
+      vim.keymap.set('n', '<CR>', '<CR>', { buffer = true })
+    end
     au('FileType', 'qf', revert_cr, 'Revert <CR>')
     au('CmdwinEnter', '*', revert_cr, 'Revert <CR>')
   end
--- a/tests/test_jump2d.lua
+++ b/tests/test_jump2d.lua
@@ -206,6 +206,12 @@ T['setup()']['resets <CR> mapping in quickfix window'] = function()
   child.cmd([[cexpr ['Hello', 'Quickfix'] | copen]])
   -- Should create not remappable buffer-local mapping
   expect.match(child.cmd_capture('nmap <CR>'), '%*@<CR>')
+
+  child.cmd('au FileType qf nnoremap <buffer> <CR> <Cmd><CR>')
+  reload_module()
+  child.cmd([[cexpr ['Hello', 'Quickfix'] | copen]])
+  -- Should not create buffer-local mapping if its already present
+  expect.no_match(child.cmd_capture('nmap <CR>'), '%*@<CR>')
 end
 
 T['setup()']['resets <CR> mapping in command-line window'] = function()
@@ -214,7 +220,15 @@ T['setup()']['resets <CR> mapping in command-line window'] = function()
   type_keys('q:')
   set_cursor(1, 0)
   type_keys('<CR>')
+  -- Should create not remappable buffer-local mapping if needed
   eq(child.get_lines(), { 'Hello', '' })
+
+  child.cmd('au CmdwinEnter * nnoremap <buffer> <CR> <Cmd><CR>')
+  type_keys('q:')
+  set_cursor(1, 0)
+  type_keys('<CR>')
+  -- Should not create buffer-local mapping if its already present
+  eq(child.get_lines(), { [[call append(0, 'Hello')]], '' })
 end
 
 T['start()'] = new_set({


lua/mini/jump2d.lua
```diff
@@ -792,7 +792,7 @@ H.create_autocommands = function(config)
   if config.mappings.start_jumping == '<CR>' then
     local revert_cr = function() vim.keymap.set('n', '<CR>', '<CR>', { buffer = true }) end
     au('FileType', 'qf', revert_cr, 'Revert <CR>')
-    au('CmdwinEnter', '*', revert_cr, 'Revert <CR>')
+    if vim.fn.has('nvim-0.13') == 0 then au('CmdwinEnter', '*', revert_cr, 'Revert <CR>') end
```
echasnovski: I think it is a good opportunity to make `revert_cr` more robust: it should not override already present buffer-local mapping. This should fix both Neovim>=0.13 new Cmdwin fail and possible future not related ones. Should also probably improve both quickfix and cmdwin tests to test that this does not override a buffer-local mapping.

tests/test_jump2d.lua
```diff
@@ -208,15 +208,6 @@ T['setup()']['resets <CR> mapping in quickfix window'] = function()
   expect.match(child.cmd_capture('nmap <CR>'), '%*@<CR>')
 end
 
-T['setup()']['resets <CR> mapping in command-line window'] = function()
```
echasnovski, line 211: Why the move? The logic here is that this behavior is enabled in `setup()`, not in `start`. Using `'setup()'` test set is the common pattern to test autocommands that are created in `setup()`.

echasnovski (changes requested):
Although this works, I think this is a good opportunity to make a more fundamental fix here.

lua/mini/jump2d.lua
abeldekat: Done

tests/test_jump2d.lua
abeldekat: I was thinking: On neovim 0.13 setup does *not* reset ....

Thinking a bit too much perhaps...)

tests/test_jump2d.lua
```diff
@@ -206,6 +206,16 @@ T['setup()']['resets <CR> mapping in quickfix window'] = function()
   child.cmd([[cexpr ['Hello', 'Quickfix'] | copen]])
   -- Should create not remappable buffer-local mapping
   expect.match(child.cmd_capture('nmap <CR>'), '%*@<CR>')
+
+  child.lua([[
+    vim.api.nvim_create_autocmd('FileType', { pattern = 'qf', callback = function()
+      vim.keymap.set('n', '<CR>', '<cmd><cr>', { buffer = true })
+    end })
+  ]])
```
echasnovski:
```suggestion
  vim.cmd('au FileType qf nnoremap <buffer> <CR> <Cmd><CR>')
```

Vimscript syntax is more concise when it matters :)

tests/test_jump2d.lua
```diff
@@ -206,6 +206,16 @@ T['setup()']['resets <CR> mapping in quickfix window'] = function()
...
+  reload_module()
```
echasnovski, line 211: I am not sure, but why reloading a module is needed here? I'd expect that closing previous quickfix list would be enough to test that custom buffer-local mapping is not overridden in the next quickfix buffer.

lua/mini/jump2d.lua
```diff
@@ -790,7 +790,12 @@ H.create_autocommands = function(config)
 
   -- Corrections for default `<CR>` mapping to not interfere with popular usages
   if config.mappings.start_jumping == '<CR>' then
-    local revert_cr = function() vim.keymap.set('n', '<CR>', '<CR>', { buffer = true }) end
+    local revert_cr = function()
+      -- Don't revert when there is an existing buffer local mapping
+      if vim.fn.maparg('<CR>', 'n', false, true).buffer == 1 then return end
+
+      vim.keymap.set('n', '<CR>', '<CR>', { buffer = true })
```
echasnovski, line 796:
```suggestion
      if vim.fn.maparg('<CR>', 'n', false, true).buffer == 1 then return end
      vim.keymap.set('n', '<CR>', '<CR>', { buffer = true })
```

echasnovski (changes requested):
Nice change! The `vim.fn.maparg('<CR>', 'n', false, true).buffer == 1` indeed seems to be the most robust+concise way to detect a buffer-local mapping.

tests/test_jump2d.lua
abeldekat: Done..:)

tests/test_jump2d.lua
abeldekat: Without the reload, I cannot make the test to fail, when I comment out the `if vim.fn.maparg`...
I was able to remove 'reload' from the test regarding `cmdwin` though

lua/mini/jump2d.lua
abeldekat: Done

tests/test_jump2d.lua
echasnovski: Ah, that's because of the autocommand order. The`FileType qf` autocommand for `revert_cr` is created during `setup()`. If it is before `FileType qf nnoremap ...` from test, then it "can't see" the buffer-local mapping as it is not yet created. Reloading the module creates the `revert_cr` autocommand after the test's one, so it *can* no see it.

Let's leave the `reload_module()` for now. It will become obsolete after #1818 (which I do want to implement sooner rather than later), as the default key will be `sj`.

echasnovski:
This should now be part of the `main` branch as a regular `fix(jump2d)` and not necessarily a reaction to the Nightly change (which now would not fail the previous tests as `<CR>` is no longer a buffer-local mapping).

PR #2483, opened 2026-06-26
fix(clue): ensure that clues work with the new cmdwin on Nightly

abeldekat:

Details:
- See PR [40325](https://github.com/neovim/neovim/pull/40325) in neovim/neovim

### DIFF
--- a/lua/mini/clue.lua
+++ b/lua/mini/clue.lua
@@ -1248,9 +1248,11 @@ H.create_autocommands = function()
   local did_ensure = {}
   local ensure_triggers = vim.schedule_wrap(function(ev)
     if not H.is_valid_buf(ev.buf) then return end
-    local skip_triggers = ev.event == 'BufWinEnter' and (did_ensure[ev.buf] or vim.fn.buflisted(ev.buf) ~= 1)
+    -- NOTE: Special case command-line window as it is not listed on Neovim>=0.13
+    local is_cmdwin = vim.fn.getbufinfo(ev.buf)[1].command == 1
+    local skip_triggers = did_ensure[ev.buf] or not (vim.bo[ev.buf].buflisted or is_cmdwin)
     did_ensure[ev.buf] = true
-    if skip_triggers then return end
+    if ev.event == 'BufWinEnter' and skip_triggers then return end
     MiniClue.ensure_buf_triggers(ev.buf)
   end)
   -- - Respect `LspAttach` as it is a common source of buffer-local mappings
--- a/tests/test_clue.lua
+++ b/tests/test_clue.lua
@@ -1753,6 +1753,10 @@ T['Showing keys']['works with multibyte characters'] = function()
 end
 
 T['Showing keys']['works in Command-line window'] = function()
+  local expect_screenshot = child.expect_screenshot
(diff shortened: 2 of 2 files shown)


lua/mini/clue.lua
```diff
@@ -1246,9 +1246,10 @@ H.create_autocommands = function()
   -- delay acting until buffer is loaded (otherwise buffer-local options can be
   -- prematurely "finalized"). Process it at most once for performance.
   local did_ensure = {}
+  local not_listed = function(buf_id) return vim.fn.buflisted(buf_id) ~= 1 and vim.fn.getbufinfo()[buf_id].command ~= 1 end
```
echasnovski: 1. `vim.fn.getbufinfo()[buf_id]` is not the correct way of getting buffer information, as the function returns array and not a map. It is `vim.fn.getbufinfo(buf_id)[1]`.
2. I don't think I like the `not_listed` name and there is no obvious name for this. So I'd go with something like this:

    ```lua
    local did_ensure = {}
    local ensure_triggers = vim.schedule_wrap(function(ev)
      if not H.is_valid_buf(ev.buf) then return end
      -- NOTE: Special case command-line window as it is not listed on Neovim>=0.13
      local is_cmdwin = vim.fn.getbufinfo(ev.buf)[1].command == 1
      local skip_triggers = did_ensure[ev.buf] or not (vim.bo[ev.buf].buflisted or is_cmdwin)
      did_ensure[ev.buf] = true
      if ev.event == 'BufWinEnter' and skip_triggers then return end
      MiniClue.ensure_buf_triggers(ev.buf)
    end)
    ```

tests/test_clue.lua
```diff
@@ -1753,21 +1753,25 @@ T['Showing keys']['works with multibyte characters'] = function()
 end
 
 T['Showing keys']['works in Command-line window'] = function()
+  local expect_screenshot = child.expect_screenshot
+  -- Skip nvim-0.13 for now. The output of the new cmdline differs slightly.
+  if child.fn.has('nvim-0.13') == 1 then expect_screenshot = function() end end
```
echasnovski, line 1758: Skipping testing for all future versions should usually be done only if it is not needed. Here the test is needed, so the ideal approach is to make the test pass on both Nightly and non-Nightly. Unfortunately, this is somewhat verbose for screenshots, so the usual approach is to redo screenshots to be future compatible (i.e. on Nightly) and skip tests on not-Nightly.

However, in this case I'd wait a bit to see how and if neovim/neovim#40431 is resolved.

echasnovski (changes requested):

lua/mini/clue.lua
abeldekat: Done

tests/test_clue.lua
abeldekat: Thanks. For now, I kept the test as is.

echasnovski:
As neovim/neovim#40431 resolved the actual problem, the actual fix does not needed anymore. Test still fails because of some visual difference, but I'll report upstream and patch the screenshot expectation myself. Thanks for the PR anyway!

PR #2508, opened 2026-07-11
test(pick): adjust `validate_buf_name()` after Nightly changes

abeldekat:

Details:
- On Nightly, the name of a directory buffer always ends with '/'.
- See PR [40552](https://github.com/neovim/neovim/pull/40552) in neovim/neovim.

### DIFF
--- a/tests/test_pick.lua
+++ b/tests/test_pick.lua
@@ -70,6 +70,11 @@ local validate_buf_name = function(buf_id, ref_name)
   local name = child.api.nvim_buf_get_name(buf_id):gsub('\\', '/')
   ref_name = ref_name ~= '' and full_path(ref_name) or ''
   ref_name = ref_name:gsub('\\', '/'):gsub('[\\/]+$', '')
+
+  if child.fn.has('nvim-0.13') == 1 then
+    local ft = child.api.nvim_get_option_value('filetype', { buf = buf_id })
+    ref_name = ref_name .. (ft == 'directory' and '/' or '')
+  end
   eq(name, ref_name)
 end
 


tests/test_pick.lua
```diff
@@ -70,6 +70,11 @@ local validate_buf_name = function(buf_id, ref_name)
   local name = child.api.nvim_buf_get_name(buf_id):gsub('\\', '/')
   ref_name = ref_name ~= '' and full_path(ref_name) or ''
   ref_name = ref_name:gsub('\\', '/'):gsub('[\\/]+$', '')
+
+  if child.fn.has('nvim-0.13') == 1 then
+    local ft = child.lua_get(string.format('vim.bo[%d].filetype', buf_id))
+    ref_name = ft == 'directory' and ref_name .. '/' or ref_name
```
echasnovski: This looks more aligned with how other modules do this:

```suggestion
    local ft = child.api.nvim_get_option_value('filetype', { buf = buf_id })
    ref_name = ref_name .. (ft == 'directory' and '/' or '')
```

(For a long time I wanted to see if it is feasible to make `child.bo[buf_id].option_name` syntax possible. Probably yes, but not sure).

abeldekat: Done.

It took me some time to figure out that the `child.bo[]` syntax *really* is not possible. 
I also think `ref_name = ref_name .. (ft == 'directory' and '/' or '') ` is more readable. However, doesn't that change always perform a string concatenation? It's hard to weigh the pros and cons between readability and (micro) efficiency.

echasnovski: > It's hard to weigh the pros and cons between readability and (micro) efficiency.

Microoptimization is almost never a concern in tests. So there is that ;)

echasnovski:
Thanks for the PR!

After taking a closer look, I think it is better to always normalize real and reference buffer names in the same way. I.e. always remove trailing `/` from real buffer name also.

Plus there are several `validate_buf_name` across tests and I think it is a good idea to use the same helper. Which I did during my current round of improving test consistency (will probably merge tomorrow). So this PR will be superseded. Hope that's okay.

abeldekat:
Yes, much better this way. Thanks!

PR #2515, opened 2026-07-18
fix(files): ensure that file-preview includes the first column

### LINKED ISSUE #2514: [mini.files] preview sometimes starts at a non-zero column position
### Contributing guidelines

### Module(s)

mini.files

### Neovim version

0.12.x

### Description

The preview window sometimes opens with a non-zero column position, so the beginning of the file content is not visible.

Expected behavior:
The preview should always start at column 0 when displaying a file.

### Reproduction

1. create 'init.lua' and populate it with the following content:

    ```lua
    vim.pack.add({ "https://github.com/nvim-mini/mini.nvim" })

    vim.g.mapleader = ' ' -- Use `<Space>` as <Leader> key

    require("mini.files").setup({ windows = { preview = true } })

    local nmap_leader = function(suffix, rhs, desc)
	    vim.keymap.set("n", "<Leader>" .. suffix, rhs, { desc = desc })
    end

    local explore_at_file = "<Cmd>lua MiniFiles.open(vim.api.nvim_buf_get_name(0))<CR>"

    nmap_leader("ef", explore_at_file, "File directory")
    ```

3. Open a C++ file using `nvim *.cpp`, then press `<Leader>ef`. The preview window starts at a non-zero column position. Press `q` to exit, then press `<Leader>ef` again. The previously previewed file now starts at column 0, while an unpreviewed file still starts at a non-zero colum. The behavior is demonstrated in the attached video.

https://github.com/user-attachments/assets/6ede0ec3-6587-4620-acce-5b04267cc2e1

abeldekat:

Resolve #2514

Test `Preview works for files` has been modified to also include the case where the preview did not start from zero cursor position. As a consequence, I had to regenerate some screenshots in test cases also using the `real_path` directory.

### DIFF
--- a/lua/mini/files.lua
+++ b/lua/mini/files.lua
@@ -2577,9 +2577,13 @@ H.window_set_view = function(win_id, view)
   buf_data.win_id = win_id
 
   -- Set cursor (if defined), visible only in directories
-  pcall(H.window_set_cursor, win_id, view.cursor)
+  local ok_cursor = pcall(vim.api.nvim_win_set_cursor, win_id, view.cursor)
+  -- - Tweak cursor here, before `CursorMoved`, event to reduce flicker
+  local is_dir = H.fs_get_type(buf_data.path) == 'directory'
+  if ok_cursor and is_dir then pcall(H.window_tweak_cursor, win_id, buf_id) end
+
   -- NOTE: set 'cursorline[opt]' here because changing buffer might remove it
-  vim.wo[win_id].cursorline = H.fs_get_type(buf_data.path) == 'directory'
+  vim.wo[win_id].cursorline = is_dir
   local culopt = vim.wo[win_id].cursorlineopt
   if culopt:find('line') == nil then vim.wo[win_id].cursorlineopt = culopt .. ',line' end
 
@@ -2589,15 +2593,6 @@ H.window_set_view = function(win_id, view)
 
   -- Update border highlight based on buffer status
   H.window_update_border_hl(win_id)
-end
-
-H.window_set_cursor = function(win_id, cursor)
-  if type(cursor) ~= 'table' then return end
-
-  vim.api.nvim_win_set_cursor(win_id, cursor)
-
-  -- Tweak cursor here and don't rely on `CursorMoved` event to reduce flicker
-  H.window_tweak_cursor(win_id, vim.api.nvim_win_get_buf(win_id))
 end
 
 H.window_tweak_cursor = function(win_id, buf_id)
--- a/tests/dir-files/real/b.txt
+++ b/tests/dir-files/real/b.txt
@@ -1,4 +1,4 @@
-Line 1
+// Preview of a file should include the first column, even when the first line is long and has 3 slashes/
 Line 2
 Line 3
 Line 4
--- a/tests/screenshots/tests-test_files.lua---Preview---works-for-files-002
+++ b/tests/screenshots/tests-test_files.lua---Preview---works-for-files-002
@@ -1,6 +1,6 @@
 --|---------|---------|---------|---------|---------|---------|---------|---------|
 01|┌OOT/tests/dir-files/real ┐┌ b.txt ──────────────────┐                          
-02|│ a.lua                  ││Line 1                   │                          
+02|│ a.lua                  ││// Preview of a file shou│                          
 03|│ b.txt                  ││Line 2                   │                          
 04|│ c.gif                  ││Line 3                   │                          
 05|│ LICENSE                ││Line 4                   │                          


tests/dir-files/real/long-line.h
```diff
@@ -0,0 +1 @@
+//===--- Options.h - Option info & table ------------------------*- C++ -*-===//
```
echasnovski: I understand that this reproduces the issue, but it still isn't quite clear as to what is the *true* culprit is here. Like, if this is only reproducible with \*.h file, then there has to be some option or combinations of them that produces this result.

Locating the problem to this level will probably allow to reuse existing files (in a separate test) by first setting these options once or in the `FileType` autocommand.

lua/mini/files.lua
```diff
@@ -2577,7 +2577,7 @@ H.window_set_view = function(win_id, view)
   buf_data.win_id = win_id
 
   -- Set cursor (if defined), visible only in directories
-  pcall(H.window_set_cursor, win_id, view.cursor)
+  pcall(H.window_set_cursor, win_id, view.cursor, view.was_focused ~= false)
```
echasnovski: I think I like the [idea of passing `view` directly](https://github.com/nvim-mini/mini.nvim/issues/2514#issuecomment-5006218653) more. Renaming the function to `H.window_set_view_cursor()` and not using early return (i.e. `if view.was_focused then H.window_tweak_cursor(...) end`) looks like the cleanest approach here.

echasnovski (changes requested):
Thanks for the PR!

I'd like to try to find the exact culprit of the problem instead of using \*.h file.

echasnovski:
Oh, I only now have seen the second edit of [this comment](https://github.com/nvim-mini/mini.nvim/issues/2514#issuecomment-5006218653). The culprit is that the file contains a line with three or more `/` which are recognized as special indicators in directory buffers. This information is definitely needed in the commit message at least.

If that's the case, then I the more appropriate solution is to not tweak cursor when window shows a buffer for a file. And instead of adding a new file, prepending the `a.txt` test file with the wide line that contains three slashes with a text explanation in between is a less intrusive and more direct way to test this.

I think inlining the `H.window_set_cursor()` directly into the only place it is used is more beneficial here, since it can reuse the `H.fs_get_type(buf_data.path)` data to determine if the cursor needs to be tweaked.

abeldekat:
I pushed a new commit. 

For the test I used `b.txt`...

tests/dir-files/real/b.txt
```diff
@@ -1,3 +1,4 @@
+// Preview of a file should include the first column, even when the first line is long and has 3 slashes/
 Line 1
```
echasnovski:
```suggestion
// Preview of a file should include the first column, even when the first line is long and has 3 slashes/
```

This makes diff even smaller while still preserving the nature of the change.

lua/mini/files.lua
```diff
@@ -2577,7 +2577,13 @@ H.window_set_view = function(win_id, view)
   buf_data.win_id = win_id
 
   -- Set cursor (if defined), visible only in directories
-  pcall(H.window_set_cursor, win_id, view.cursor)
+  local is_table = type(view.cursor) == 'table'
+  if is_table then pcall(vim.api.nvim_win_set_cursor, win_id, view.cursor) end
+  if is_table and H.fs_get_type(buf_data.path) ~= 'file' then
+    -- Tweak cursor here and don't rely on `CursorMoved` event to reduce flicker
+    pcall(H.window_tweak_cursor, win_id, vim.api.nvim_win_get_buf(win_id))
+  end
```
echasnovski:
```suggestion
  local is_dir = H.fs_get_type(buf_data.path) == 'directory'
  if type(view.cursor) == 'table' then
    pcall(vim.api.nvim_win_set_cursor, win_id, view.cursor)
    -- Tweak cursor here and don't rely on `CursorMoved` event to reduce flicker
    if is_dir then pcall(H.window_tweak_cursor, win_id, buf_id) end
  end
```

This approach allows reusing the single check (please also use it when setting `'cursorline'`) of the path type and buffer identifier.

echasnovski (changes requested):
Already much better! Both the code+test changes and the commit message. There are still small tweaks to make it more polished.

lua/mini/files.lua
abeldekat: Done. After all the nested "if" I used recently, I was hesitant to apply that here...))

tests/dir-files/real/b.txt
abeldekat: Done

lua/mini/files.lua
echasnovski: > Done. After all the nested "if" I used recently, I was hesitant to apply that here...))

That's actually a good point. I think it is okay to reduce this even further without outer `if`. Made a force push.

echasnovski (approved):
Thanks for working on this!

PR #2543, opened 2026-08-30
test(jump2d): react to Nightly improving stuffed keys handling

abeldekat:

Details:
- See PR [41529](https://github.com/neovim/neovim/pull/41529) in neovim/neovim

Test `T['start()']['works in Operator-pending mode']` failed. I think it is a timing problem. When I manually follow the test steps everything seems right. 
Not sure if adding `sleep` is the best solution.

### DIFF
--- a/tests/test_jump2d.lua
+++ b/tests/test_jump2d.lua
@@ -296,7 +296,7 @@ T['start()']['works in Operator-pending mode'] = function()
 
   -- Allows dot-repeat
   type_keys('.')
-  child.expect_screenshot()
+  child.expect_screenshot({ ignore_text = { 4, 5 }, ignore_attr = { 4, 5 } })
   type_keys('c')
   child.expect_screenshot()
 end


abeldekat:
>  Not sure if adding sleep is the best solution.

Perhaps it's best to regenerate screenshot `tests/screenshots/tests-test_jump2d.lua---start()---works-in-Operator-pending-mode-003`

The "dot" on line 5 is similar to the "d" on line 5 in `tests/screenshots/tests-test_jump2d.lua---start()---works-in-Operator-pending-mode`

echasnovski:
> > Not sure if adding sleep is the best solution.
> 
> Perhaps it's best to regenerate screenshot `tests/screenshots/tests-test_jump2d.lua---start()---works-in-Operator-pending-mode-003`
> 
> The "dot" on line 5 is similar to the "d" on line 5 in `tests/screenshots/tests-test_jump2d.lua---start()---works-in-Operator-pending-mode`

Yeah, this does look like some issue with redraw. Which I think is some kind of regression of the linked PR, but not sure if it is consequential enough to create a contained repro and make an issue.

As the core tested stuff is showing the labels after the dot, I think it is safe to ignore non-buffer lines here. There are two approaches here:
- Add a `child.cmd('mode')` to clear and redraw before problematic screenshot(s). It can also be used instead of `child.cmd('redrawstatus')` earlier.
- Explicitly ignore text and attributes of non-buffer lines. In this case lines 4 and 5.

Both are fine with me. Let's go with whichever you prefer.

abeldekat:
> Both are fine with me. Let's go with whichever you prefer.

Thanks! I opted to explicitly ignore.

echasnovski (approved):
Thanks!

